PR #8, opened 2020-03-12
Ht 2219 add holdings

jsteverman:
Holdings can be added and updated. 

bin/add_holdings.rb accepts a delete flag that deletes all holdings for a member before adding a new set. I don't believe we have decided what to do about deleted records. Do we remove them? Mark them 'WD'?


aelkiss:
Superceded by #12

PR #18, opened 2020-06-05
Let bin/add_ht_items.rb run at all

billdueber:
* Add `ocn_resolution` model to requires
  * Skip HT items that don't have any OCNs to avoid
    error "E11000 duplicate key error collection: test.clusters index: ocns_1 dup key: { ocns: undefined } (11000) (on mongo:27017, legacy retry, attempt 1) (on mongo:27017, legacy retry, attempt 1) (Mongo::Error::OperationFailure)"


billdueber:
RE: logging stuff -- I didn't know the preferred way to do the logging stuff from a git perspective, since both PRs were in review at the same time. Planning on adding it once things are merged into master.

aelkiss:
Sounds good. Merging now

PR #19, opened 2020-06-05
HT-2377: Allow deleting concordance rule

aelkiss:
- Deletion works in the same way it did with ClusterMapper: delete the
resolution rule, delete the cluster, then re-add all the components

- Removes now-unused clustermapper

The specs for deleting an ocn resolution from a cluster that has HTItems in it should probably go in a separate spec -- they're really more of an integration test -- and we should consider doing that for HT-2376.


aelkiss:
will rebase and re-open - want to check something regarding travis

PR #20, opened 2020-06-05
run travis tests via docker

aelkiss:
Configures travis & adds mongo_test container separate from mongo_dev so that dev work doesn't stomp on test work.


aelkiss:
travis does say it's succeeded, so I'm going to go ahead and merge

PR #22, opened 2020-06-05
HT-2377: Allow deleting concordance rule

aelkiss:
- Deletion works in the same way it did with ClusterMapper: delete the
resolution rule, delete the cluster, then re-add all the components

- Removes now-unused clustermapper


jsteverman:
All looks good to me.

PR #24, opened 2020-06-11
HT-2376 update / delete

jsteverman:
Uses reclusterer. Delete is similar to OCNResolution delete, but searches for item_id instead of OCN. ClusterHTItem.update deletes then clusters updated record.

### DIFF
--- a/lib/cluster_ht_item.rb
+++ b/lib/cluster_ht_item.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "cluster"
+require "reclusterer"
 
 # Services for clustering HT Items
 class ClusterHtItem
@@ -32,4 +33,27 @@ class ClusterHtItem
     end
   end
 
+  # Removes an HTItem
+  def delete
+    Cluster.where("ht_items.item_id": @ht_item.item_id).each do |c|
+      c.ht_items.delete_if {|h| h.item_id == @ht_item.item_id }
+      Reclusterer.new(c).recluster
+    end
+  end
+
+  # Deletes an HTItem, then re-adds it
+  def update
+    Cluster.where("ht_items.item_id": @ht_item.item_id).each do |c|
+      ht = c.ht_items.to_a.find {|h| h.item_id == @ht_item.item_id }
+      if ht.ocns != @ht_item.ocns
+        ht.delete
+        Reclusterer.new(c).recluster
+      else
+        ht.update_attributes(@ht_item.to_hash)
+        c.save
+        return c
+      end
+    end
+    cluster
+  end
 end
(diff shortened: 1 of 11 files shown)


lib/cluster_ht_item.rb
```diff
@@ -32,4 +32,17 @@ def move(new_cluster)
...
+  def delete
+    Cluster.where("ht_items.item_id": @ht_item.item_id).each do |c|
+      c.ht_items.delete_if {|h| h.item_id == @ht_item.item_id }
+      Reclusterer.new(c).recluster
+    end
+  end
+
+  # Deletes an HTItem, then re-adds it
+  def update
+    delete
```
aelkiss, line 45: This is definitely the simplest implementation and good for now. I think we will probably want to add an optimization for the usual case where an update doesn't change OCNs, but I think we can wait for that until we have a process that handles the updates and we have a sense of how many things get updated and how long it takes to process on a daily basis.

spec/cluster_ht_item_spec.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "pp"
```
aelkiss, line 3: leftover from debugging?

lib/cluster_ht_item.rb
jsteverman: When I ran this I was seeing ~10  records/second. This implies somewhere in the range of an hour to handle an average days worth of updates depending upon a slew of other variables. Manageable, but definitely not ideal.

I changed it to update_attributes for an HTItem, and only delete/recluster when an OCN changed. Running a days worth of updates twice (add, then update) took ~2 minutes as opposed to the hour long naive approach.

jsteverman:
I added a paste-eater's version of an update flag to the bin/add_ht_items.rb so it will update rather than a simple cluster if given a -u. 

Considering the ClusterHtItem.update is now relatively performant (~200 r/s), it might make sense to always run an update. 
.update is basically:
   check ht_item.item_id index, do an update or delete as appropriate
   if not found or old one deleted then cluster it

lib/cluster_ht_item.rb
```diff
@@ -42,7 +43,17 @@ def delete
 
   # Deletes an HTItem, then re-adds it
   def update
-    delete
+    Cluster.where("ht_items.item_id": @ht_item.item_id).each do |c|
```
aelkiss, line 46: Do we have something that makes an index for this?

jsteverman: class Cluster
  ...
  index({ "ht_items.item_id": 1 }, unique: true, sparse: true)
  ...

PR #27, opened 2020-06-29
Ht 2428 (n_enum & n_chron for mpms)

mwarin:
Added n_enum and n_chron to ht_items. They are populated by the enumchronparser (such as it is) when an ht_item is created with a enum_chron value.

Based on the changes in the autoscrub branch (where the enumchronparser first got added) which hasn't been merged to master yet.


aelkiss:
We should do this in the same place for holdings records (i.e. at load time), but currently it's doing it via the scrub process. We'll create a follow-up issue for that.

PR #32, opened 2020-07-11
First attempt at generating holdings overlap reports

jsteverman:
Adds `condition` to Holdings records.
Adds `content_provider_code` to HTItems

Adds Overlap record class to compute holdings *_counts for a Serial cluster (does almost nothing since we don't really need it)
Adds SinglePartOverlap for SPMs.
Adds MultiPartOverlap for MPMs.

Adds CollectOverlap to iterate over overlapping HTItems.

Adds `bin/export_overlap_report.rb umich` to dump the results from CollectOverlap as tsv

Known issues:
It raises an error when dealing with "ser/spm" clusters. What are we supposed to do with them?

access_count is wrong. I am only seeing documentation for UMich's brittle access program. Is this relevant for ETAS and is it computed differently.

MPMs are likely computed incorrectly or at least differently from existing practice.

Handling of MPMs without enum_chrons is undefined.

Test coverage isn't great, especially inre to corner cases. 

SPO and Overlap.copy_count do not handle missing holdings like MPO.copy_count, i.e. when the content_provider_code matches the member_id.

### DIFF
--- a/bin/add_ht_items.rb
+++ b/bin/add_ht_items.rb
@@ -18,12 +18,13 @@ Mongoid.load!("mongoid.yml", :test)
 def hathifile_to_record(hathifile_line)
   fields = hathifile_line.split(/\t/)
   {
-    item_id:    fields[0],
-    ocns:       fields[7].split(",").map(&:to_i),
-    ht_bib_key: fields[3].to_i,
-    rights:     fields[2],
-    bib_fmt:    fields[19],
-    enum_chron: fields[4]
+    item_id:               fields[0],
+    ocns:                  fields[7].split(",").map(&:to_i),
+    ht_bib_key:            fields[3].to_i,
+    rights:                fields[2],
+    bib_fmt:               fields[19],
+    enum_chron:            fields[4],
+    content_provider_code: fields[21]
   }
 end
 
--- a/bin/export_overlap_report.rb
+++ b/bin/export_overlap_report.rb
@@ -1,0 +1,52 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "dotenv"
+Dotenv.load(".env")
+
+$LOAD_PATH.unshift(File.join(File.dirname(__FILE__), "..", "lib"))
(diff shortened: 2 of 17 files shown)


jsteverman:
MPM HTItems without enum chrons now match every holding in the cluster. 
MPM Holdings without enum chrons match every ht_item in the cluster.

SerialOverlap is now a separate thing. SerialOverlap.new.copy_counts returns 1 if there is any matching holding in the cluster

SerialOverlap and SinglePartOverlap both check for content_provider_code when computing copy_counts.

Serial/SPM clusters still need to be dealt with. 

content_provider_code is still a headache in waiting. 

Solution to computing overlaps per HTItem rather than holdings org is still pending. Possible solution, HtItemOverlap to hold overlap records for every organization along with some rollup fields.

jsteverman:
CollectOverlap is now ClusterOverlap which cycles through `ht_items.each -> organization.each` where organization is either a given organization or every organization found in that cluster's ht_item.content_provider_code and holding.organization.

bin/export_overlap_report.rb is now responsible for selecting which clusters to report on, using criteria for organization == nil (all) or a particular organization. 

A solution to holdings organization <=> content_provider_code is unresolved. 

SER/SPM clusters are still not handled appropriately.

jsteverman:
MPMs and Serials appear to have special handling in the existing print holdings code. I cannot find evidence of special handling for ser/spm clusters. I spot checked some ser/spm clusters and many have copy_counts > 1 indicating there is special item level handling I was unable to find, or they are processed as SPMs. 

ClusterOverlap now uses SinglePartOverlap for ser/spm clusters.

aelkiss:
I think it would be helpful to get a review from @billdueber just for general structural things but otherwise I'm happy to proceed with this.

aelkiss:
It's been sitting in review for two weeks and we haven't had the time to do additional review, so I'm making an executive decision to go ahead and merge it. If we find things we want to change or refactor later we can open new issues to do that.

PR #33, opened 2020-07-21
HT-2415 ensure indices on clusters

jsteverman:
Dodging rake for the moment to simply run Cluster.ensure_indexes for Travis.

### DIFF
--- a/lib/tasks/build_database.rb
+++ b/lib/tasks/build_database.rb
@@ -2,7 +2,7 @@
 
 require "mongoid"
 
-Mongoid.load!("mongoid.yml", :test)
+Mongoid.load!("mongoid.yml", ENV["MONGOID_ENV"] || :development)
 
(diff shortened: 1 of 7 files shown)


jsteverman:
I added a very simple lib/tasks/build_database.rb which can be run after bringing up docker. 

Should not be necessary if rspec will be run.

lib/tasks/build_database.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require "mongoid"
+
+Mongoid.load!("mongoid.yml", :test)
```
aelkiss, line 5: It should probably do this for all the environments defined in `mongoid.yml`?

aelkiss:
I changed it (and the bin scripts) to use the 'development' environment by default, and adjusted the travis config appropriately. That should also help with the issue of the tests clobbering experimentation using the bin scripts.

PR #35, opened 2020-08-18
Ht 1971 calculate member fees

jsteverman:
Naive approach to computing hscores

HtItemOverlap uses cluster_overlap to provide a list of unique members that hold an item. bin/compile_hscores leverages this to compute a frequency table of member=> { num_members : num_matching_items } and a total hscore.
Actual report format is TBD; currently a JSON dump of the totals table.

The set of members that hold a particular item is generated when the hscores are being compiled, which necessitates a retrieval of the holdings records. This is suboptimal as the set of members could be retained in the HtItem, if we ensure it is updated when appropriate.


jsteverman:
See https://tools.lib.umich.edu/jira/browse/HT-1971 for more discussion.

The report format from bin/compile_hscores.rb remains an open question.

jsteverman:
The bulk of it has been moved into lib/cost_report.rb. The spec for it is a bit of a mess, but I think that's more a result of our complicated overlap rules than bad code.

aelkiss:
Looks pretty good to me.

aelkiss:
It might be worthwhile to add a test to ensure correct h-counts when an institution reports multiple holdings of a particular item -- I think the behavior is correct but it isn't clear to me there's a test that fully covers that case.

jsteverman:
Overlap was previously calculated as [holdings.org] + ht_item.content_provider_code.
It is now [holdings.org] + [matching ht_item.content_provider_code]. 

https://github.com/hathitrust/holdings-backend/blob/ea7f197d6cbffb5e466ca2c08a72f3cb42e4f047/lib/overlap.rb#L25-L27

PR #37, opened 2020-08-25
Ht 2454

mwarin:
PR for https://tools.lib.umich.edu/jira/browse/HT-2454


lib/ht_members.rb
```diff
@@ -1,17 +1,151 @@
...
+      mapto_inst_id
+    ) AS member_id,
+    country_code,
+    weight,
+    oclc_sym
+    FROM ht_institutions
+    WHERE enabled = '1'
+  >.join(' ')
+
+  CANNED_DATA = {
```
aelkiss, line 65: What's the use case for having the canned data rather than just passing it in, if one doesn't want to read it from the database?

aelkiss: I can see wanting to have sample data for development purposes, but that seems like perhaps it should be the responsibility of whatever constructs `HTMembers` to provide?

lib/ht_members.rb
```diff
@@ -1,17 +1,151 @@
...
+      @members[member_id]
+    else
+      warn "No member_info data for member_id:#{member_id}"
+      {"country_code"=>"xx", "weight"=>0.0, "oclc_sym"=>""}
+    end
   end
+
+end
+
+if $0 == __FILE__ then
```
aelkiss, line 134: Seems like this is leftover from development?

spec/ht_members_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+  let(:mock_umich) {
+    {
+      "umich" => HTMembers.template("us", 1.33, "EYM")
+    }
+  }
+
+  describe "mocking" do
+    it "allows you to send in mock data through new({...})" do
+      htm = HTMembers.new(mock_umich)
+      expect(htm.mocked).to be(true)
```
aelkiss, line 16: `mocked`/`canned` also seem to maybe be leftovers from development -- which is to say their main purpose seems to be to support this and similar tests. Those tests seem to be primarily for the purpose of making sure that as you're writing tests the tests pass or fail for the reasons you expect -- but now that you've verified that, I think those tests aren't continuing to do a lot of work and can be removed.

lib/ht_members.rb
aelkiss: Or rather should be moved to the specs to ensure that we can pass in data and get back what we expect..

aelkiss:
See https://github.com/hathitrust/holdings-backend/pull/38, which adds the container for mariadb. I haven't actually tried the integration tests, though.

lib/ht_members.rb
```diff
@@ -1,17 +1,151 @@
...
+        :connect_timeout => 5,
+      )
+
+      res = mysql_client.query(SQL)
+      res.each do |row|
+        data[row[:member_id]] = HTMembers.template(
+          row["country_code"], row["weight"], row["oclc_sym"]
+        )
+      end
+    rescue Mysql2::Error::ConnectionError => e
```
billdueber, line 113: This fall-through to canned data makes me uncomfortable. Am I right that if the DB is down or times out or something, we could end up using canned data when we don't want it and not notice until all sorts of damage had been done to the mongo datastore.

lib/ht_members.rb
```diff
@@ -1,17 +1,151 @@
...
+
+  attr_reader :mocked, :canned, :members
+  def initialize(members = {})
+    @mocked  = !members.empty?
+    @canned  = false
+    @members = @mocked ? members : load_from_db
+  end
+
+  # Are all the ENV vars required for DB connection set?
+  def db_env_set?
```
billdueber, line 80: At least at the moment this isn't being used at all except in the specs. Just noting.

lib/ht_members.rb
```diff
@@ -1,17 +1,151 @@
...
+      "weight"       => wt,
+      "oclc_sym"     => sym,
+    }
+  end
 
-  def initialize(*members)
-    @members = members.empty? ? load_from_db : members
+  # If we get more cases like universityofcalifornia then this might
+  # look better as a series of SELECT ... UNION SELECT ....
+  SQL = %w<
```
aelkiss, line 51: The more I think about it I don't think we need the distinct/mapto query. In particular, we aren't using this to check if a holdings ID is valid as such or that we expect to get it, and so long as we load the holdings under the correct ID I don't think we need to worry about filtering or mapping the institution IDs here. If we need that functionality for billing calculation we can use it.

aelkiss:
Superceded by #40

aelkiss:
@mwarin Do you want to keep anything from this branch, or can we delete it?

PR #38, opened 2020-09-01
HT-2595: mariadb instance w/ institution & collections data

aelkiss:
This only adds relevant columns & rows from ht_institutions. Not sure if
we need even all this, but I don't think it's a problem to have this
public - our members & the formula for computing the weights are all
public.

This only adds the service in docker-compose.yml; it still needs to be configured to actually be used either by tests or load scripts running in development.


aelkiss:
Superceded by #39

PR #39, opened 2020-09-04
Create basic infrastructure for Mysql Database

billdueber:
This is based on (and obviates) HT-2595 / PR #38

* Pull in files from PR #38 that create a mariadb
container and define initial tables/grants/data for
ht_institutions and ht_collections (via files in
sql/*)

* Change docker-compose so mariadb uses a random
root password (which we don't use because of the
grants in sql/002)

* Create a simple connection-builder in
`lib/holdings_db` that exposes a single method
`#connection` to get a connection to the DB based
on a connection string, a set of key/value pairs,
environment variables, or a combination of the latter
two.

* basic tests (connect when expected, find tables
populated) in `spec/holdings_db_spec.rb`


aelkiss:
@billdueber  Are we OK with going ahead and merging this now?

PR #40, opened 2020-09-08
HT-2454: HTMember with country code & weight

aelkiss:
- Uses canister for holdings DB & global cache of member info

- Auto-sets organization & country code when constructing holding

Supercedes #37

### DIFF
--- a/lib/ht_members.rb
+++ b/lib/ht_members.rb
@@ -55,8 +55,16 @@ class HTMembers
     if @members.key?(inst_id)
       @members[inst_id]
     else
-      raise "No member_info data for inst_id:#{inst_id}"
+      raise KeyError, "No member_info data for inst_id:#{inst_id}"
     end
+  end
+
+  # Adds a temporary member to the member data cache for the lifetime of the
+  # object; does not persist it to the database
+  #
+  # @param member The HTMember to add
+  def add_temp(member)
+    @members[member.inst_id] = member
   end
 
 end
--- a/lib/ht_members.rb
+++ b/lib/ht_members.rb
@@ -1,17 +1,62 @@
 # frozen_string_literal: true
 
-# WIP: cache of information about HathiTrust members
+require "mysql2"
+require "services"
+
+# Information about an individual HathiTrust institution
+class HTMember
+  attr_reader :inst_id, :country_code, :weight, :oclc_sym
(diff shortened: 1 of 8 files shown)


aelkiss:
Need to check the rubocop failures & one spec failure

lib/ht_members.rb
```diff
@@ -1,17 +1,66 @@
...
   end
 
-  def get_list
-    @members
+  # Given a inst_id, returns a hash of data for that member.
+  def [](inst_id)
+    if @members.key?(inst_id) then
+      @members[inst_id]
+    else
+      raise "No member_info data for inst_id:#{inst_id}"
```
aelkiss, line 61: Might want to change this back to a warning

aelkiss: Could just remove the warning altogether?

aelkiss: Could also add the new member to the cached member data when doing an estimate

aelkiss:
Also need to make sure the spec for the holding gets added here.

lib/ht_members.rb
billdueber: I say leave it as an error. If we want to pretend some institution is a partner, we should change it at the source layer. That means HTMember needs `#temp_add` or something. I'd resist just calling it `add` or `<<`  -- make it clear it's not changing the underlying datastore.

lib/ht_members.rb
```diff
@@ -1,17 +1,66 @@
...
+      HTMember.new(inst_id: inst_id)
```
billdueber, line 62: Should this line be outside the next `end`? Right now it'll never be reached.

billdueber (commented):
Failing test is due to the new error for unknown institution -- need to change the definition of `holdings_serial` to use a real org. Maybe also add a test to make sure the error is raised on bad inst name?

lib/ht_members.rb
aelkiss: No, it should just be removed.

lib/ht_members.rb
aelkiss: Sounds good. Pushed a commit with this.

aelkiss:
@billdueber Comments address and Travis is happy. Merge when ready.

PR #41, opened 2020-09-09
Ht 2597 cpc to member

jsteverman:
Based off HT-2454-country-code, and essentially the same implementation. 
 
Needs some code clean up, documentation and git rebase.


aelkiss (approved):
Looks good to me, but it might be worth double-checking that all the replacements in cost_report_spec.rb were intended.

spec/cost_report_spec.rb
```diff
@@ -198,7 +198,7 @@
               ocn: ht_serial.ocns.first,
               enum_chron: "3",
               n_enum: "3",
-              organization: "not_a_cpc")
+              organization: "okstate")
```
aelkiss, line 201: I assume this change is because it was trying to look up billing entity for `not_a_cpc` and couldn't find it? I can't think of a case where we would need to look up billing entity for a temporarily-added institution, since it only comes into play for HTitems -- and if we have an htitem, we should have the collection & billing entity info correct..

jsteverman: I think this test was blowing up when I had the HT-2597 branch based on an incomplete HT-2454 (HTMembers/holdings) branch. I chose "okstate" because I knew it wouldn't actually show up in the collection_code for the factory generated ht_item, but also wanted something otherwise valid. 

The version of HT-2454 that made it to master solved the problem by add_temp("not_a_cpc"). When I rebased on master we ended up with 2 solutions to the same problem, one of which is now unnecessary. At least, I think that's what happened.

The add_temp for HTCollections is a separate issue that I had considered and rejected for the reasons you described.

PR #42, opened 2020-09-10
HT-2475 load overlap export into db

billdueber:
Provide LOAD DATA INFILE capability to get overlap data into `holdings_htitem_htmember`

* New sql table definition in `/sql` (`holdings_htitem_htmember` with extra `cluster_id` column)
* Make `HoldingsDB` a real class with instances. Instances delegate pretty much everything to the underlying Sequel connection. Change spec to suit.
* Now that it's an object, add `HoldingsDB#load_tab_delimited_file` which takes a table and filename and:
   - uses unix utility _split_ to split the source into (default) one-million-line files in a temporary directory
   - load each file with a pause between (default 10 seconds)
   - clean up after itself
* Create a spec. Most of it is copy-paste from `overlap_spec` to make a few cluster/items and then export them.
* Minor changes to `bin/export_overlap_report.rb` to make it easier to call the code in a spec

## TODO
* check to make sure `split` exists (or maybe just use ruby? Slower, I'm sure.)
* actually make the script in `./bin/` that calls all this
* Have the import drop keys beforehand and then recreate them afterwards, for performance increase
* (?) Implement an upsert method with a signature similar to `#load_tab_delimited_file` (?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@ config/environments/*.local.yml
 /docs/_build/
 /docs/_yard/
 *.iml
+/testdata
--- a/bin/export_overlap_report.rb
+++ b/bin/export_overlap_report.rb
@@ -26,6 +26,17 @@ def matching_clusters(org = nil)
   end
 end
 
+def overlap_line(overlap_hash)
+  [overlap_hash[:cluster_id],
+   overlap_hash[:volume_id],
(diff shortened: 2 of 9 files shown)


aelkiss:
@billdueber What tasks remain before this PR is ready for review?

PR #44, opened 2020-09-15
HT-2570 Generate ETAS overlap reports

jsteverman:
* use existing overlap classes
 * write to files at .env defined path

Bit kludgier than I would like. Generates files: <.env path>/<member_id>_Y-m-d.tsv

### DIFF
--- a/bin/export_etas_overlap_report.rb
+++ b/bin/export_etas_overlap_report.rb
@@ -1,0 +1,64 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "dotenv"
+Dotenv.load(".env")
+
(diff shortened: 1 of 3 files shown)


aelkiss (commented):
Might be nice to extract the body of export_etas_overlap_report.rb to a class and have a spec that its format is correct, but the format is simple enough that I think it isn't critical since `CalculateFormat` and `ClusterOverlap` are already tested. If nothing else I think it would be good to get feedback from @mwarin that the format looks correct.

jsteverman:
I added an ETASOverlap class to hold the record and .to_s. It seems a little silly, but gives us something to test against.

mwarin (approved):
The tests are rather simple but I don't have anything constructive to say.
Does the report output contain a header line?
I see no reason not to merge.

PR #45, opened 2020-09-17
HT-2463, HT-2491, HT-2480: Batching and transaction support

aelkiss:
This is a large PR, but primarily addresses concerns around:

1) performance with respect to large clusters, by batching operations that pertain to a single OCN

2) concurrency support, by:
 - using multi-document transactions when merging and splitting
 - using the underlying Mongo update operator to allows detection of if we updated a document deleted out from under us
 - retrying operations when they fail
 - adding specs to verify this all works.

It also:
 - adds some support towards using this code in loading data in production
 - cleans up the output from specs by using `Logger` instead of `puts`

The next step is to apply the changes to `ClusterHtItem` to the other `Cluster` classes, and likely extract some of the common functionality from those objects, but I would like to get this PR merged and into master before I attempt that.

### DIFF
--- a/lib/services.rb
+++ b/lib/services.rb
@@ -9,6 +9,11 @@ require "logger"
 Services = Canister.new
 
 Services.register(:holdings_db) { HoldingsDB.new }
+Services.register(:ht_members) { HTMembers.new }
 Services.register(:ht_collections) { HTCollections.new }
 Services.register(:ht_members) { HTMembers.new }
-Services.register(:logger) { Logger.new(STDERR) }
+Services.register(:logger) do
+  Logger.new(STDERR).tap do |l|
+    l.level = Logger::INFO
+  end
+end
--- a/spec/cluster_ht_item_spec.rb
+++ b/spec/cluster_ht_item_spec.rb
@@ -1,151 +1,121 @@
-# frozen_string_literal: true
-
 require "spec_helper"
 require "cluster_ht_item"
 RSpec.describe ClusterHtItem do
-  let(:ht) { build(:ht_item) }
-  let(:c) { create(:cluster, ocns: ht.ocns) }
+  let(:item) { build(:ht_item) }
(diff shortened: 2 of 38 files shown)


lib/services.rb
```diff
@@ -7,6 +7,12 @@
 
 Services = Canister.new
 
+Services.register(:logger) do
+  Logger.new(STDOUT).tap do |l|
```
aelkiss, line 11: Should log to STDERR by default.

spec/cluster_ht_item_spec.rb
```diff
@@ -64,97 +123,99 @@
...
+
+      it "removes the old cluster" do
+        first = described_class.new(item).cluster
+        described_class.new(update_item).cluster
+        expect(Cluster.each.to_a.size).to eq(1)
+        new_cluster = Cluster.each.to_a.first
+        expect(new_cluster).not_to eq(first)
+        expect(new_cluster.ht_items.first).to eq(update_item)
+      end
+    end
```
jsteverman, line 138: I am not clear how this is testing what it says it is testing. Talk about it monday.

jsteverman: I confirmed it does what it says it is doing.

aelkiss: FWIW I think this was a test that was there before, but was moved to a different context. One of the changes in behavior in this PR is that "cluster" always does an update if necessary.

jsteverman: I didn't understand how update_or_add_htitems could satisfy this test, which had me doubting the meaning of 
`expect(new_cluster).not_to eq(first)`
I will surely forget this again, "default comparison is via the string version of the id," i.e. changes to the ocns are insufficient to pass this test.

jsteverman (approved):

PR #46, opened 2020-09-18
HT-2578 Compile estimated HScore for list of ocns

jsteverman:
- computes only HScore not cost
  - doesn't deal with public domain fees
  - lacks tests, although I'm not sure what to test besides a comparison to a full build

### DIFF
--- a/bin/compile_estimated_hscore.rb
+++ b/bin/compile_estimated_hscore.rb
@@ -1,0 +1,53 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "dotenv"
+Dotenv.load(".env")
+
+$LOAD_PATH.unshift(File.join(File.dirname(__FILE__), "..", "lib"))
+require "bundler/setup"
+require "utils/waypoint"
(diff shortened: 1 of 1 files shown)


aelkiss:
As far as a test, I think it would probably make sense to:

- load a few (maybe 5?) synthetic HTItems and holdings records that cover various scenarios -- spm, mpm, serial, with holdings from 1 or 2 institutions

- "estimated" holdings should have perhaps 3 of those items as well as a couple that aren't in the synthetic data

- then hand-compute the expected report and compare to the actual generated report

I think that would be helpful for some of the other reports as well.

That said I think we can go ahead and merge this and leave that integration test for a future issue

mwarin (approved):
Looks fine. 
I wonder what the memory requirements will be when we run this for real against a big list of ocns matching a big set of ht_items. But a) new members these days tend to be smaller and b) we'll deal with it if/when we get there.

aelkiss:
Looks like memory requirements should be minimal, since the only thing it keeps between iterations of the loop is the total h score.

PR #50, opened 2020-09-29
HT-2366: Batch loading & idempotence for all updates

aelkiss:
This adds support for batch loading for holdings & OCN resolutions.

It also ensures that re-running a holdings file or OCN resolution table does not result in duplicates or errors. This is important because we want to make sure we can support the case where loading a file encountered an error in the middle of the file as well as the case where we have to retry a batch with potential uncertainty about whether some pieces were already handled or not.

This also includes some refactoring around getting a cluster, including merging if necessary, to reduce the duplication and simplify the code path in `ClusterHTItem`, `ClusterOCNResolution` and `ClusterHolding`. It also ensures retryability and transactions (where necessary) for all operations on a cluster via `Retryable` (rather than splitting this functionality between `Cluster` and `ClusterHTItem`)

It also changes the concurrency spec to ensure that we're still testing what we claim to be testing there; in particular `ClusterGetter` now needs to be instrumented as well.

It also refactors and tests the bin scripts for loading data; there's probably still more refactoring and integration testing that could be done, but this should be substantially less brittle than before.

Finally, it adds a script to add a unique UUID to incoming holdings data.


lib/retryable.rb
```diff
@@ -0,0 +1,64 @@
...
+  MONGO_DUPLICATE_KEY_ERROR=11_000
+
+  def self.ensure_transaction
+    if (s = Mongoid::Threaded.get_session)
+      raise "In a session but not in a transaction??" unless s.in_transaction?
+
+      yield
+    else
+      Cluster.with_session do |session|
+        session.with_transaction { new.run { yield } }
```
aelkiss, line 17: Something funny is going on here -- `with_transaction` ends up wrapping it in two `Retryable` blocks. I think the intention was that `ensure_transaction` would wrap the block in a transaction if needed but not do retryable part. I will change `ensure_transaction`

bin/add_uuid.rb
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# Appends a UUID to each incoming holdings line
+
+require "securerandom"
+
+ARGF.each do |line|
+  if /^OCN\tBIB/.match?(line)
+    puts [line.strip, "UUID"].join("\t")
```
jsteverman, line 10: The `line.strip` concerned me for a moment. I have confirmed holdings file lines will not end in an empty field because that is uniformly the GOV_DOC field which gets set one way or the other.

spec/cluster_getter_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    let(:c1) { create(:cluster, ocns: [ocn1]) }
+    let(:c2) { create(:cluster, ocns: [ocn2]) }
+    let(:htitem1) { build(:ht_item, ocns: [ocn1]).to_hash }
+    let(:htitem2) { build(:ht_item, ocns: [ocn2]).to_hash }
+    let(:holding1) { build(:holding, ocn: ocn1).attributes }
+    let(:holding2) { build(:holding, ocn: ocn2).attributes }
+    let(:ocn_resolution1) { build(:ocn_resolution, resolved: ocn1, deprecated: ocn3).attributes }
+    let(:ocn_resolution2) { build(:ocn_resolution, resolved: ocn2, deprecated: ocn4).attributes }
+
+    let(:merged_cluster) { described_class.new([ocn1, ocn2]).get }
```
jsteverman, line 28: These bother me a little bit, because they generate clusters without "glue", i.e. htitem1 and htitem2 shouldn't be in the same cluster. On the other hand, part of the test is that it's not ClusterGetter's responsibility to manage cluster integrity, so maybe I should get over it.

jsteverman:
Looks good. I would like a quick walk through of spec/concurrency_spec.rb at one of our future meetings. More for my personal edification than actual concern.

spec/cluster_getter_spec.rb
aelkiss: Right -- what would happen "for real" is that if you went to add that glue (either an htitem or ocn resolution containing both ocn1 and ocn2), `ClusterGetter` would first merge the clusters for ocn1 and ocn2, and the actual glue would be added after the cluster was changed. If the cluster gets merged, that happens inside a transaction, and the "glue" should get added within the scope of that transaction (so if it fails / gets rolled back, the cluster merge should get rolled back as well). We could add a test for that if we feel it would be helpful.

billdueber (approved):
I'm still trying to work my way through understanding all the sync stuff, but in general this was a pleasure to read. Seriously.

bin/add_uuid.rb
aelkiss: Yes, it should be `chomp`

lib/retryable.rb
```diff
@@ -0,0 +1,64 @@
...
+  def run
+    @tries = 0
+
+    begin
+      @tries += 1
+      yield
+    rescue Mongo::Error::OperationFailure => e
+      @error = e
+      retryable_error? && more_tries? && retry || raise
+    rescue ClusterError => e
```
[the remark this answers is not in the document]
aelkiss: Yes, we could write this as 

```ruby
rescue Mongo::Error::OperationFailure, ClusterError => e
  @error = e
  retryable_error? && more_tries? && retry || raise   
```

Not sure what happened there.

lib/cluster_ht_item.rb
```diff
@@ -2,118 +2,72 @@
...
+  def initialize(*ht_items)
+    @ht_items = ht_items.flatten
+    @ocns = @ht_items.first.ocns
 
-  def initialize(*htitems)
-    @htitems = htitems.flatten
-    @ocns = @htitems.first.ocns
-
-    if @htitems.find {|h| h.ocns != @ocns }
+    if @ht_items.find {|c| c.ocns != @ocns }
```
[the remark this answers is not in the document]
aelkiss: For consistency this should probably call `Holding#batch_with?`. While something with the ocns `[2,1]` could be batched with ocns `[1,2]`, it isn't really a problem that it isn't, since the batching is just an optimization to being with.

lib/file_loader.rb
```diff
@@ -0,0 +1,44 @@
...
+  def initialize(batch_size: 10_000, batch_loader:)
+    @logger = Services.logger
+    @waypoint = Utils::Waypoint.new(batch_size)
+    @batch_size = batch_size
+    @batch_loader = batch_loader
+  end
+
+  def load(filename, filehandle: Zinzout.zin(filename))
+    logger.info "Loading #{filename}, batches of #{ppnum @batch_size}"
+
```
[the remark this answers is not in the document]
aelkiss: Cool, I didn't know about `chunk_while`.

PR #53, opened 2020-10-01
Bugfix: HTItems w/o n_enum match all other items

jsteverman:
- HTItems with n_enum were matching those without, but not the reverse.
  - n_enum and n_chron default to "" instead of nil

### DIFF
--- a/lib/enum_chron.rb
+++ b/lib/enum_chron.rb
@@ -7,8 +7,9 @@ require "services"
 module EnumChron
 
   def initialize(params = nil)
+    params ||= {}
+    params[:enum_chron] ||= ""
     super
-    normalize_enum_chron
   end
 
   def enum_chron=(enum_chron)
@@ -18,7 +19,6 @@ module EnumChron
 
   def normalize_enum_chron
     # normalize into separate n_enum and n_chron
-    enum_chron = self.enum_chron || ""
     ec_parser = EnumChronParser.new
     ec_parser.parse(enum_chron)
     self.n_enum  = ec_parser.normalized_enum || ""
--- a/lib/enum_chron.rb
+++ b/lib/enum_chron.rb
@@ -18,11 +18,10 @@ module EnumChron
 
   def normalize_enum_chron
     # normalize into separate n_enum and n_chron
-    unless enum_chron.nil?
-      ec_parser = EnumChronParser.new
-      ec_parser.parse(enum_chron)
-      self.n_enum  = ec_parser.normalized_enum
-      self.n_chron = ec_parser.normalized_chron
-    end
+    enum_chron = self.enum_chron || ""
+    ec_parser = EnumChronParser.new
+    ec_parser.parse(enum_chron)
+    self.n_enum  = ec_parser.normalized_enum || ""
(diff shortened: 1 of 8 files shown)


lib/enum_chron.rb
```diff
@@ -18,11 +18,10 @@ def enum_chron=(enum_chron)
 
   def normalize_enum_chron
     # normalize into separate n_enum and n_chron
-    unless enum_chron.nil?
-      ec_parser = EnumChronParser.new
-      ec_parser.parse(enum_chron)
-      self.n_enum  = ec_parser.normalized_enum
-      self.n_chron = ec_parser.normalized_chron
-    end
+    enum_chron = self.enum_chron || ""
```
aelkiss, line 21: Is `enum_chron` here a local that's shadowing `self.enum_chron`? I'm a little confused about why it's written this way -- `self.enum_chron ||= ""` might be more clear if that's the intention

lib/multi_part_overlap.rb
```diff
@@ -6,9 +6,13 @@
 class MultiPartOverlap < Overlap
 
   def members_with_matching_ht_items
-    @cluster.ht_items.where("$or": [{ n_enum: @ht_item.n_enum },
-                                    { n_enum: nil }])
-      .pluck(:billing_entity).uniq
+    if @ht_item.n_enum == ""
+      @cluster.ht_items.pluck(:billing_entity).uniq
+    else
+      @cluster.ht_items.where("$or": [{ n_enum: @ht_item.n_enum },
```
aelkiss, line 12: Not sure if the mongo syntax is preferable here, or if we could get by with

```ruby
@cluster.ht_items.select {|h| h.n_enum.empty? || h.n_enum == @ht_item.n_enum}
```

spec/enum_chron_spec.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "enum_chron"
+
+class DummyRecord
+  include Mongoid::Document
```
aelkiss, line 7: What happens if we don't include this?

aelkiss (commented):
See comments inline. Most of it doesn't directly relate to the changes here - any nontrivial fixes don't need to block merging this IMO.

lib/enum_chron.rb
jsteverman: `self.enum_chron ||= ""` within `normalize_enum_chron` will blow up because `def enum_chron=` calls `normalize_enum_chron`. Made irrelevant shortly.

spec/enum_chron_spec.rb
jsteverman: Mongoid's initialize has behavior I have been unable to entirely grok or duplicate. Something, something dynamic field assignment. I can remove the dependency in DummyRecord with a small change to EnumChron, but then Mongoid::Documents have duplicate calls to normalize_enum_chrons.

aelkiss: OK -- fine to leave as-is then -- most of the test suite depends on mongo at this point anyway

aelkiss (approved):

PR #54, opened 2020-10-06
HT-2662 Account for h_shares by bib format

jsteverman:
Total hshares for a particular member had been compiled, but our desired reports have them separated by bib format. The frequency table needed to be adjusted accordingly. 

This is a direct lead into https://tools.lib.umich.edu/jira/browse/HT-2635, integration testing of the reports, so it may make more sense to carry this branch forward than to merge now.


jsteverman:
HT-2635 Cost Report testing
     * cost_report.to_tsv (which is hideous)
     * Calculates costs per format
     * Calculates total IC costs
     * Calculates weighted PD costs
     * Distributes "extra" costs
    
Rubocop does not like the length of CostReport class.

lib/cost_report.rb
```diff
@@ -1,15 +1,89 @@
...
 class CostReport
-  attr_accessor :organization, :freq_table
+  attr_accessor :organization, :logger, :maxlines, :target_cost
 
-  def initialize(org = nil)
+  def initialize(org = nil, cost: ENV["target_cost"], lines: 1_000_000, logger: Services.logger)
     @organization = org
-    # { org => { 1 org : count, 2 org : count }
-    @freq_table = Hash.new {|hash, key| hash[key] = Hash.new(0) }
+    @target_cost = cost.to_f
```
billdueber, line 13: I'm worried this will silently come in as a zero or as something stupid. `Float(n)` will throw an error if `n` is `nil` or can't be parsed into a float.

lib/cost_report.rb
```diff
@@ -1,15 +1,89 @@
...
+        add_ht_item_to_freq_table(ht_item)
+        waypoint.on_batch {|wp| logger.info wp.batch_line }
+      end
+    end
+  end
+
+  def add_ht_item_to_freq_table(ht_item)
+    overlap = HtItemOverlap.new(ht_item)
+    overlap.matching_orgs.each do |org|
+      @freq_table[org.to_sym][ht_item._parent.format.to_sym][overlap.matching_orgs.count] += 1
```
billdueber, line 85: I would move the `#to_sym` calls into the definition of `freq_table`; easier to not forget them.

lib/cost_report.rb
```diff
@@ -24,23 +98,56 @@ def matching_clusters
     end
   end
 
-  def add_ht_item_to_freq_table(ht_item)
-    overlap = HtItemOverlap.new(ht_item)
-    overlap.matching_orgs.each do |organization|
-      @freq_table[organization.to_sym][overlap.matching_orgs.count] += 1
+  def total_hscore(member)
+    spm_total(member.to_sym) + mpm_total(member.to_sym) + ser_total(member.to_sym)
```
billdueber, line 102: Not sure I understand why "org" and "member" are being used when they are. If it's arbitrary, it's worth unifying, and if it's meaningful in some way, make it clear somewhere.

lib/cost_report.rb
```diff
@@ -24,23 +98,56 @@ def matching_clusters
...
+  end
+
+  [:spm, :ser, :mpm].each do |format|
+    # HScore for a particular format
+    define_method "#{format}_total".to_sym do |member|
+      total = 0.0
+      freq_table[member.to_sym][format].each do |num_orgs, freq|
+        total += (1.0 / num_orgs * freq)
```
billdueber, line 110: I had to read this three times. Why not just `total += freq.to_f / num_orgs`?

lib/cost_report.rb
```diff
@@ -24,23 +98,56 @@ def matching_clusters
...
+    # Costs for a particular format
+    define_method "#{format}_costs".to_sym do |member|
+      public_send("#{format}_total", member) * cost_per_volume
     end
   end
 
-  def total_hscore
-    return @total_hscore unless @total_hscore.nil?
+  def total_ic_costs(member)
+    total_hscore(member) * cost_per_volume.to_f
```
billdueber, line 122: cost_per_volume is already defined to return a float; don't need the safety `#to_f`.

lib/cost_report.rb
```diff
@@ -24,23 +98,56 @@ def matching_clusters
...
-    @freq_table.each do |org, h_scores|
-      h_scores.each do |num_orgs, freq|
-        @total_hscore[org.to_sym] += (1.0 /num_orgs * freq)
-      end
+  def extra_per_member
+    total_ic_costs(:hathitrust).to_f / Services.ht_members.members.count
+  end
+
+  def total_cost_for_member(member)
+    total_ic_costs(member.to_s) + pd_cost_for_member(member.to_s) + extra_per_member
```
billdueber, line 130: Why the `#to_s`?

lib/cost_report.rb
```diff
@@ -24,23 +98,56 @@ def matching_clusters
...
+
+  def to_tsv
+    report = []
+    report << ["member_id", "spm", "mpm", "ser", "pd", "weight", "extra", "total"].join("\t")
+    Services.ht_members.members.keys.sort.each do |member|
+      next unless organization.nil? || (member == organization.to_s)
+
+      report << [
+        member,
+        spm_costs(member),
```
billdueber, line 141: Maybe format all these as #.## on the way out? Is there a reason for the CSV to have arbitrary-precision floats in it?

billdueber (changes requested):
Overall, looks good. I have niggles with the use of `#to_sym` and `#to_s` -- either force the caller to use one, or allow the receiving methods to take either string or symbol, but no need to scatter conversions all over. I'd prefer to just accept either at the method level. 

Ditto with `#to_f`; make it part of the return type of the relevant methods. 

Finally, it's not obvious to me that any tests are dealing with a weight <> 1.0. Might be worth doing.

Sanity checking on some params is probably warranted.

lib/cost_report.rb
jsteverman: Only because arbitrary-precision is current behavior.

lib/cost_report.rb
jsteverman: Definition in my head is sum of hshares for n items. 6 of one, half dozen of the other. Will simplify.

lib/cost_report.rb
jsteverman: freq_table is a stupid doubly nested hash. 

I tried turning this:
` @freq_table = Hash.new do |hash, member|
  hash[member] = Hash.new {|fmt_hash, fmt| fmt_hash[fmt] = Hash.new(0) }
end`

into this:
`@freq_table = Hash.new do |hash, member|
    hash[member.to_sym] = Hash.new {|fmt_hash, fmt| fmt_hash[fmt.to_sym] = Hash.new(0) }
end`

which doesn't work. The alternative would be a smarter `class FreqTable` which may or may not be worth doing.

lib/cost_report.rb
jsteverman: Me neither. In theory, not every "org" is a "membe" so there is a distinction to be made, but in practice it's been pretty sloppy.

lib/cost_report.rb
jsteverman: Because mistake.

lib/cost_report.rb
mwarin: For tasks like cost reporting and scrubbing it makes sense to talk about members, since that's something we only do for actual members. 
Then there are prospective members (who want a fee estimate). 
Then there are orgs that are not members that we may still have to at least account for, like keio, kennesaw, augusta.
Just talking out loud, not offering an actual clever solution.

lib/cost_report.rb
billdueber: Are you sure? Not that it's a huge deal either way...

```ruby
f = Hash.new do |hash, member|
  hash[member.to_sym] = Hash.new do |fmt_hash, fmt|
    fmt_hash[fmt.to_sym] = Hash.new(0)
  end
end
f["bill"]["book"]["umich"] = 2

ap f
{
    :bill => {
        :book => {
            "umich" => 2
        }
    }
}
```

jsteverman: If you try incrementing f["bill"] it won't find it, creates a new hash, and assigns it to f["bill".to_sym]. I think. Anyway, the specs failed when I tried it.

lib/cost_report.rb
```diff
@@ -1,15 +1,89 @@
...
+    @maxlines = lines
+    @logger = logger
+    # Member Hash of a format hash of a member count hash
+    # { org => { ser : { 1 org : count, 2 org : count }, mpm : {...
+    @freq_table = Hash.new do |hash, member|
+      hash[member] = Hash.new {|fmt_hash, fmt| fmt_hash[fmt] = Hash.new(0) }
+    end
+  end
+
+  def num_volumes
```
billdueber, line 23: Just raising here the idea that at some point we should extract all these mongo queries into methods that can be reused throughout the codebase. Mongoid queries support composition, so they'll come in handy.

billdueber (approved):
This all looks good to me. I have _not_ stepped through the actual logic, just the code, so we're counting on the tests to make sure it works the way it's supposed to. Adding more tests once we get a common set of test data and start doing integration testing is warranted.

aelkiss (approved):

PR #55, opened 2020-10-23
HT-2366, HT-2705: cleanup for loading all data

aelkiss:
- script to delete OCN resolutions (for loading delta)
- handle issn field in holdings


jsteverman (approved):
No objections.

PR #57, opened 2020-10-27
Allow skipping header line

billdueber:
Allow FileLoader to skip a header line, identified by regexp

FileLoad.load now takes an optional keyword argument
`skip_header_match`, a regex used to identify a line of
column headers (e.g.,` /\AOLN/`). If found, the first line
will be skipped prior to processing the data.

Added code and tests, and changed `/bin/add_print_holdings` to use it


aelkiss (approved):
Looks good. Currently I don't think the hathifiles and OCN resolution tables have headers, but this gives us a mechanism to easily skip a header.

PR #59, opened 2020-10-30
HT-2638: case for conditions codes

aelkiss:
The existing data all uses uppercase for condition codes, and we did not normalize it to lowercase prior to loading. Not sure if this is the right solution, but it is a solution.


billdueber:
If everyone uses uppercase, then this'll work 'til it doesn't.  Merging.

PR #61, opened 2020-11-05
Add small but complete set of test data for SPMs

billdueber:
This adds a small set of inter-related data, with fake
hathifile, OCLC concordance, and a script to generate
fake holdings with characteristics useful for integration
testing. The latter exercise the concordance, shared and
unshared items, items not in the hathifiles, repeated
items, etc.  See `generate_mono.rb` for comments
describing all the effects.


aelkiss (approved):
Looks good. I see the promise here in generating integration tests in that we can start from (more or less) a scenario with a particular expected output, then generate the data that should generate that output.

PR #62, opened 2020-11-12
HT-2717 improve perf of Cost Report generation

jsteverman:
- compiles ec=>org and org=>ec data structures in Cluster objects
  - removes dependence on ClusterOverlap and overlap records for cost report generation
  - simple conditional in HtItemOverlap for computing item/holding overlap
Processes ~740 access:deny items per second so roughly a 4 hour run of 10m items.   

ETAS and the other overlap report likely have performance problems due to its reliance on:
 * ClusterOverlap
 * MultiPartOverlap
* SinglePartOverlap
* SerialOverlap

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -44,7 +44,7 @@ Metrics/BlockLength:
     - 'spec/support/**/*.rb'
     - 'spec/**/*_spec.rb'
 
-# Offense count: 4
+# Offense count: 5
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
   Exclude:
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -52,6 +52,7 @@ Metrics/ClassLength:
     - 'lib/enum_chron_parser.rb'
     - 'lib/scrub_fields.rb'
     - 'lib/cost_report.rb'
+    - 'lib/cluster.rb'
 
 # Offense count: 6
 # Configuration parameters: CountComments, ExcludedMethods.
--- a/lib/cluster.rb
+++ b/lib/cluster.rb
@@ -5,6 +5,7 @@ require "holding"
 require "ht_item"
 require "commitment"
 require "ocn_resolution"
+require "calculate_format"
 require "cluster_ht_item"
 require "cluster_error"
 
@@ -127,21 +128,23 @@ class Cluster
 
   # Maps enumchrons to list of orgs that have an item with that enumchron
   # Not necessary for SPMs/SERs
-  def item_enum_chron_orgs(enum)
+  def item_enum_chron_orgs
     @item_enum_chron_orgs ||= ht_items.group_by(&:n_enum)
       .transform_values {|htitems| htitems.map(&:billing_entity) }
-    @item_enum_chron_orgs[enum] || []
+      .tap {|h| h.default = [] }
+    @item_enum_chron_orgs
   end
 
   # Maps enumchrons to list of orgs that have a holding with that enumchron
-  def holding_enum_chron_orgs(enum)
+  def holding_enum_chron_orgs
     @holding_enum_chron_orgs ||= holdings.group_by(&:n_enum)
       .transform_values {|holdings| holdings.map(&:organization) }
-    @holding_enum_chron_orgs[enum] || []
+      .tap {|h| h.default = [] }
+    @holding_enum_chron_orgs
   end
 
-  def enum_chron_orgs(enum_chron)
-    (item_enum_chron_orgs(enum_chron) + holding_enum_chron_orgs(enum_chron)).uniq
+  def enum_chron_orgs(enum)
+    (item_enum_chron_orgs[enum] + holding_enum_chron_orgs[enum]).uniq
   end
 
   def billing_entities
--- a/lib/cluster.rb
+++ b/lib/cluster.rb
@@ -127,25 +127,21 @@ class Cluster
 
   # Maps enumchrons to list of orgs that have an item with that enumchron
   # Not necessary for SPMs/SERs
-  def item_enum_chron_orgs
-    return @item_enum_chron_orgs if @item_enum_chron_orgs
-
-    @item_enum_chron_orgs = Hash.new {|h, k| h[k] = [] }
-    ht_items.each {|ht| @item_enum_chron_orgs[ht.n_enum] << ht.billing_entity }
-    @item_enum_chron_orgs
+  def item_enum_chron_orgs(enum)
+    @item_enum_chron_orgs ||= ht_items.group_by(&:n_enum)
+      .transform_values {|htitems| htitems.map(&:billing_entity) }
+    @item_enum_chron_orgs[enum] || []
   end
 
   # Maps enumchrons to list of orgs that have a holding with that enumchron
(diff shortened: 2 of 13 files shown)


lib/cluster.rb
```diff
@@ -120,6 +120,38 @@ def format
...
+  end
+
+  # Maps enumchrons to list of orgs that have an item with that enumchron
+  # Not necessary for SPMs/SERs
+  def item_enum_chron_orgs
+    return @item_enum_chron_orgs if @item_enum_chron_orgs
+
+    @item_enum_chron_orgs = Hash.new {|h, k| h[k] = [] }
+    ht_items.each {|ht| @item_enum_chron_orgs[ht.n_enum] << ht.billing_entity }
+    @item_enum_chron_orgs
```
aelkiss, line 135: Could write as:

```ruby
@item_enum_chron_orgs ||= ht_items.group_by { |ht| ht.n_enum }
  .transform_values { |htitems| htitems.map(&:billing_entity) }
```

I bet @billdueber  might even have a more concise way to do this...

lib/cluster.rb
```diff
@@ -120,6 +120,38 @@ def format
...
+  end
+
+  # Maps enumchrons to list of orgs that have a holding with that enumchron
+  def holding_enum_chron_orgs
+    return @holding_enum_chron_orgs if @holding_enum_chron_orgs
+
+    @holding_enum_chron_orgs = Hash.new {|h, k| h[k] = [] }
```
aelkiss, line 142: As above, could use `group_by`

aelkiss (commented):
Looks pretty good. I left a suggestion for making `item_enum_chron_orgs` more concise, but what you have certainly works.

lib/cluster_overlap.rb
```diff
@@ -31,7 +31,7 @@ def each
   end
 
   def overlap_record(ht_item, org)
-    case ht_item._parent.format
+    case @cluster.format
     when "ser"
       SerialOverlap.new(@cluster, org, ht_item)
     when "spm"
```
mwarin, line 37: Is this a #todo, or is there something going on here that I don't get?

lib/ht_item_overlap.rb
```diff
@@ -16,11 +16,13 @@ def initialize(ht_item)
 
   # Find all organization with holdings that match the given ht_item
   def organizations_with_holdings
```
mwarin, line 18: I do not understand this.

spec/cluster_overlap_spec.rb
```diff
@@ -42,13 +42,6 @@
     end
   end
 
-  describe "#organization_in_cluster" do
```
mwarin, line 45: why was this test taken out? no longer relevant? should be replaced with another test?

mwarin (approved):
Could do with more comments, but based on my understanding of what is going on I have no issue with these changes.

spec/cluster_overlap_spec.rb
jsteverman: Duplicative with the `def organizations_in_cluster` found in lib/cluster.rb and its spec.

.rubocop_todo.yml
```diff
@@ -44,14 +44,15 @@ Metrics/BlockLength:
...
-# Offense count: 4
+# Offense count: 5
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
   Exclude:
     - 'lib/autoscrub.rb'
     - 'lib/enum_chron_parser.rb'
     - 'lib/scrub_fields.rb'
     - 'lib/cost_report.rb'
+    - 'lib/cluster.rb'
```
aelkiss, line 55: Is there something we could pull out of `Cluster`?

aelkiss: Doesn't need to be now, but maybe something to keep in mind in the future.

lib/ht_item_overlap.rb
aelkiss: Consider extracting conditions as methods with intention-revealing names?

lib/cluster.rb
```diff
@@ -120,6 +120,34 @@ def format
...
   end
 
+  def organizations_in_cluster
+    @organizations_in_cluster ||= (holdings.pluck(:organization) +
+                                  ht_items.pluck(:billing_entity)).uniq
+  end
+
+  # Maps enumchrons to list of orgs that have an item with that enumchron
+  # Not necessary for SPMs/SERs
+  def item_enum_chron_orgs(enum)
```
billdueber, line 130: I'm worried that by caching the values (via `||=`) in all these organization-counting methods we'll be ignoring any additions to the cluster. Might not be a real issue if we're never doing both adds and analysis at the same time, but thought I'd bring it up as at least a candidate for documenting it.

```ruby
cluster.organizations_in_cluster
#=> ['anu', 'bc', 'cmu']
cluster.item_enum_chron_orgs(enum).include?('umich')
#=> false
cluster.add_ht_items(items_from_umich)
cluster.organizations_in_cluster
#=> still ['anu', 'bc', 'cmu']
cluster.item_enum_chron_orgs(enum).include?('umich')
#=> still always gonna be false
```

This potentially drags us down the garden path of dirty-record tracking, which no one wants. It could be mitigated by either zeroing out the cached values (like `@item_enum_chron_orgs`) in `add_ht_items`, or by just documenting and being careful. The former could be made easier if all these computed values were kept in a single data structure which could then be unassigned, making sure we don't miss any. Or maybe there's an `HtItemSet` class in here waiting to be defined, but that seems like it wouldn't mesh super well with all the Mongoid niceties around embedded docs.

I'd also set `*_enum_chron_orgs` to  `Hash.new{|k,h| h[k] = []}` in the initializer, instead of dragging that default-value-is-a-hash logic into each method, or maybe use a dead-simple hash-of-arrays class to make sure that you never get a scalar in there:

```ruby
class HashOfArrays < SimpleDelegator
  def initialize
    @h = Hash.new{|h,k| h[k] = []}
    self.__setobj__(@h)
  end
  def []=(k,v)
    @h[k] = [v]
  end
end

h = HashOfArrays.new
#=> {}
h[1] #=> []
h[2] << :two 
h[2] << :too 
h[3] = :three

 pp h
#=> {1=>[], 2=>[:two, :too], 3=>[:three]}`

```

jsteverman: I *think* running cost report generation and record insertion at the same is mostly a non-starter for anything but guesstimation. Whether a cluster is altered in the midst of h_share calculation or 5 ms after is meaningless. If we are running a holdings update, we can't get any more precise than "cost report generated while in the middle of a holdings update" with either approach.

jsteverman:
Last push took Aaron's suggestions for the enum=>holding mappings and added a .tap{|h| h.default = []} to keep the dreaded nils away.
I cleaned up organizations_with_holdings and added a few comments per Martin's question. That's going to get messed with yet again by 2726 and 2727 so I am comfortable leaving imperfect.

Cost Reports uses HtItemOverlap and a bit of MultiPartOverlap. ETAS reports use ClusterOverlap and the MultiPart/Serial/SinglePartOverlap. Basically it's a rat's nest of inheritance and almost-not-quite duplicative behavior. I would like to nail down the spec and shameless solution in 2726 and 2727, and then have a separate issue for refactoring the embarrassment away.

PR #65, opened 2020-12-17
Prevent duplicate holdings in large clusters

jsteverman:
* Services.large_clusters is little more than a list of OCNs
 * ClusterHolding.update prevents cluster.large? from having more than 1 holding per member
 * Does NOT resolve cases where there are already multiple holdings per member
 * Does NOT prevent duplication in merged clusters
 * CalculateFormat returns "ser" when cluster.large?

### DIFF
--- a/lib/cluster_getter.rb
+++ b/lib/cluster_getter.rb
@@ -50,6 +50,11 @@ class ClusterGetter
         raise ClusterError, "clusters disappeared, try again" if source.nil?
         next if source._id == @target._id
 
+        if source.large? ^ @target.large?
+          warn("Merging into a large cluster. " \
+                "OCNs: [#{source.ocns.join(",")}] and [#{@target.ocns.join(",")}]")
+        end
+
         source.delete
         @target.add_members_from(source)
         @target.add_to_set(ocns: source.ocns)
--- a/lib/cluster_getter.rb
+++ b/lib/cluster_getter.rb
@@ -50,6 +50,7 @@ class ClusterGetter
         raise ClusterError, "clusters disappeared, try again" if source.nil?
         next if source._id == @target._id
 
+        # Boolean xor. One and only one is "large"
         if source.large? ^ @target.large?
           warn("Merging into a large cluster. " \
                 "OCNs: [#{source.ocns.join(",")}] and [#{@target.ocns.join(",")}]")
--- a/lib/cluster_holding.rb
+++ b/lib/cluster_holding.rb
@@ -22,13 +22,9 @@ class ClusterHolding
     getter.get do |c|
       to_add = []
 
+      common_uuids = check_for_duplicate_uuids!(c.holdings.to_a, @holdings)
       @holdings.each do |holding|
-        if (existing = c.holdings.to_a.find {|h| h.uuid == holding.uuid })
-          next if existing.same_as?(holding)
-
-          raise "Found holding #{existing.inspect} with same UUID " \
-            "but different attributes from update #{holding.inspect}"
-        end
+        next if common_uuids.include? holding.uuid
 
         old_holdings = find_old_holdings(c, holding)
 
@@ -91,6 +87,21 @@ class ClusterHolding
     cluster.large? &&
       (cluster.holdings.to_a.find {|h| h.organization == holding.organization } ||
        to_add.find {|h| h.organization == holding.organization })
+  end
+
+  # Check and see if any items across existing/new holdings share a UUID but no other attributes.
+  # Assumes uuids will not be duplicated within either list
+  # Common case is that there is no overlap
+  def check_for_duplicate_uuids!(existing_holdings, new_holdings)
+    common_uuids = existing_holdings.map(&:uuid) & new_holdings.map(&:uuid) # set intersection
+    common_uuids.each do |uuid|
+      existing = existing_holdings.find {|h| h.uuid == uuid }
+      holding = new_holdings.find {|h| h.uuid == uuid }
+      unless existing.same_as?(holding)
+        raise "Found holding #{existing.inspect} with same UUID " \
+                "but different attributes from update #{holding.inspect}"
+      end
+    end
   end
 
 end
--- a/lib/cluster_holding.rb
+++ b/lib/cluster_holding.rb
@@ -17,14 +17,8 @@ class ClusterHolding
     raise ArgumentError, "Holding must have exactly one OCN" if @ocn.nil?
   end
 
-  def cluster(getter: ClusterGetter.new([@ocn]))
-    getter.get do |cluster|
-      cluster.add_holdings(@holdings)
-    end
-  end
-
   # Updates a matching holding or adds it
-  def update(getter: ClusterGetter.new([@ocn]))
+  def cluster(getter: ClusterGetter.new([@ocn]))
     getter.get do |c|
       to_add = []
 
@@ -50,6 +44,7 @@ class ClusterHolding
       c.add_holdings(to_add)
     end
   end
+  alias_method :update, :cluster
 
   def delete
     raise ArgumentError, "Can only delete one holding at a time" unless @holdings.length == 1
--- a/lib/cluster_holding.rb
+++ b/lib/cluster_holding.rb
@@ -44,7 +44,6 @@ class ClusterHolding
       c.add_holdings(to_add)
     end
   end
-  alias_method :update, :cluster
 
   def delete
     raise ArgumentError, "Can only delete one holding at a time" unless @holdings.length == 1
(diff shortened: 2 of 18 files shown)


spec/cluster_holding_spec.rb
```diff
@@ -78,6 +78,20 @@ def new_submission(holding, date: Date.today)
         eq(h2.date_received)
     end
 
+    it "updates multiple existing holdings if cluster.large?" do
+      c.ocns << Services.large_clusters.ocns.first
+      c.save
+      described_class.new(h).cluster
+      described_class.new(h.clone).cluster
```
aelkiss, line 85: This shouldn't add an additional holding, right?

spec/cluster_holding_spec.rb
```diff
@@ -78,6 +78,20 @@ def new_submission(holding, date: Date.today)
...
+      h2.date_received = Date.today
+      h2.uuid = SecureRandom.uuid
+      described_class.new(h2).update
+      cluster = Cluster.first
+      expect(cluster.holdings.first.date_received).to eq(h2.date_received)
```
aelkiss, line 90: This spec confuses me because it doesn't seem to take into account that large clusters should skip duplicative holdings. There's only one holding to update. Maybe this is an artifact of the order the tests were written in? If nothing else, it might be worth a comment or two to explain that the observed behavior is a side effect of only retaining one holding record.

spec/cluster_holding_spec.rb
```diff
@@ -117,6 +131,29 @@ def new_submission(holding, date: Date.today)
       h2 = build(:holding, ocn: h.ocn, uuid: h.uuid)
       expect { described_class.new(h2).update }.to raise_exception(/same UUID/)
     end
+
+    it 'skips duplicative holdings for a member in "large clusters"' do
```
aelkiss, line 135: Might be worth organizing the new specs like
```ruby
describe '#update' do
  context 'with a "large cluster"' do
    it 'updates multiple existing holdings'
    it 'skips duplicative holdings'
    ...
  end
end
```
```

File: spec/cluster_holding_spec.rb [hunk 4]
```diff
@@ -117,6 +131,29 @@ def new_submission(holding, date: Date.today)
...
+      c.ocns = [Services.large_clusters.ocns.first]
+      c.save
+      dupes = [build(:holding, organization: "umich", ocn: c.ocns.first),
+               build(:holding, organization: "umich", ocn: c.ocns.first)]
+      described_class.new(dupes).update
+      cluster = Cluster.first
+      expect(cluster.holdings.count).to eq(1)
+    end
+
+    it 'updates the date_received for "large clusters"' do
```
aelkiss, line 145: Does this test something different from the "updates multiple existing holdings if cluster.large?" spec I commented on above?

spec/large_clusters_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      expect(Cluster.first.ht_items.count).to eq(2)
+      ht2.ocns = [1]
+      expect do
+        ClusterHtItem.new(ht2).cluster
+      end.to raise_exception(LargeClusterError)
+    end
+  end
+
+  describe "merging of large clusters" do
+    # Merging a non-"large cluster" into a "large cluster" means future holdings will be deduped.
```
aelkiss, line 58: I think this behavior is implied from the fact that we recluster by reloading constituent data and the update spec for deduping above. I think it's probably worth a spec to verify this behavior given the desire to have this comment and that it shouldn't need much additional setup beyond what's here -- i.e. create a non-large cluster, give it two holdings that should be collapsed, merge it into an empty large cluster, verify the holdings were collapsed.

spec/cluster_holding_spec.rb
jsteverman: Maybe? `.cluster` is little more than `cluster.add_holdings` which is `.push_to_field`. 
`.update` performs the check for existing matching holdings and, with this PR, large clusters. 

Should `.cluster` check for one or both of those special cases? Should `.update` really just be renamed `.cluster`?   

bin/add_print_holdings.rb requires a -u flag for updates. Other than the initial load, I can't imagine a use case for `.cluster` as written.

spec/cluster_holding_spec.rb
jsteverman: See above.

spec/cluster_holding_spec.rb
jsteverman: Previous test was both existing dupes get updated by a single holding. This test is a single holding being updated by a batch of dupes. The exact motivation escapes me, but I assume I managed to screw this up at some point along the way.

spec/large_clusters_spec.rb
jsteverman: As far as I can tell, a cluster merge doesn't perform a recluster. It performs `add_members_from` which is `push_to_field`. Holdings won't be de-duped on merge, only future holdings.

aelkiss: hmm, ok. Not sure that's the behavior we want. Let's discuss Friday.

spec/cluster_holding_spec.rb
aelkiss: OK. I was assuming that large cluster should always imply no dupes. What are the consequences of potentially ending up with dupes in a large cluster? It shouldn't matter for ETAS overlap reports, since we force copy_count to 1 for that; does it make a difference for cost calculation? I suppose we could have a "dedupe" function to clean up large clusters post-merge/update. Again, let's discuss Friday.

spec/cluster_holding_spec.rb
aelkiss: Let's get rid of `#cluster` (or just alias it to `#update`)

spec/cluster_holding_spec.rb
aelkiss: It might be worth a spec to verify that multiple reported holdings for a serial do not affect cost calculation.

aelkiss: We already have a spec for the cost report that multiple holdings do not affect h-share: https://github.com/hathitrust/holdings-backend/blob/master/spec/cost_report_spec.rb#L222

So I think we can go ahead and merge this as-is.

spec/large_clusters_spec.rb
aelkiss: Discussed and we are OK with some potential duplicates since it doesn't affect reports and simplifies implementation.

lib/cluster_holding.rb
```diff
@@ -76,4 +71,27 @@ def self.delete_old_holdings(org, date)
...
 
+  private
+
+  def update_holding(old, new)
+    old.update_attributes(date_received: new.date_received,
+                          uuid: new.uuid)
+  end
+
+  def find_old_holdings(cluster, holding)
+    if cluster.large?
```
aelkiss, line 82: The parallel conditionals here might suggest that there should be a subclass that knows how to deal with large clusters, but I'm OK with this for now. If we start adding significantly more support around large clusters, we should perhaps rethink it.

aelkiss (approved):
Sorry I didn't get a chance to look at this today. Changes all look good to me.

billdueber (changes requested):
Nothing big here -- logic looks sound as near as I can tell and I didn't see any obvious important edge cases being missed. Little cleanup and it's good to go.

lib/cluster_holding.rb
```diff
@@ -17,14 +17,8 @@ def initialize(*holdings)
     raise ArgumentError, "Holding must have exactly one OCN" if @ocn.nil?
```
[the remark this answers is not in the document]
jsteverman: It should error out, no?

lib/cluster_holding.rb
```diff
@@ -36,12 +30,12 @@ def update(getter: ClusterGetter.new([@ocn]))
             "but different attributes from update #{holding.inspect}"
```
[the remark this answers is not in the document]
jsteverman: Added nearly verbatim.

lib/cluster_holding.rb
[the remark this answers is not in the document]
billdueber: Yeah, sorry, was thinking it should have a gate that produces a more useful error message than "NoMethod ocn on nil", parallel to the other errors.

lib/cluster_getter.rb
```diff
@@ -50,6 +50,11 @@ def merge
         raise ClusterError, "clusters disappeared, try again" if source.nil?
         next if source._id == @target._id
 
+        if source.large? ^ @target.large?
```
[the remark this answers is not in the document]
jsteverman: Commented it. Was gonna blame this on rubocop being clever until I checked my search history. I totally had to ddg ruby xor to figure this one out (twice).

lib/cluster_holding.rb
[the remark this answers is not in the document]
jsteverman: `raise ArgumentError, "Must have holdings to cluster" unless @holdings.any?

PR #66, opened 2020-12-18
HT-2752, HT-2753: Draft ADRs for performance & consistency

aelkiss:
First attempt at drafting architecture decision records for performance related improvements from HT-2480, HT-2491, HT-2463, HT-2591, HT-2590 (see #45)


jsteverman (approved):
This is all clear to me and could be merged as-is. 

Moving forward, as we touch code we should probably reference the relevant ADRs.

PR #67, opened 2020-12-18
HT-2754 - automatically publish docker image

aelkiss:
Cribbed from https://github.com/mlibrary/patron_account/blob/main/.github/workflows

@daaang Is the plan to at some point factor this out to actions shared under the "phineas" brand? As is, I'm a little worried about the potential for meaningless arbitrary variation between repositorie and/or a bunch of stuff that nobody fully understands or commits to owning (à la the capistrano installs), or is this relatively minimal enough that the duplication across repositories isn't a big deal? Certainly https://github.com/mlibrary/patron_account/blob/main/.github/workflows/deploy-testing.yaml at least seems like it would benefit from abstraction (not duplicating it here since there's no deployment as such to roll out yet..)

### DIFF
--- a/.github/workflows/build-master.yml
+++ b/.github/workflows/build-master.yml
@@ -1,0 +1,34 @@
+name: Run CI
+
+on:
+  push:
+    branches:
+      - $default_branch
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up tests
+        run: ./bin/setup_test.sh
+
+      - name: Run tests
+        run: docker-compose run --rm -e MONGOID_ENV=test dev bin/wait-for mariadb:3306 -- bundle exec rspec
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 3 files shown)


aelkiss:
I need to merge this to master to see if the workflows work but @daaang  please still feel free to comment

daaang:
Yes, specifically the deploy action is the main one we feel we should have covered (it's https://github.com/mlibrary/phineas-deploy but currently we haven't done anything except create the repo and look around).

My intention is that the workflows will end up looking something like this

```yaml
steps:
- if: ${{ github.event_name == 'workflow_run' }}
  uses: mlibrary/phineas-deploy-action@v1
  with:
    image: hathitrust/myapp-unstable:${{ github.sha }}
    cluster-address: https://macc.kubernetes.hathitrust.org
    namespace: myapp-testing
    cluster-ca: ${{ MACC_CLUSTER_CA }}
    token: ${{ MACC_TESTING_TOKEN }}
- if: ${{ github.event_name == 'workflow_dispatch' }}
  uses: mlibrary/phineas-deploy-action@v1
  with:
    repo: hathitrust/myapp-unstable
    commitish: ${{ github.event.inputs.tag }}
    cluster-address: https://macc.kubernetes.hathitrust.org
    namespace: myapp-testing
    cluster-ca: ${{ MACC_CLUSTER_CA }}
    token: ${{ MACC_TESTING_TOKEN }}
```

In the meantime, I don't want to slow anybody down, so, I guess, "Lots Of Copies Keeps Stuff Safe"

PR #68, opened 2020-12-18
Remove deprecated methods from Cluster

jsteverman:
The following methods in the Cluster class were untested and unused.
  Likely remnants of approaches discarded or refactored.
 * merge
 * merge_many
 * self.merge_many
 * add_ocns
 * billing_entities


aelkiss (approved):
Yup, looks like the `#merge`/`#merge_many` and `#add_ocns` was all superceded by `ClusterGetter` in #50.

PR #71, opened 2021-02-05
Ht 2792 reclustering

jsteverman:


lib/cluster_ht_item.rb
```diff
@@ -67,19 +67,25 @@ def update_or_add_ht_items(cluster)
...
 
-    unless to_append.empty?
-      cluster.add_ht_items(to_append)
-    end
+    cluster.add_ht_items(to_append) unless to_append.empty?
+
+    Reclusterer.new(cluster).recluster if needs_reclustering
+  end
+
+  def current_is_subset_of_new(current_ocns, new_ocns)
```
mwarin, line 87: Won't this also, unnecessarily, trigger if current_ocns == new_ocns?
(bacause [1,2] is a subset of [1,2])

jsteverman: needs_reclustering is false unless current_is_subset_of_new

PR #74, opened 2021-03-15
Members must have inst_id and weight

jsteverman:


aelkiss (approved):
I don't think there's functionality for persisting these, but this will at least make sure that null weights/ids don't mess something up elsewhere. Longer term we need to think about how to avoid having multiple models in different code bases for the same underlying databases, or at the least making it more clear what components own specific database tables. Longer-term that probably means having an API for access to things like institutions, etc, and a shared gem that handles serialization/deserialization

PR #75, opened 2021-04-01
Calc cost rpt on item format not cluster format

jsteverman:
* also makes explicit in spec that shared bib_id not cluster is req'd for mpm item format contagion.


jsteverman:
I discovered that current practice only uses `n_enum` to determine MPM status. I considered changing the new process to match, but instead settled for n_enum_chron instead of enum_chron.

aelkiss (approved):
:+1:

PR #76, opened 2021-04-08
HT-2904: remove enum_chron from format algo

jsteverman:


spec/calculate_format_spec.rb
```diff
@@ -20,21 +20,32 @@ def add_serial(ht_item)
...
-    it "is a MPM if it or another item has enum_chron" do
+    it "is a MPM if it or another item has enum" do
       c_mpm = ClusterHtItem.new(ht_mpm).cluster
       expect(
         described_class.new(c_mpm).item_format(ht_mpm)
       ).to eq("mpm")
     end
 
-    it "is a MPM if another ht item on this record has enum_chron" do
+    it "is NOT an MPM if it has a chron but no enum" do
```
aelkiss, line 30: Rubocop complains about the expectations here.

You could break this up with a context and multiple examples, or (probably preferably) you could get rid of:

```ruby
      expect(ht_chron.n_enum).to be("")
      expect(ht_chron.n_chron).to eq("1994")
```
I think that's a worthwhile test, but it would be better to put it in a test for `EnumChronParser`.

jsteverman: I was already in the middle of fixing it. I just kept the last expectation which is the important one.

aelkiss (approved):

PR #77, opened 2021-04-13
HT-2898: Format and Overlap determination documentation

jsteverman:


format_and_overlap_determination.md
```diff
@@ -0,0 +1,118 @@
...
+: mono (spm), multi (mpm), serial (ser)
+
+**Cluster**
+: Collection of HT Items, Holdings, and Concordance pairs united by a list of OCNs determined to be equivalent.
+
+**Bib Record**
+: A preferred record from the Zephir clustering system containing one or more HT Items. Has a single Hathitrust record ID.
+
+**HT Item**
+: Volume ID and associated metadata taken from the Hathifiles.
```
nfulkers, line 15: Add link to HathiFiles description page on HT website

format_and_overlap_determination.md
```diff
@@ -0,0 +1,118 @@
+# Description of Format and Overlap Determinations #
+This attempts to describe what this system actually does. This does not attempt to exhaustively describe all corner cases or emergent properties.
+
+## Definitions ##
+**Format**
+: mono (spm), multi (mpm), serial (ser)
```
nfulkers, line 6: Is it implied from this set-like notation that an item can only have one format?

Perhaps add that items can have only one value, but within a cluster items can have different values?

format_and_overlap_determination.md
```diff
@@ -0,0 +1,118 @@
...
+
+**Cluster**
+: Collection of HT Items, Holdings, and Concordance pairs united by a list of OCNs determined to be equivalent.
```
nfulkers, line 9: specify which system we're talking about, for example the very next definition (Bib record) refers to Zephir clustering.  If we're talking about both Zephir and PH for purposes of this definition, it might be worth saying that the logic/process for determining what is equivalent varies between systems.

format_and_overlap_determination.md
```diff
@@ -0,0 +1,118 @@
...
+
+**Bib Record**
+: A preferred record from the Zephir clustering system containing one or more HT Items. Has a single Hathitrust record ID.
+
+**HT Item**
+: Volume ID and associated metadata taken from the Hathifiles. 
+
+**Print Serials List**
+: List of Record Ids determined to be serials by some unknown UMich process. List is limited to those held by UMich.
```
nfulkers, line 18: How do we get this list?  How often?

format_and_overlap_determination.md
```diff
@@ -0,0 +1,118 @@
...
+
+**Large Clusters**
+: Clusters large enough that they create implementation problems. Identified by an OCN, there are currently 11 on the list.
+
+**Billing Entity**
+: The `billing_entity` field taken from the `ht_collections` table associated with a particular HT Item through the collection code. Used for cost allocation when holdings are incomplete or no clustering can be performed. Also has subtle and likely inconsequential implications for items contributed by UCM and KEIO.
```
nfulkers, line 24: define collection code?  What does it mean when holdings are incomplete of clustering can't be performed?

format_and_overlap_determination.md
```diff
@@ -0,0 +1,118 @@
...
+
+**n_enum**
+: A normalized enumeration. Description fields are processed for enumeration and chronology. Only enumeration ends up being used.
```
nfulkers, line 27: "Description fields are processed" from where?

format_and_overlap_determination.md
```diff
@@ -0,0 +1,118 @@
...
+
+**Billing Entity**
+: The `billing_entity` field taken from the `ht_collections` table associated with a particular HT Item through the collection code. Used for cost allocation when holdings are incomplete or no clustering can be performed. Also has subtle and likely inconsequential implications for items contributed by UCM and KEIO.
+
+**n_enum**
+: A normalized enumeration. Description fields are processed for enumeration and chronology. Only enumeration ends up being used. 
+
+
+## Item Format ##
+
```
nfulkers, line 31: Is it worth saying explicitly here why item format matters?  Same for Cluster Format below.

format_and_overlap_determination.md
```diff
@@ -0,0 +1,118 @@
...
+## Cost Allocation ##
+
+**Target Cost**
+: Set by the Board and Executive Director. Portion of the budget to be paid for by item allocations.
+
+**Cost Per Volume**
+: Target Cost / Total Number of HT Items
+
+**Member Weight**
```
nfulkers, line 66: Is this the same as the "tier"?  If not we should add what a tier is.

format_and_overlap_determination.md
```diff
@@ -0,0 +1,118 @@
...
+**Public Domain Cost**
+: Cost Per Volume * Number of Public Domain Volumes (see Public Domain vs In-Copyright)
+
+**Public Domain Cost for Member**
+: Total Public Domain Cost / Total Weight * Member Weight
+
+**Extra Per Member**
+: The IC Costs assigned to Hathitrust / Number of members found in the `ht_billing_members` table that aren't Hathitrust. Hathitrust has IC costs due to being the billing entity for some items.
+
+**HScore**
```
nfulkers, line 81: I might expand on this - the share is the inverse of the number of members who report holding the item

format_and_overlap_determination.md
jsteverman: We are not talking about Zephir for the purposes of this definition.

format_and_overlap_determination.md
jsteverman: Now: Billing Entity is used for cost allocation when a member's submitted holdings do not include all of their deposited HT Items, or no clustering can be performed due to a lack of OCNs.

format_and_overlap_determination.md
```diff
@@ -0,0 +1,118 @@
...
+**SER / SPM**
+: Is not MPM. One or more of the items in the cluster are SPM _and_ one or more of the items in the cluster are SER.
+
+**SER**
+: Is not MPM or SER/SPM. Any of the items in the cluster are SER (and by implication ALL are SER).
+
+**SPM**
+: Is not MPM, SER/SPM, or SER.
+
+## Cost Allocation ##
```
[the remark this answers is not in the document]
jsteverman: There is no unique processing for FLBOG or the Georgia State system in this repository.

format_and_overlap_determination.md
jsteverman: It's the same thing.

format_and_overlap_determination.md
```diff
@@ -0,0 +1,118 @@
...
+: Sum of HScores for Member for each of SPM, MPM and SER
+
+**Total IC Cost for Member**
+: Total HScore for Member * Cost Per Volume
+
+**Total Cost for Member**
+: Total IC Cost for Member + Total PD Cost for Member + Extra Per Member
+
+
+### Public Domain vs In-Copyright ###
```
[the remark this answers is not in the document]
jsteverman: Moved to the first definition in the Cost Allocation section.

format_and_overlap_determination.md
```diff
@@ -0,0 +1,118 @@
...
+For cost allocation purposes, items with access "allow" in the Hathifiles are considered "public domain" and access "deny" are considered "in copyright". Rationale for the "allow" and "deny" found in the Hathifiles is underdefined. 
+
+## Item Overlap ##
+Determining who holds a particular IC item.
+
+For items in **clusters** that are not MPM, all organizations with a holding in the cluster and the billing entity for the item are allocated a share. 
+
+For items in **clusters** that are MPM, the process is more complicated. Organizations with holdings:
```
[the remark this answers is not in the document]
jsteverman: A topic of conversation for our weekly meeting.

format_and_overlap_determination.md
```diff
@@ -0,0 +1,118 @@
...
+- Organizations with holdings with empty **n_enum**. The reverse is **not** true; i.e. Items with empty n_enum don't match everything.
+- Organizations with holdings with an **n_enum** matching the **n_enum** of the Item.
+- Organizations with **holdings** in the cluster with **n_enum** that don't match any of the **n_enum** found in any of the Items in the cluster. [^2]
+- Billing entity for the item.
+
+**NB**: Billing entities apply only to the Item they are on. For example, a billing entity on an HT Item with an empty n_enum will **not** match other items in the cluster.
+
+## Frequency Table ##
+Item shares are held in a frequency table separated by member and **item** format.[^3] This gets compiled into per format per member cost allocations.
```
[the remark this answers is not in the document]
jsteverman: "HScore" is terminology taken from the existing system, and it's a little awkward when "share" seems to be perfectly adequate. Changed to HScore all the same.

format_and_overlap_determination.md
jsteverman: Hathifiles and submitted holdings. Added.

format_and_overlap_determination.md
jsteverman: Done

nfulkers: Is "calculate" really right here?  Like, is it math?  Or would something like "assign" capture what happens?  Sorry to nitpick.

format_and_overlap_determination.md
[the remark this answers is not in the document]
nfulkers: does that mean we need the system to be doing something that it currently doesn't do?

format_and_overlap_determination.md
jsteverman: According to wikipedia, "To calculate means to determine mathematically in the case of a number or amount, or in the case of an abstract problem to deduce the answer using logic, reason or common sense." In this case it also matches the terminology used in the code.

nfulkers: okie dokie

format_and_overlap_determination.md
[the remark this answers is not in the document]
jsteverman: I don't think so. I think the system handles those appropriately. If there are some separate requirements documents for those systems we can walk through the logic and check.

format_and_overlap_determination.md
mwarin: Margaret Kelly (megrust@umich.edu) emails this as a .txt file to Martin & Josh at the end of each month. We could request instead that it be uploaded to a certain dropbox location.

format_and_overlap_determination.md
mwarin: True, and this is mentioned below under **Cluster Format** and **HT Item Format**.

format_and_overlap_determination.md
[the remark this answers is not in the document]
mwarin: The old system does not do anything different for these. 
From its point of view they are 2 normal members.

format_and_overlap_determination.md
mwarin: Well. The **Member Tier** is a label that exists outside the system. 

Melissa will tell me "Please make an estimate for member xyz, they are a US institution and tier 1". She determines tier based on the IPEDS. From her boilerplate that goes out to prospective members: "Each member is assigned to one of three tiers based on the member's Total Library Expenditures as reported in public sources."

Each tier (1,2,3) has a weight associated. This association is not codified, other than perhaps in Confluence, but as far as I'm concerned lives in our minds.
Tier 1: 0.67
Tier 2: 1.00
Tier 3: 1.33

Then there are other members that fall outside the tier system, with a weight of 0.00 or above 1.33. These are:

`+------------+--------+`
`| member_id  | weight |`
`+------------+--------+`
`| ucmerced   |   0.00 |`
`| ucsf       |   0.00 |`
`| hathitrust |   0.00 |`
`| utexas     |   3.00 |`
`| flbog      |   5.00 |`
`| usg        |   3.00 |`
`+------------+--------+`

mwarin: Those non-tier weights were, I believe, rather arbitrarily cobbled together, probably by Mike.

format_and_overlap_determination.md
aelkiss: :+1:

format_and_overlap_determination.md
```diff
@@ -0,0 +1,123 @@
...
+: Set by the Board and Executive Director. Portion of the budget to be paid for by HT Item allocations.
+
+**Cost Per Volume**
+: Target Cost / Total Number of HT Items
+
+**Member Weight**
+: A number assigned to a member by the ED that determines their share of Public Domain costs. 
+
+**Total Weight**
+: The sum of member weights found in the `ht_billing_members` table of the holdings database. **Which process/system has ownership of this table?**
```
mwarin, line 78: This probably starts as part of the Estimate Prospective Member Fee process. We need to know member weight when making an estimate. If not then, it ought to happen as part of the Add New Member process. Data entered through OTIS (or, less ideal, manually through mysql).

format_and_overlap_determination.md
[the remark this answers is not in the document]
aelkiss: What about:

For items in *clusters* that are not MPM (i.e. none of the items in the cluster have the item format multi-part monograph), organizations are considered to hold the item and are allocated a share for it if:
  * They are the billing entity for the item  (i.e. the depositor of an item is always assumed to hold it)
  * They submitted a holding that is in the cluster
 
 For items in clusters that are MPM, the process is more complicated. Organizations are considered to hold the item and are allocated a share if:

* They are the billing entity for the item (i.e. the depositor of an item is always assumed to hold it)
* They have a holding in the cluster with an n_enum matching the n_enum of an item.
* They have a holding in the cluster that has an empty n_enum. The same is not true of items with empty n_enums: holdings with empty n_enums match all items in an MPM cluster; but items with empty n_enums match only holdings with n_enums.
* They have holdings in the cluster, but none of the reported n_enums match any of the Item n_enums: that is, if no holdings n_enums match, then this is assumed to be a data problem, and the organization is considered to hold all the items in the cluster. [^2]

@nfulkers  Does that address any of the confusion? I think it is important this section be as clear as possible. I don't know whether my option is any better. It may be helpful for you to try stating the behavior in your words as you understand it, and then seeing if there's any spots we need to correct or clarify.

format_and_overlap_determination.md
nfulkers: So is this better characterized as a "holdings cluster" or something?  Or alternatively, just take out the word "clustering" from the first sentence in the Bib Record definition below.

format_and_overlap_determination.md
nfulkers: ooookay, this is where the system thing from my comment above gets worked out.

format_and_overlap_determination.md
nfulkers: do we need to spell this out more in the documentation itself?

format_and_overlap_determination.md
```diff
@@ -0,0 +1,123 @@
...
+: Sum of HScores for Member for each of SPM, MPM and SER
+
+**Total IC Cost for Member**
+: Total HScore for Member * Cost Per Volume
+
+**Total Cost for Member**
+: Total IC Cost for Member + Total PD Cost for Member + Extra Per Member
+
+
+## HT Item Overlap ##
```
nfulkers, line 102: In the legacy system, I think, we somehow allocate a share for each _copy_ of an item, if a member reports holding more than one copy?

Is that right and if so does the new system also do this?

format_and_overlap_determination.md
```diff
@@ -0,0 +1,123 @@
...
+Determining who holds a particular in-copyright HT Item.
+
```
nfulkers, line 104: In the legacy system, I think, we somehow allocate a share for each copy of an item, if a member reports holding more than one copy?

Is that right and if so does the new system also do this?

format_and_overlap_determination.md
```diff
@@ -0,0 +1,123 @@
...
+For HT Items in **clusters** that are not MPM, all organizations with a holding in the cluster and the billing entity for the HT Item are allocated a share. 
+
+For HT Items in **clusters** that are MPM, the process is more complicated. Organizations with holdings:
+- Organizations with holdings with empty **n_enum**. The reverse is **not** true; i.e. Items with empty n_enum don't match everything.
```
nfulkers, line 108: "Organizations with holdings with empty **n_enum**...what?  What do they match to?  Is there some kind of logic implied by the syntax or notation of this sentence that I'm missing?

format_and_overlap_determination.md
aelkiss: My impression was that the weights (including for the system members, which I think is based on the number of R1 research schools in the system) are encoded in the billing model which has been approved by the Board, so I would not classify them as "arbitrary". It would be worth checking with Mike that this matches his understanding, but yes, I think that given our current billing model this is the right way to do it.

format_and_overlap_determination.md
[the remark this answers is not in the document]
aelkiss: They have a weight in the old system, the same as in the new system. There is no special handling for system members in that the fee is always scaled by the weight; the thing that differs for system members (in both old and new systems) is the weight.

format_and_overlap_determination.md
```diff
@@ -0,0 +1,123 @@
...
+**Total Cost for Member**
+: Total IC Cost for Member + Total PD Cost for Member + Extra Per Member
+
+
+## HT Item Overlap ##
+Determining who holds a particular in-copyright HT Item.
+
+For HT Items in **clusters** that are not MPM, all organizations with a holding in the cluster and the billing entity for the HT Item are allocated a share. 
+
+For HT Items in **clusters** that are MPM, the process is more complicated. Organizations with holdings:
```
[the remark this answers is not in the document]
aelkiss: Probably should be written as "members with holdings" or perhaps "institutions with holdings"?

format_and_overlap_determination.md
aelkiss: The system weights come from this: https://www.hathitrust.org/sites/www.hathitrust.org/files/member-criteria-formalize.pdf which Natalie quoted from below

format_and_overlap_determination.md
aelkiss: @mwarin  I know I've been confused on this point in the past. I don't think the old system does this, though. As far as I can tell this test addresses this in the new system: https://github.com/hathitrust/holdings-backend/blob/master/spec/cost_report_spec.rb#L222-L233 -- multiple reported holdings of the same OCN lead to only one allocated share of each matching HTItem

format_and_overlap_determination.md
aelkiss: Does my proposed revision above https://github.com/hathitrust/holdings-backend/pull/77#discussion_r632687803 make this point any clearer?

format_and_overlap_determination.md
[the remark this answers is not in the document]
aelkiss: I edited the above to replace 'organization' with 'institution'.

format_and_overlap_determination.md
aelkiss: I think that's out of scope for this document, but certainly worth recording as part of the documentation of how to do estimates.

format_and_overlap_determination.md
aelkiss: Will change to 'preferred record from Zephir' and omit the reference to Zephir clustering.

format_and_overlap_determination.md
aelkiss: I think we should say:

- we get the file from AIM (do not specify individuals)
- how the file is generated is not in scope for this document

format_and_overlap_determination.md
aelkiss: The tier to weight mapping for Tier 1, 2, 3 is in a policy here:
https://www.hathitrust.org/sites/www.hathitrust.org/files/fee-model-change-proposal.pdf

@nfulkers  will add info about systems & specific R1 schools in Confluence

format_and_overlap_determination.md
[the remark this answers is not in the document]
aelkiss: See above discussion about systems

format_and_overlap_determination.md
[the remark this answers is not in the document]
aelkiss: Might need some more introductory text to make it clear that we are not actually matching holdings -- we are figuring out who holds an item, which includes (but is not limited to) matching holdings to HT items

format_and_overlap_determination.md
aelkiss: We should add these two links above ^^ to this documentation

aelkiss:
> re: the section on member weight, [this google sheet](https://docs.google.com/spreadsheets/d/1C74IUynslWOSCAkdlcLO8jgRDvuXq-JcQUsyapD9JEI/edit#gid=1388953783) is public-facing and describes how the PD weight is determined for system members (see the "2. System Members tab, Column H). So it is technically the ED who assigns the weight, but that determination is made based on a [formula](https://www.hathitrust.org/Cost) approved by the HT membership in 2019. Can we revise the section to reflect that (and include the link to the worksheet)?
> 
> I have edited the corresponding Confluence page to include these details as well.

We will add these links to the documentation

format_and_overlap_determination.md
aelkiss: @nfulkers believes my revisions above are sufficient

format_and_overlap_determination.md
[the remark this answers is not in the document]
aelkiss: @nfulkers believes these revisions address the confusion

format_and_overlap_determination.md
mwarin: We were told that the latest file (end of May 2021) would be the last file Margaret would be able to produce, as Umich is switching from Aleph to Alma (and presumably this report was based on Aleph).

format_and_overlap_determination.md
mwarin: For fees it doesn't matter how many copies you hold (>0). 

This makes sense considering that 
the fee for each member associated with a volume 
is multiplied by 1/H, where H is the number of members associated by that volume, 
not the total number of copies (of the title) reported held.

Number of copies (of a title) reported held only matter for access (e.g. ETAS).

PR #78, opened 2021-04-19
HT-2835: Daily Hathifile loading

aelkiss:
This orchestrates loading hathifiles on a daily basis. In particular, it adds:

-  a database table, currently called `holdings_files`, to track what files we have loaded on a daily basis or not
-  a `FileMutex` class to ensure processes aren't trying to run reports while we're loading data

It also adds the notion of top-level configuration separately from `Services`, which should help some to reduce the number of environment variables and generally make configuration more visible.

TODO before this is fully ready for review:

- rename `holdings_files` to `holdings_loaded_files`; class should be named `LoadedFiles`
- re-add some of the logic from database configuration, matching what's done here to ensure we can use either a url or hash options from config: https://github.com/mlibrary/keycard/blob/master/lib/keycard/db.rb#L71


aelkiss:
Given the changes for config here we will need to adjust our Kubernetes cronjob / pod configuration to include appropriate configuration. I can work on that, including adding config for a cron job for this, once we think this is headed in the right direction.

lib/holdings_file.rb
```diff
@@ -0,0 +1,19 @@
...
+require "sequel"
+require "services"
+
+# Information about a file of records loaded into the system
+class HoldingsFile < Sequel::Model(Services.holdings_db.rawdb[:holdings_files])
+  # primary key isn't autogenerated; we need to set it at creation
+  unrestrict_primary_key
+
+  def self.latest(**kwargs)
+    where(**kwargs).order_by(Sequel.desc(:produced)).limit(1).first
```
aelkiss: The splat here and below on line 16 are unnecessary, I think

sql/001_users.sql
```diff
@@ -1,4 +1,4 @@
 GRANT USAGE ON *.* TO 'ht_repository'@'%' IDENTIFIED BY 'ht_repository';
 GRANT SELECT ON `ht_rights`.* TO 'ht_repository'@'%';
-GRANT SELECT, INSERT, UPDATE, DELETE, LOCK TABLES ON `ht_repository`.* TO 'ht_repository'@'%';
-GRANT SELECT, INSERT, UPDATE, DELETE, LOCK TABLES ON `ht`.* TO 'ht_repository'@'%';
+GRANT ALL ON `ht_repository`.* TO 'ht_repository'@'%';
```
aelkiss, line 3: This probably isn't necessary -- I had changed this because at one point it was doing a `TRUNCATE` in the tests, which needed drop table privilege, but now it's doing it with transactions.

lib/hathifile.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+require "services"
+require "settings"
+require "loaded_file"
+
+# Responsible for locating and loading a single Hathifile, a file containing
+# HtItems represented as tab-separated values.
+#
+# Usage: Hathifile.new(date).load
```
mwarin, line 10: Add expected date format to this documentation.

lib/holdings_db.rb
```diff
@@ -71,61 +74,44 @@ def load_data_infile(table, file)
...
-  # # Accept all from env
-  # DB = HoldingsDB.connect
-  # # Full connection string
-  # DB = HoldingsDB.connect('mysql2://username:password@host:3435/myDatabase')
-  # DB = HoldingsDB.connect(user: 'otheruser', password: 'herPassword') # change user
-  def self.connection(connection_string = ENV["DB_CONNECTION_STRING"],
-    **kwargs)
+  # Connection connects to the database using the connection information
+  # specified by Settings.database, which should contain keyword parameters
+  # matching those taken by Sequel.connect.
```
aelkiss, line 79: This is not accurate

mwarin (approved):
I like the Ettin/Settings stuff. 
Haven't seen a lot of the rspec lingo you are using.
FileMutex currently allows any path as lockfile. I think we should specify a list of allowed paths, perhaps in a specified dir, just to make sure we don't proliferate lockfiles all over the file system.

No reason not to approve.

lib/settings.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require "ettin"
```
jsteverman, line 3: Not a real objection, but it does not appear there is a current maintainer of Ettin.

aelkiss: True. The canonical homes are http://github.com/mlibrary/ettin and http://github.com/mlibrary/canister and we can certainly update it as needed. @botimer will also reach out to @malakai97 and make sure that we (as MLibrary) can publish new versions on rubygems as needed.

aelkiss:
> Haven't seen a lot of the rspec lingo you are using.

Is there a specific rspec thing you're seeing that isn't clear? I'm certainly open to changing it especially in cases where what a test is testing or where its data is coming from isn't obvious

jsteverman (approved):
Do we expect to use FileMutex and LoadedFile for Holdings/OCNResolutions as well?

This all looks reasonable to me. HoldingsDB is an ugliness I would like to deal with at some point, but it's not new and is not a pressing matter.

aelkiss:
Yes, I'm expecting we'll want to use these for the other kinds of data as well, although it's not 100% clear to me when we need to use the `FileMutex`.

PR #80, opened 2021-04-29
Optimizations for when to recluster

aelkiss:
- Don't delete and create a new cluster when updating an OCNless HTItem
- Don't recluster if the old OCNs are all covered by concordance rules (one case when they will all still be in the new cluster)

This doesn't address:

- Optimizing when to recluster when removing an HTItem from a cluster entirely (although again, we shouldn't need to recluster if the old OCNs are covered by concordance rules)

- Any optimization for reclustering around concordance rules - hopefully, concordance deletes are rare enough this won't matter
- More generally, only reclustering if the last thing gluing two OCNs together was removed from the cluster - this is potentially brittle and difficult to determine
- Any optimizations of reclustering itself (i.e. batching updates to a particular cluster)

### DIFF
--- a/config/environments/production.yml
+++ b/config/environments/production.yml
@@ -1,0 +1,7 @@
+database:
+  opts:
+    adapter: 'mysql2'
+    host: <%= ENV['MYSQL_HOST'] %>
+    database: <%= ENV['MYSQL_DATABASE'] %>
+    user: <%= ENV['MYSQL_USER'] %>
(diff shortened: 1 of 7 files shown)


jsteverman (approved):
Maybe the config and docker-compose changes don't belong in this merge, but doesn't really matter.

aelkiss:
Agreed -- although I needed them there in order to actually run this, so it needs to be merged anyway.

PR #82, opened 2021-05-04
Further recluster optimizations

aelkiss:
- Batching during recluster
- Optimize when to recluster when HTItems move clusters

### DIFF
--- a/lib/cluster_ht_item.rb
+++ b/lib/cluster_ht_item.rb
@@ -46,17 +46,12 @@ class ClusterHtItem
     Retryable.with_transaction do
       if (cluster = cluster_with_htitem(ht_item))
         Services.logger.debug "removing old htitem #{ht_item.item_id}"
-        cluster.ht_item(ht_item.item_id).delete
+        old_item = cluster.ht_item(ht_item.item_id)
+        old_item.delete
 
-        # Note that technically we only need to do this if there were multiple
-        # OCNs for that HT item and nothing else binds the cluster together.
-        # It may be worth optimizing not to do this if the htitem has only
-        # one OCN, since that will be the common case. Not sure if it's worth
-        # optimizing away if there are multiple OCNs (i.e. doing the check to
-        # see if anything else binds the cluster together). In general the
-        # operation is probably rare enough that we don't need to worry about
-        # this for the time being.
-        Reclusterer.new(cluster).recluster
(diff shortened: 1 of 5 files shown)


aelkiss:
The May full hathifiles load succeeded with the optimizations here

lib/cluster_ht_item.rb
```diff
@@ -46,7 +46,8 @@ def delete
     Retryable.with_transaction do
       if (cluster = cluster_with_htitem(ht_item))
         Services.logger.debug "removing old htitem #{ht_item.item_id}"
-        cluster.ht_item(ht_item.item_id).delete
+        old_item = cluster.ht_item(ht_item.item_id)
+        old_item.delete
 
         # Note that technically we only need to do this if there were multiple
```
aelkiss, line 52: This comment is out of date now

lib/cluster_ht_item.rb
```diff
@@ -98,17 +101,39 @@ def update_or_add_ht_items(cluster)
...
+    # - There was 0 or 1 old OCN (so it couldn't have been 'glue')
+    # - old_ocns are all in concordance rules (so this item is not the 'glue')
+    # - old_ocns are a subset of the new ocns (so it remains 'glue')
+
+    old_ocns = old_ocns.to_set
     new_ocns = new_ocns.to_set
     concordance_ocns = cluster.ocn_resolutions.collect(&:ocns).flatten.to_set
 
-    !(current_ocns.subset?(new_ocns) || current_ocns.subset?(concordance_ocns))
+    !(old_ocns.count <= 1 ||
```
aelkiss, line 119: Worth pondering more whether we should extract the common logic here and with `move_needs_recluster?` - is there a named concept here we can extract or is the similarity incidental?

jsteverman: I *think* this is redundant. `move_needs_cluster?` is simply `update_needs_recluster?` with `new_ocns` as an empty set. In that case `old_ocns.subset?(new_ocns)` will always be false.

Changing `update_needs_recluster?(cluster, old_ocns, new_ocns=[])` and aliasing `move_needs_recluster?` got no complaints from rspec.

jsteverman (changes requested):
If my comment was correct about `update_needs_recluster?` and `move_needs_recluster?` being functionally identical then they should be boiled down to `needs_recluster?`

If I'm wrong, we need to do something to clarify the difference or they're going to confuse me for an afternoon in a year or two.

PR #84, opened 2021-05-07
HT-2837 (despite the branch name)

mwarin:
This is a refactor of lib/autoscrub.rb, and an addition of bin/process_new_holdings.sh that takes unseen member holdings files, scrubs them and loads the scrubbed files into mongo.

Pardon the mess.


aelkiss:
FWIW if we're running into unhelpfulness with rubocop we could try switching to standardrb (https://github.com/testdouble/standard) - we did this for otis and it seems fine so far -- less fiddliness and less to argue about than with the default rubocop settings.

lib/member_holding_header_factory.rb
```diff
@@ -0,0 +1,41 @@
...
+  SER = "serial"
+
+  def initialize(item_type, header_line)
+    @item_type   = item_type
+    @header_line = header_line
+  end
+
+  # Return a proper subclass of MemberHoldingHeader
+  # rubocop:disable Naming/AccessorMethodName
+  def get_instance
```
aelkiss, line 28: We have a couple different styles of factories going on here; I'd generally lean towards preferring the `Whatever.for(something)` style, and I think it's preferable to be consistent if possible

lib/scrub_fields.rb
```diff
@@ -94,24 +79,32 @@ def ocn(str)
...
-      @logger.error "Too many items (#{candidates.size}) in ocn #{str}"
+      Services.scrub_logger.warn "Too many ocns (#{candidates.size}) in ocn #{str}"
+    end
+
+    if candidates.size > 1
+      count_x(:multi_ocn_record)
     end
 
-    # DO WHILE MAYBE COMEFROM reject_value
+    # PLEASE DO WHILE MAYBE COMEFROM reject_value
```
aelkiss, line 94: ???

lib/scrub_fields.rb
```diff
@@ -94,24 +79,32 @@ def ocn(str)
...
     candidates.each do |candidate|
       catch(:rejected_value) do
```
aelkiss, line 96: Maybe worth extracting the contents of the block to a separate method?

lib/scrub_fields.rb
```diff
@@ -94,24 +79,32 @@ def ocn(str)
...
         candidate.strip!
 
-        # Try to find a reason to reject the candidate ocn
-        # Any time reject_value triggers, go ^^ to the catch.
+        # Try to find a reason to reject the candidate ocn.
+        # Any time reject_value triggers, it throws :rejected_value
+        # and we go back up to the catch statement inside the each-loop.
         candidate.nil? &&
-          reject_value("ocn is nil", "")
+          reject_value("ocn is nil")
```
aelkiss, line 104: I haven't seen `throw`/`catch` in Ruby before and had to look it up. Curious if others have seen it before. I don't think we're using the return value stuff that `catch` does, so could we just do something like `log_why_it_is_bad && next`?

aelkiss (commented):
I got midway through `scrub_fields` but will continue more later. Most of this is stylistic concerns rather than functional concerns so far -- let's talk as a group how much of this we want to address now, how much to save for a refactoring / cleanup pass after we have something working in production, and how much not to worry about.

bin/add_print_holdings.rb
```diff
@@ -10,9 +10,9 @@
 Services.mongo!
 
 filename = ARGV[0]
-Services.logger.info "Adding Print Holdings."
+Services.logger.info "Adding Print Holdings from #{filename}."
 
-holding_loader = HoldingLoader.new
+holding_loader = HoldingLoader.get_for(filename)
```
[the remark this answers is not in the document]
mwarin: I guess I wasn't sure about the naming convention. 
Fixed in the yet-to-be-pushed commit 089bc4f55cb7602d7c519e6b8eb3bae921adae6d

lib/holding_loader.rb
```diff
@@ -5,16 +5,24 @@
 
 # Constructs batches of Holdings from incoming file data
 class HoldingLoader
+
+  def self.get_for(filename)
```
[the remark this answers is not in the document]
mwarin: OK, fixed.

lib/holding_loader.rb
```diff
@@ -27,3 +35,24 @@ def finalize
     end
   end
 end
+
+## Subclass that only overrides item_from_line
+class HoldingLoaderTsv < HoldingLoader
```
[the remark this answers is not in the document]
mwarin: OK, fixed.

lib/member_holding.rb
```diff
@@ -0,0 +1,105 @@
...
+  end
+
+  def log(str)
+    Services.scrub_logger.info(str)
+  end
+
+  # Takes a line from a member holding file
+  # and populates a MemberHolding-object
+  def parse_str(str)
+    if str.nil? || str.class != String || str.empty?
```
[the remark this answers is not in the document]
mwarin: Fixed in to-be-pushed 03c2de8573e4fb7c189103225020fbfb0b3f7e46.

lib/member_holding.rb
```diff
@@ -0,0 +1,105 @@
...
+    end
+
+    @violations.empty?
+  end
+
+  # Given a column name and a value, sets the proper attribute
+  # via ScrubFields
+  # rubocop:disable Metrics/MethodLength
+  def set(col_type, col_val)
+    if col_val.nil? || col_val.class != String
```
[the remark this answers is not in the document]
mwarin: I don't know... probably a remnant from older times when I hadn't decided the shape of everything, and I saw no problem with keeping the redundancy. Fixed.

lib/member_holding.rb
```diff
@@ -0,0 +1,105 @@
...
+      raise ColValError,
+            "col_val for col_type #{col_type} empty/nil/wrong class"
+    end
+
+    case col_type
```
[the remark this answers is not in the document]
mwarin: Noted.

lib/member_holding.rb
```diff
@@ -0,0 +1,105 @@
...
+
+require "securerandom"
+require "json"
+require "services"
+require "scrub_fields"
+require "custom_errors"
+
+# Represents the information in one line from a member holding file
+class MemberHolding
+  attr_accessor :oclc, :local_id, :organization, :status, :condition,
```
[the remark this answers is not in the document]
mwarin: :organization and :mono_multi_serial needs to be writeable by other classes, moved the rest to attr_reader.

lib/member_holding.rb
```diff
@@ -0,0 +1,105 @@
...
+    else
+      raise ColValError,
+            "cannot handle column type #{col_type} (value: #{col_val})"
+    end
+  end
+  # rubocop:enable Metrics/MethodLength
+
+  def to_json(*_args)
+    {
+      ocn:               @ocn,
```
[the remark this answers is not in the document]
mwarin: Fixed.

lib/member_holding_file.rb
```diff
@@ -0,0 +1,230 @@
...
+
+    log("Scrub stats:")
+    scrub_stats.keys.sort.each do |ssk|
+      log("#{ssk}\t#{scrub_stats[ssk]}")
+    end
+
+    scrub_stats = {}
+  end
+
+  def get_member_id_from_filename(fn_str = @filename)
```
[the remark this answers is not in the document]
mwarin: Dropped the get_ prefixes.
I did the variable business thus for testing purposes.
`it "raises error if no member_id in filename" do
  expect { ok_mhf.member_id_from_filename("") }.to raise_error(FileNameError)
  expect { ok_mhf.member_id_from_filename(nil) }.to raise_error(FileNameError)
end                                                                           
`

lib/member_holding_file.rb
```diff
@@ -0,0 +1,230 @@
...
+
+    if SPEC_RX[:MEMBER_ID].match(member_id)
+      log("OK member_id #{member_id} in filename #{@filename}")
+      member_id
+    else
+      raise FileNameError, "Did not find a member_id in filename (#{fn_str})"
+    end
+  end
+
+  def get_item_type_from_filename(fn_str = @filename)
```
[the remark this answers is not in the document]
mwarin: Noted & dropped get_

lib/member_holding_file.rb
```diff
@@ -0,0 +1,230 @@
...
+      "item_type\t\"#{f_item_type}\"\t#{analyze_item_type(f_item_type)}",
+      "update_type\t\"#{f_update_type}\"\t#{analyze_update_type(f_update_type)}",
+      "date_str\t\"#{f_date_str}\"\t#{analyze_date_str(f_date_str)}",
+      "rest\t\"#{f_rest.join(" ")}\"\t#{analyze_rest(f_rest)}"
+    ].join("\n")
+    log(msg)
+    false
+  end
+
+  def analyze_member_id(str)
```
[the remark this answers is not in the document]
mwarin: Changed str to potential_member_id etc. It's only a member_id if the subsequent code thinks it is.

lib/member_holding_file.rb
```diff
@@ -0,0 +1,230 @@
...
+  end
+
+  def analyze_date_str(str)
+    not_nil_and_match(str, SPEC_RX[:DATE])
+  end
+
+  # Shortcut for the analyze_x methods above.
+  # Checks that str is not nil, and matches regexp
+  # otherwise issues a warning.
+  def not_nil_and_match(str, regexp)
```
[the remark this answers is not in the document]
mwarin: Agreed.

lib/member_holding_file.rb
```diff
@@ -0,0 +1,230 @@
...
+    elsif regexp.match?(str)
+      "ok"
+    else
+      "not ok: #{warning}"
+    end
+  end
+
+  # We allow the 'rest' to contain arbitrary labels, it just has to end
+  # with our required file extension(s).
+  def analyze_rest(arr)
```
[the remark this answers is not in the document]
mwarin: Added some more to the comment.

lib/member_holding_file.rb
```diff
@@ -0,0 +1,230 @@
...
+        yield item_from_line(line, col_map)
+      rescue BadRecordError => e
+        log("Rejected record #{filename}:#{line_no}, #{e.message}")
+      end
+    end
+  end
+
+  def item_from_line(line, col_map)
+    if line.nil? || line.class != String || line.empty?
+      raise "bad line"
```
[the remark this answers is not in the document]
mwarin: Just bad habits from strictly typed languages and/or remnants from earlier versions of the code where I wasn't sure what I was passing around. Removed.

lib/member_holding_header_factory.rb
mwarin: Agree. Fixed.

lib/scrub_fields.rb
mwarin: Just a bit of harmless INTERCAL-72

lib/scrub_fields.rb
mwarin: Done.

lib/scrub_fields.rb
mwarin: I was under the impression that `next` only works inside the current block. So especially now that I've broken out a big chunk of the ocn method, a lot of the potential rejections happen in another block.

In perl I would have labeled the loop I want to return to as "foo" and said `next foo`, and this is the closest I've come to it in Ruby. I like that I don't have to sprinkle `next` everywhere in the code too.

lib/holding_loader.rb
```diff
@@ -5,16 +5,24 @@
...
     @organization = nil
     @current_date = nil
   end
 
-  def item_from_line(line)
-    Holding.new_from_holding_file_line(line).tap do |h|
-      @organization ||= h.organization
-      @current_date ||= h.date_received
-    end
+  def item_from_line(_line)
```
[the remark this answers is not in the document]
mwarin: I did this in order to screw with the class as little as possible. 
I'm not sure I understand your point, so I'll leave it as is unless you really think it's worth doing.

lib/member_holding.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  def initialize(col_map = {})
+    @col_map       = col_map
+    @scrubfields   = ScrubFields.new
+    @violations    = []
+    @date_received = Time.new.strftime("%Y-%m-%d")
+    @uuid          = SecureRandom.uuid
+  end
+
+  def log(str)
```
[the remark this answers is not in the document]
mwarin: Done.

lib/scrub_fields.rb
aelkiss: Essentially I'm suggesting (inlining the relevant part of `reject_value`)

```ruby
if candidate.nil? 
  count_x("value rejected: ocn is nil")
  next
end
```

We could make that a little terser by doing e.g.

```ruby
next if candidate.nil? && count_x("value rejected: ocn is nil")
```

or you could extract methods and do something like 
```ruby
next if candidate_nil?
```
and extract a method along the lines of:
```ruby
def candidate_nil?
  if candidate.nil?
    count_x("value rejected: ocn is nil")
  end
```
(this will return the value of `count_x(...)`, but you could always explicitly return true there if necessary

Would that work, or am I missing something? I'm not sure the conciseness is worth the (to me) obscure construct -- let's discuss tomorrow

lib/holding_loader.rb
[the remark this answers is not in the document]
aelkiss: Generally I try to avoid abstract superclasses in Ruby, but it's not a big issue right now. If we find later that we're having trouble testing `HoldingsLoader` it might be worth considering doing the refactoring.

lib/member_holding_file.rb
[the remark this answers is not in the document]
aelkiss: :+1:

lib/member_holding_file.rb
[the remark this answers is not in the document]
aelkiss: OK, I guess this is another case of a more abstract kind of function - I was hoping we could rename `arr` to something more intention revealing, but I guess it doesn't make sense here.

lib/scrub_fields.rb
aelkiss: :crying_cat_face:

lib/scrub_fields.rb
```diff
@@ -229,21 +224,26 @@ def govdoc(str)
...
   # DRY code for the status, condition and govdoc functions
   def simple_matcher(rx, str)
-    output = []
-    str.strip!
-    m = rx.match(str)
-    output << m[0] unless m.nil?
-
     # Get the name of the calling method
-    cmeth = caller_locations[0].label
+    cmeth  = caller_locations[0].label
```
aelkiss, line 228: Consider moving cmeth to a parameter with a default value, e.g.

```ruby
def simple_matcher(rx, str, label=caller_locations[0].label
```

Which is to say, it seems like the important part is that we'd like some way to refer to what it is we're matching, not that it happens to be the calling method -- but it's fine to have it default to the name of the calling method.

I also wonder a bit if it might make sense to extract these to (tiny) classes `StatusMatcher`, `ConditionMatcher`, ... - that would also give a better way to hold the constants than as constants in `ScrubFields`. Not worth doing now, but maybe worth doing rather than adding additional regexes to `ScrubFields`/

lib/scrub_output_structure.rb
```diff
@@ -0,0 +1,159 @@
...
+  # /usr/src/app/data/member_data/xyz/log/2020-12-30/
+  def date_subdir!(subdir, date_str = Time.new.strftime("%F"))
+    mkdir!(
+      validate_subdir(subdir),
+      validate_date_str(date_str)
+    )
+  end
+
+  # Remove the dir that belongs to @member_id and everything under it.
+  def remove_recursive!
```
aelkiss, line 71: I was trying to understand why you were masking `FileUtils.remove_dir`'s return value, but AFAICT this doesn't appear to be called anywhere except tests -- consider removing?

lib/scrub_output_structure.rb
```diff
@@ -0,0 +1,159 @@
...
+# - member_data/
+# - - x/
+# - - - log/
+# - - - output/
+#
+# ... where x is the member_id.
+# Subdirs (loaded, log, output, ready_to_load) are generated upon init.
+# They can have dated subdirs.
+#
+# - - - loaded/
```
aelkiss, line 18: The directory structure looks good to me, and I like having this class to manage it.

lib/scrub_output_structure.rb
```diff
@@ -0,0 +1,159 @@
...
+  attr_reader :member_id, :member_dir, :member_log, :member_output,
+              :member_ready_to_load, :member_loaded
+
+  def initialize(member_id)
+    unless member_id.is_a?(String)
+      raise "member_id must be a string"
+    end
+
+    @member_id     = member_id
+    @member_dir    = mkbase!
```
aelkiss, line 52: Consider inlining these... e.g.

```ruby
@member_dir = mkdir!
@member_log = mkdir!("log")
# etc
```

lib/scrub_output_structure.rb
```diff
@@ -0,0 +1,159 @@
...
+
+  # Remove the dir that belongs to @member_id and everything under it.
+  def remove_recursive!
+    FileUtils.remove_dir(member_dir.path)
+    true
+  end
+
+  def latest(subdir)
+    base_dir      = mkdir!(validate_subdir(subdir))
+    latest_subdir = base_dir.children.select {|str| str.match?(DATE_STR) }.max
```
aelkiss, line 78: We're not expecting these directories to get extremely large -- only adding new files/directories when a member has new submissions, right?

lib/scrub_output_structure.rb
```diff
@@ -0,0 +1,159 @@
...
+    if latest_subdir.nil?
+      nil
+    else
+      Dir.new(File.join(base_dir, latest_subdir))
+    end
+  end
+
+  def to_json(*_args)
+    dir_hash = {
+      "member_id"            => member_id,
```
aelkiss, line 89: If we find ourselves wanting to add more kinds of subdirectories we might want to refactor to specify the kinds of subdirectories only once rather than in multiple places in this class; no need to change that now.

lib/scrub_output_structure.rb
```diff
@@ -0,0 +1,159 @@
...
+  end
+
+  # A call like:
+  # ScrubOutputStructure.new("xyz").date_subdir!("log")
+  # ... on 2020-12-30 makes the dir
+  # /usr/src/app/data/member_data/xyz/log/2020-12-30/
+  def date_subdir!(subdir, date_str = Time.new.strftime("%F"))
+    mkdir!(
+      validate_subdir(subdir),
+      validate_date_str(date_str)
```
aelkiss, line 66: It doesn't look like you're currently calling `date_subdir!` with a non-default `date_str` except from the tests. You could get away from needing to validate the date string here (always a losing proposition, as you note) by expecting a `Date` here rather than a `String`. e.g...

```ruby
def date_subdir!(subdir, date = Date.today)
  mkdir!(validate_subdir(subdir), date.to_s)
end
```

lib/scrub_output_structure.rb
```diff
@@ -0,0 +1,159 @@
...
+  end
+
+  # private private private private private private private
+  private
+
+  def validate_subdir(str)
+    if VALID_SUBDIR.include?(str)
+      str
+    else
+      raise "invalid subdir name: #{str}"
```
aelkiss, line 111: `raise ArgumentError, "invalid subdir name: #{str}"`

lib/scrub_output_structure.rb
```diff
@@ -0,0 +1,159 @@
...
+    end
+  end
+
+  # Allows 2020-50-50 and isn't Y100K-proof but hey
```
aelkiss, line 115: Consider removing (see comment above)

spec/auto_scrub_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "autoscrub"
+
+RSpec.describe AutoScrub do
```
aelkiss, line 6: `AutoScrub` currently has no test coverage, so it might be good to add something. I would consider allowing the autoscrub logger to be injected and then adding some basic integration tests -- does it create a file with the expected output; does it log errors in the place you expect; etc.

aelkiss:
I noticed `ScrubFields#issn` has no test coverage. Do any members currently submit ISSNs? If not, we should remove it. If so, we should test it.

spec/member_holding_file_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+
+  it "raises error if no item_type in filename" do
+    expect { bad_mhf.item_type_from_filename }.to raise_error(FileNameError)
+  end
+
+  it "checks a whole filename" do
+    expect(ok_mhf.valid_filename?).to be(true)
+  end
+
+  it "Raises an error if it can't figure out member_id or item_type" do
```
aelkiss, line 60: Consider inlining the `let`s you only use once, or moving them closer to the examples in which they're used, e.g.

```ruby
context "with bad filenames" do
    # each badness = !+
  let(:bad_fn_1) { "!!!!!_mono_full_20201230.tsv" } # bad member_id
  let(:bad_fn_2) { "umich_!!!!_full_20201230.tsv" } # bad item_type
  let(:bad_fn_3) { "umich_mono_!!!!_20201230.tsv" } # bad update type
  let(:bad_fn_4) { "umich_mono_full_!!!!!!!!.tsv" } # bad date
  let(:bad_fn_5) { "umich_mono_full_20201230.!!!" } # bad extension

  it "rejects ..."
end
```

aelkiss:
Tests generally look good and are easy to follow. As I suggested above you might consider inlining some of the `let`s especially when they're only used once or are very simple -- this is somewhat a matter of taste but I generally find that having the data closer to the examples makes them easier to follow.

lib/scrub_output_structure.rb
mwarin: Right, this makes more sense as spec teardown rather than something in the class itself.
Moved.

lib/scrub_output_structure.rb
mwarin: Yeah I can imagine we'll at least be adding something for shared print here.

lib/scrub_output_structure.rb
mwarin: Noted.

lib/scrub_output_structure.rb
mwarin: Ah yes, fixed the 3 occurrences of this.

mwarin (commented):
> I noticed `ScrubFields#issn` has no test coverage. Do any members currently submit ISSNs? If not, we should remove it. If so, we should test it.

I'd say most members do. I'll add something.

lib/scrub_output_structure.rb
mwarin: 👍

spec/member_holding_file_spec.rb
mwarin: By "inlining let's" do you mean just straight up declaring a variable inside the test?

lib/scrub_output_structure.rb
mwarin: I don't understand what you mean, so I'm not going to do it now unless this gets elevated.

spec/auto_scrub_spec.rb
mwarin: Added tests that check for output files and lack of exceptions.

spec/scrub_output_structure_spec.rb
```diff
@@ -42,10 +42,9 @@
     expect { test.date_subdir!("log", "JAN 1ST LAST YEAR") }.to raise_error(StandardError)
   end
 
-  # last
-  it "removes itself" do
-    expect(Dir.exist?(test.member_dir)).to be(true)
-    expect(test.remove_recursive!).to be(true)
-    expect(Dir.exist?(test.member_dir)).to be(false)
+  it "otherwise an after-all but rubocop does not like" do
```
aelkiss, line 45: What's rubocop complaining about? Disabling the cop would be preferable to working around it like this, IMO

aelkiss:
New specs look good. My leaning would be to squash this down at least some before merging, but let's see what the rest of the team prefers. If we do want to do that I'm happy to assist, and I'm free Friday right after our holdings meeting.

PR #86, opened 2021-05-14
Add cluster last modified

jsteverman:
This adds the field `last_modified` to Clusters and sets as Time.now.utc with a before_save callback. This may be the wrong name for it, as a save doesn't guarantee modification. 

Without a way to filter clusters of interest, consuming services will need to process the entire database. That's acceptable when we are performing monthly updates, but not so when we are performing updates at much higher frequency. 

edit: I should note that this doesn't deal with deletes, although in practice this shouldn't be too much of an issue.


aelkiss (approved):
:+1: I'm happy with this. I suppose instead of testing output to stdout, `full_report` could accept an IO-type object to output to, and then you could pass it a `StringIO` and test against the contents of that... but, I think this is fine.

PR #87, opened 2021-05-20
Ht 2940 load overlap reports for etas

jsteverman:
Overlap records for ETAS purposes need to make it to the MySQL database incrementally. This means upserts, which also means we need to know the unique identifiers for the records in the `holdings_htitem_htmember` table. I *think* those are `volume_id` and `member_id`, but they might also include `n_enum`? Anyway, this is described, if not successfully accomplished in `bin/update_overlap_report_database.rb`

These aren't print holdings operations and shouldn't really be a concern here. If ETAS needs an overlap computation saved in a SQL database, that seems like a problem for ETAS. However, it seems we're stuck with prior practice.

### DIFF
--- a/bin/export_etas_overlap_report.rb
+++ b/bin/export_etas_overlap_report.rb
@@ -1,9 +1,9 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
-require "services"
-
 $LOAD_PATH.unshift(File.join(File.dirname(__FILE__), "..", "lib"))
+require "services"
+require "settings"
 require "bundler/setup"
 require "utils/waypoint"
 require "utils/ppnum"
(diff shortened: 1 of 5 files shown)


aelkiss:
There are a couple of uses of the holdings database (or have been in the past) beyond ETAS, but I agree it's worth thinking about whether this is the right home longer term, or how we might envision this changing in the future. One thing I do want to try to avoid is things outside the scope of holdings directly connecting to mongodb - I would rather have some kind of defined interface rather than having things on the ETAS side need to know about and connect to mongo - but I think we could split the tasks of exporting what needs to be exported (a responsibility here) and loading it into MySQL (a responsibility somewhere else).

I would say that now is probably a good time we should discuss whether "overlap report" is the terminology we want to use here -- it's certainly led to confusion for me from the various kinds of overlap reports we produce.

aelkiss:
Is this superceded by #88 ?

jsteverman:
> Is this superceded by #88 ?

Yes, superceded by #88.

PR #90, opened 2021-06-09
HT-2909: updates to devops workflows

aelkiss:
- Update actions to use GitHub Container Registry rather than dockerhub (this seems to be the preference for now)

- Update tanka to use current k8s library

- Add wrapper script for creating a `Job` in kubernetes and an example for running an estimate. tanka isn't really suitable for doing this kind of thing, so for the time being this incorporate the k8s YAML directly. Any updates to the container spec via tanka would need to be manually applied to the wrapper script as well. Perhaps in the future we will have a better way to manage these.

### DIFF
--- a/bin/run_estimate.sh
+++ b/bin/run_estimate.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+if [[ -z $1 ]]; then
+  echo "Usage: $0 /path/to/ocns" 
+  exit 1;
+fi
+
+echo /usr/src/app/bin/run_generic_job.sh bundle exec ruby /usr/src/app/bin/compile_estimated_IC_costs.rb $1
--- a/bin/run_generic_job.sh
+++ b/bin/run_generic_job.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+SUFFIX=$1
+shift
+ARGS=$(perl -e 'print "[ \"", join("\", \"",@ARGV), "\"]"' "$@")
+kubectl -n holdings create -f - <<EOT
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: holdings-client-${SUFFIX}
+  namespace: holdings
+spec:
+  template:
+    spec:
+      containers:
+      - command: ${ARGS}
+        env:
+        - name: MONGODB_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              key: mongodb-password
+              name: holdings-mongodb
+        - name: MYSQL_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              key: mysql-password
+              name: holdings-mysql
+        - name: MONGOID_ENV
+          value: production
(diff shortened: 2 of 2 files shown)


aelkiss:
@botimer @daaang Just looking for a thumbs-up on the k8s-alpha and ghcr stuff. I know we discussed the question of running jobs -- in this case I generated the YAML via tanka and then pasted it into the shell script. I'm sure there's a better way to do it, but I haven't come across it.

bin/run_estimate.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+echo "Usage: $0 /path/to/ocns"
+run_generic_job.sh bundle exec ruby bin/compile_estimated_IC_costs.rb $1
```
botimer, line 4: This looks like it will be CWD-sensitive. Should it be? Also, should it have any error output if $1 is missing?

bin/run_generic_job.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+SUFFIX=$1
+shift
+ARGS=$(perl -e 'print "[ \"", join("\", \"",@ARGV), "\"]"' "$@")
+kubectl -n holdings create -f - <<EOT
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: holdings-client-$SUFFIX
```
botimer, line 10: I tend to use the ${} syntax in these embedded situations; maybe superstition.

kube/foo/generic-job.yaml
```diff
@@ -0,0 +1,65 @@
...
+        - mountPath: /usr/src/app/config/large_cluster_ocns.txt
+          name: large-cluster-ocns
+          subPath: large_cluster_ocns.txt
+        - mountPath: /htprep
+          name: htprep
+      restartPolicy: OnFailure
+      securityContext:
+        fsGroup: 1190
+        runAsGroup: 1191
+        runAsUser: 1000
```
botimer, line 46: Are these magic and confirmed good? Or arbitrary?

botimer (commented):
The jsonnet-lib stuff looks good with the possible exception of the kausal import. If you're able to show/diff, I guess it's correct as-is. The change to "all pluses" matches my understanding of the current recommendations -- that is, to flatten all of the mixin calls and let that handle merging, rather than doing any chaining.

bin/run_estimate.sh
aelkiss: :shrug: Probably not, and probably. Good catch.

bin/run_generic_job.sh
aelkiss: :+1:

kube/foo/generic-job.yaml
aelkiss: Magic and confirmed good (matches groups assigned on the NFS mounts)

PR #92, opened 2021-06-10
minor scrub/load fixes

mwarin:
Addressing:
* bin/process_new_holdings.sh did not recognize .gz files in data/new/
* issues with nil values in records that had fewer columns than the header
* lib/autoscrub.rb calculating the wrong batch size for .gz files

### DIFF
--- a/lib/member_holding.rb
+++ b/lib/member_holding.rb
@@ -52,7 +52,10 @@ class MemberHolding
   # rubocop:disable Metrics/MethodLength
   def set(col_type, col_val)
     if col_val.nil?
-      raise ColValError, "col_val for col_type #{col_type} is nil"
+      Services.scrub_logger.warn(
+        "col_val for col_type #{col_type} is nil"
+      )
+      col_val = ""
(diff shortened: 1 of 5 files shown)


aelkiss (approved):
:+1:

lib/member_holding.rb
```diff
@@ -52,7 +52,10 @@ def parse(str)
   # rubocop:disable Metrics/MethodLength
   def set(col_type, col_val)
     if col_val.nil?
-      raise ColValError, "col_val for col_type #{col_type} is nil"
+      Services.scrub_logger.warn(
```
aelkiss, line 55: Maybe worth a test that it warns but doesn't raise?

mwarin: Sure.

mwarin: Added https://github.com/hathitrust/holdings-backend/pull/92/commits/579c3cf38d7de7dd0518ffbf97e9e140f897ce23

PR #95, opened 2021-06-14
Autoscrub multi OCN

mwarin:
Based on:
https://tools.lib.umich.edu/confluence/display/HAT/Print+Holdings%3A+Options+for+Multiple+OCNs
https://github.com/hathitrust/holdings-backend/blob/master/adr/0004-order-independence.md

... and the fact that a submitted holdings record with multiple OCNs previous to this fix comes through as an array of numbers, which upon loading gets interpreted as 0 and eventually causing a Mongo::Error::MaxBSONSize

It used to be:
```
Autoscrub input:  ocn1,ocn2 \t ...
Autoscrub output: {"ocn":[1,2], ...} <<-- root of problem
Loader internal: {"_id"=>BSON::ObjectId(...), "ocn"=>0, ...}
```

But now it is:
```
Autoscrub input:  ocn1,ocn2 \t ...
Autoscrub output: {"ocn":1, ...}
Autoscrub output: {"ocn":2, ...}
Loader internal: {"_id"=>BSON::ObjectId(...), "ocn"=>1, ...}
Loader internal: {"_id"=>BSON::ObjectId(...), "ocn"=>2, ...}
```

Also added bin/inspect_ocn.rb as a utility/debug tool, unrelated to the multi OCN issue.

### DIFF
--- a/bin/inspect_ocn.rb
+++ b/bin/inspect_ocn.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+# For manual/ocular inspection of a single OCN.
+# Takes an OCN and outputs a pretty-printed matching cluster.
+
+$LOAD_PATH.unshift(File.join(File.dirname(__FILE__), "..", "lib"))
+require "cluster"
+require "services"
+require "json"
+
+Services.mongo!
+
+def look_up(ocn)
+  cluster = Cluster.find_by(ocns: ocn.to_i)
+  if cluster.nil?
+    "No cluster found for OCN #{ocn}."
+  else
+    JSON.pretty_generate(cluster.as_document)
+  end
+end
+
+if $PROGRAM_NAME == __FILE__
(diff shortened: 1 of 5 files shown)


bin/inspect_ocn.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# For manual/ocular inspection of a single OCN.
```
jsteverman, line 3: I have a couple like this sitting in my local repo. We need to do some re-org at **some point** so we can both add all these helper type scripts and clean up some clutter.

jsteverman (approved):
Addresses the problem and improves what we have. 

The MemberHolding and Holding classes should be consolidated, or they should be better differentiated, but that doesn't need to happen now.

mwarin:
Merge!

PR #97, opened 2021-06-25
HT-3026: Nightly etas overlap updates

jsteverman:
Caches cluster level counts. Could still be done for MPMs.

### DIFF
--- a/bin/full_etas_overlap.rb
+++ b/bin/full_etas_overlap.rb
@@ -1,0 +1,38 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "pathname"
+$LOAD_PATH.unshift(Pathname.new(__dir__).parent + "lib")
+
+require "services"
+require "settings"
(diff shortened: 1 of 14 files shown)


aelkiss:
Probably should be rebased, maybe should be squashed some.

aelkiss (approved):
Looks good to me. I like that this moves some of the Mongoid-specific stuff out of the overlap computation; given that we have to fetch the entire cluster anyway, this seems like this minimizes the number of times we need to look through the entire set of holdings for a cluster.

It may be worth spending a little bit of time looking at performance vs. size of the cluster to make sure that there isn't still an n^2 issue. @mwarin  and I set up a test case with synthetic data to do something similar with holdings updates, which was helpful to verify that our changes were actually making an improvement; it was also helpful for being able to run the profiler without getting extremely noisy data. I think both those things would be helpful to do here as well, but shouldn't hold up merging this.

PR #101, opened 2021-07-07
Pry shell

mwarin:
DRY script so you don't have to do:

```
bundle exec pry -I lib
require "this"
require "that"
...
```

every time.


aelkiss (approved):
:+1:. Essentially this is the same point as `rails console`. Might be worth doing additional `require`s, but that isn't necessarily the job of this script -- we need to do some reorganization anyway and perhaps have a top-level `holdings.rb` that requires eeverything.

mwarin:
Merging with 1 approval from @aelkiss

PR #103, opened 2021-07-13
script to get all holdings for 1+ member(s)

mwarin:
So I can get all holdings to file, so I can split them up and let loose a bunch of pods and delete them.


mwarin:
For clarification, this script generates data that can be sorted & chunked & fed into bin/delete_holdings_by_uuid.rb.

aelkiss (approved):
I haven't tried the mongo query, but assuming it does what you want this seems fine. If we rely on this elsewhere we should add a test (and we could certainly do that now) but I think it's fine to merge as-is for now.

PR #105, opened 2021-07-22
HT-3038: Individual etas reports

jsteverman:
The previous version only produced records for holdings that matched HT Items. This version has at least one record for every holding.


aelkiss (approved):
Looks good to me; definitely glad to see specs added here. If performance is still lacking, it would probably be worth doing some profiling to verify where the issues are. I'm not seeing any obvious problems where it's iterating over the cluster multiple times unnecessarily. I do wonder how well `missed_holdings` will perform,  but at least we're only calling it once per cluster.

bin/export_etas_overlap_report.rb
```diff
@@ -13,41 +13,83 @@
...
+
+  def missed_holdings(cluster, holdings_matched)
+    if organization.nil?
+      cluster.holdings - holdings_matched.to_a
+    else
+      cluster.holdings.group_by(&:organization)[organization] - holdings_matched.to_a
+    end
+  end
+
+  def write_overlaps(cluster, organization)
```
mwarin, line 64: I'd like to see the different outputs of write_overlap (what is passed to write_record and what is returned) explained in a comment.

mwarin (commented):
I don't think bin/export_etas_overlap_report.rb needs to require zinzout or waypoint anymore.

mwarin (approved):
OK with me.

mwarin (approved):

PR #106, opened 2021-07-27
HT-3072 add narrative and stat tracking to estimate generation

jsteverman:
In addition to the IC cost, this adds information about the number of unique OCNs and the count/percent of matches. In the old system we also reported the number of usable holdings (MPM, SPM, SER), but that is not readily available to `bin/compile_estimated_IC_costs.rb` because it is only given a list of OCNs.

### DIFF
--- a/bin/compile_estimated_IC_costs.rb
+++ b/bin/compile_estimated_IC_costs.rb
@@ -12,6 +12,8 @@ require "services"
 
 Services.mongo!
 
+# Given a file of unique OCNs, generate a cost estimate.
+
 if __FILE__ == $PROGRAM_NAME
   BATCH_SIZE = 10_000
   waypoint = Utils::Waypoint.new(BATCH_SIZE)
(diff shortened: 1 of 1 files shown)


bin/compile_estimated_IC_costs.rb
```diff
@@ -21,26 +23,41 @@
...
+  num_items_pd = 0
+  num_items_ic = 0
 
   File.open(ocn_file).each do |line|
     ocn = line.to_i
+    ocns << ocn
     waypoint.incr
 
     # Find a cluster
     cluster = Cluster.find_by(ocns: ocn.to_i,
```
mwarin, line 41: ocn has already been to_i'd on line 36.

mwarin (approved):
Looks solid. My only concern is that the temporary org_id is "prospective_member", which I assume means you can't run 2 estimates at the same time for different members, not that we're super likely to ever want to do that.

bin/compile_estimated_IC_costs.rb
mwarin: This comment pertains to "ocn.to_i" on line 41. I must have clicked wrong.

aelkiss (approved):
Looks good to me as well. `prospective_member` isn't persistent in mongodb, so we should be able to run multiple estimates simultaneously without issue.

PR #107, opened 2021-08-03
Costerport utils

mwarin:
Copied and modified the scripts I've been using to generate the across-report comparisons for cost reports.
The only change to existing code is letting ht_members.rb output all current members to stdout when called from commandline.

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -20,8 +20,8 @@ jobs:
 
     - name: Tag latest release in DockerHub
       run: |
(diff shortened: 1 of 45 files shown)


aelkiss (changes requested):
There's a bunch of WIP changes from me here -- can we make a new branch that only includes the relevant commits for the new bin scripts?

mwarin:
Yeah, just saw that. Bad branch management on my part.

mwarin:
abort abort abort

PR #108, opened 2021-08-03
added scripts for generating cross-costreport comparisons

mwarin:
This time without a buncha commits from other branches!

Copied and modified from https://github.com/hathitrust/print_holdings, a couple of scripts that look at old cost reports and generates 3 files:

1. A history of total-cost per member per report
2. A history of total-cost diff (month to month) per member per report
3. A history of total-cost diff percentage (month to month) per member per report

These all go into the finished product, the Google Sheet, as separate tabs.

### DIFF
--- a/bin/append_sheets.pl
+++ b/bin/append_sheets.pl
@@ -1,0 +1,193 @@
+use strict;
+use warnings;
+
+=pod
+
+Take any number of tsv files greater than or equal to 2.
+For each row in file x, add all cols to a corresponding row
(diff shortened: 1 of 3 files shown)


aelkiss (approved):
Looks OK to me for now. If we find that we need to add a feature or otherwise make a non-trivial change to `append_sheets.pl` or `cost_changes.sh`, I think it's worth translating those to Ruby and getting them under test.

PR #110, opened 2021-08-04
temp solution for prepping HT003-files

mwarin:
As long as we're working with HT003, this way we'll have something shared in the master branch, and I have something to document for https://tools.lib.umich.edu/jira/browse/HT-3063 .

### DIFF
--- a/bin/prep_loadfiles.sh
+++ b/bin/prep_loadfiles.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+# Temporary solution until we're using autoscrub.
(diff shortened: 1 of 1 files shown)


jsteverman:
What does this replace? Or does it codify some previously manual processes in our loading of holdings?

mwarin:
> What does this replace? Or does it codify some previously manual processes in our loading of holdings?

Yeah this is stuff that I had/ran on punch and or locally. I think @aelkiss  has some version of this in a WIP branch too.

jsteverman (approved):

PR #111, opened 2021-08-06
HT-2907: lib organization

jsteverman:
This introduces the following modules:
lib/clusterable
lib/clustering
lib/loader

The multiple commits for each are because a `git mv` and file modification need to happen in separate commits or the file history gets lost. (I think)


mwarin (approved):
Given the nature of these diffs, I just bounced an eyeball on them. I trust in rubocop and rspec. Approve.

aelkiss (approved):
Looks like a good start. It would probably be helpful (when the reorganization is done) to give some indication of what belongs in each top-level module.

PR #115, opened 2021-08-23
Test that update_overlap_table only updates modified clusters

aelkiss:
Start of an attempt to figure out why the current run of holdings updates seems to be updating all or most of the holdings table; this adds a test to ensure that the mongo condition for having htitems & being modified is working as expected.


aelkiss:
Superceded (essentially) by #132

PR #118, opened 2021-08-30
Fix add print holdings loader

mwarin:


aelkiss (approved):
:+1:

jsteverman:
`spec/add_print_holdings_spec.rb` should probably be ` spec/loader/add_print_holdings_spec.rb`

PR #120, opened 2021-08-31
Ht 3115 refactor reports

jsteverman:
Pulling logic out of bins into lib/report and adding some specs. 

lib/report/overlap_report.rb and bin/reports/export_overlap_report.rb appear to be unnecessary. They are the single member special case of the full_etas_overlap. I didn't do much with full_etas_overlap as it is already very simple.

### DIFF
--- a/bin/reports/compile_cost_reports.rb
+++ b/bin/reports/compile_cost_reports.rb
@@ -1,7 +1,7 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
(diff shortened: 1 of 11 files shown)


aelkiss (approved):
Changes here look fine, but I notice that generating overlap reports is not documented in `routine_tasks.md`.  I think the two use cases are the daily update for the production table and the per-member overlap reports that we upload to Box/Dropbox. It would be good to make sure that we have appropriate documentation and specs for both those cases, and then we can remove the unnecessary reports.

PR #121, opened 2021-09-01
Prevent ETAS update session time out

jsteverman:


aelkiss (approved):
:+1: Yuck. We can certainly give it a try.

aelkiss:
If this does fix the problem we'll probably want to move this elsewhere rather than needing to have the report set it up.

PR #123, opened 2021-09-02
HT-3082: Compute item format using only HTItem

aelkiss:
- use bib_fmt = 'SE' to detect serials
- remove SerialsFile
- Add traits to ht_item factory for constructing items of expected type
- default to 0 for various counts to fix mysql warning


aelkiss:
I will build an image from this branch and run a cost report using it; then we can compare what difference it makes.

jsteverman (approved):
I am very happy to see the serials file go. I don't know where the bib_fmt: "s", "m", "r" came from which gives me some pause, but it might have just been a mistake.

PR #129, opened 2021-09-08
proof of concept

mwarin:
Instead of requiring each file to be aware of the relative path to `lib/`, I propose a class that is repeated in each ruby dir that recursively:

- adds `lib/` to `$LOAD_PATH` if `lib/` is in the current dir
- or asks its clone in the dir above to try

In this example, `lib-finder-test/a/b/c/d/e/f/test.rb` can get `Services.logger` by virtue of `require_relative "lib_finder"`. 
`test.rb` can be moved to any other dir (as long as they all have a clone of `lib_finder.rb`).

### DIFF
--- a/lib-finder-test/a/b/c/d/e/f/lib_finder.rb
+++ b/lib-finder-test/a/b/c/d/e/f/lib_finder.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require "digest"
+
+this_finder   = __FILE__
+this_checksum = Digest::MD5.file(this_finder)
+dir           = File.dirname(this_finder)
+
+if File.exist?(File.join(dir, "lib"))
+  # We found lib in the current dir and we're done.
+  $LOAD_PATH.unshift(File.join(dir, "lib"))
+elsif File.exist?(File.join(dir, "..", "lib_finder.rb"))
+  # No lib/ in this dir but there's another lib_finder above, so recurse upwards.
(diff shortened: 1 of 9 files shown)


aelkiss:
It sounds like this is to address issues like:

```
$LOAD_PATH.unshift(File.join(File.dirname(__FILE__), "../..", "lib"))
```

There are some ways in which it's possible to rely on bundler/rubygems to figure out the path to `lib` and add it to the path, but those are always a bit mystifying to me, and (at least in the past) have required adding a `.gemspec` to your project even when it really isn't conceptually a gem.

The other option is not to rely on the bin scripts to do this at all, and just do `bundle exec ruby -I lib bin/whatever.rb`, or (probably preferable in docker-land) put the lib directory in the `RUBYLIB` environment variable.

That said, I think ultimately my preference for holdings is rather than having 598723948234 bin scripts, we cut that down dramatically and take CLI arguments to determine which report to run or what kind of data to load, etc,  in which case it might be less odious to use the `$LOAD_PATH.unshift` approach.

If we do want to continue to have bin scripts responsible for somehow finding & adding lib to the path, I think it would probably also be worth investigating existing approaches to this (e.g. one could imagine a gem that when `require`ed finds the nearest relevant `lib` path and adds it to `$LOAD_PATH` -- although I don't know of something off the top of my head).

aelkiss:
Superceded by #134  I think

mwarin:
PR #134 solves the issue I was exploring with this branch, so closing it without merging.

PR #130, opened 2021-09-13
Merge concordance

jsteverman:
Replace concordance bash with ruby
    
    * eliminates data/<test_data>
    * adds lib/concordance_validation/delta.rb in place of bash + comm
    * documents concordance handling
    * rubocop cleanup
    * Delta accepts gzip


lib/concordance_validation/delta.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    def initialize(old_conc_filename, new_conc_filename)
+      @old_conc = open_concordance(old_conc_filename)
+      @new_conc = open_concordance(new_conc_filename)
+      @adds = Hash.new {|h, resolved| h[resolved] = Set.new }
+      @deletes = Hash.new {|h, resolved| h[resolved] = Set.new }
+    end
+
+    def run
+      old_conc_lines = old_conc.readlines.to_set
```
aelkiss, line 20: `comm` is more efficient memory-wise in that it doesn't need to read the entire old file into memory; rather, it relies on both files being sorted and does a sort of merge. Given that this diff format doesn't seem to be totally working for us and that we might need to make some other changes here anyway to reduce churn, I think this is OK for the time being.

aelkiss (commented):
Organization looks fine. I think it's worth verifying that `ConcordanceValidation::Delta` will work on an entire (validated) concordance file without choking due to the large amount of data. Other than that, should be good to go once the tests pass.

lib/concordance_validation/delta.rb
jsteverman: While it's definitely less efficient than `comm`, `Delta.new.run` only took ~10 minutes so worth the trade off for future flexibility.

aelkiss: :+1:

PR #132, opened 2021-09-16
add/delete only changed overlap records

jsteverman:
* moves ETAS database ops to Module ETAS


lib/etas/overlap_table_update.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+    end
+
+    def overlap_table
+      Services.holdings_db[TABLENAME]
+    end
+
+    def run
+      cutoff_str = cutoff_date.strftime("%Y-%m-%d %H:%M:%S")
+      Services.logger.info "Upserting clusters last_modified after #{cutoff_str} to #{TABLENAME}"
```
aelkiss, line 36: `TABLENAME` not defined here

lib/etas/overlap_table_update.rb
```diff
@@ -0,0 +1,53 @@
...
+      Cluster.with_session do |session|
+        session_refresh = keep_alive(session)
+        Cluster.where("ht_items.0": { "$exists": 1 },
+                      last_modified: { "$gt": cutoff_date }).no_timeout.each do |cluster|
+                        cu = ETAS::ClusterUpdate.new(overlap_table, cluster)
```
aelkiss, line 41: Consider extracting this block to a separate method to better expose what it is we're doing here and appease rubocop

bin/etas/update_overlap_table.rb
```diff
@@ -0,0 +1,31 @@
...
+if __FILE__ == $PROGRAM_NAME
+
+  # Default: Now - 36 hours
+  cutoff_date = Date.today - 1.5
+
+  unless ARGV.empty?
+    cutoff_date = Date.parse(ARGV.shift)
+  end
+
+  ETAS::OverlapTableUpdate.new(cutoff_date).run
```
aelkiss, line 30: I'm not sure this is really the right name. This table existed before ETAS and is used for other (albeit less visible and perhaps in some cases theoretical) things like section 108 access and direct SSD access (for Michigan anyway) and for driving 'held by my library' facets in solr.

Prior to ETAS, we generally referred to this table as the "production holdings table". I think calling the "overlap table" is more specific while still being accurate. I'm not immediately sure where to put it -- it has some affinity with reports but isn't a report as such. That may be an argument for leaving it where it is (i.e. not in a submodule) for the time being esp. as ultimately we want to see if we can do away with it and have an API direct to the source instead.

aelkiss: Perhaps the right concept is something like "production holdings" or "relational database holdings", but nothing immediate comes to mind as far as a better choice for the module name here.

lib/etas/cluster_update.rb
```diff
@@ -0,0 +1,63 @@
...
+        @new_overlaps = ClusterOverlap.new(cluster, nil).each.to_a.map(&:to_hash)
+      end
+
+      @new_overlaps ||= []
+    end
+
+    # Overlaps for ht_items currently in the cluster
+    # We will not find overlaps for items that have moved to another cluster,
+    # those will be deleted when the other cluster adds their new overlaps.
+    def existing_overlaps
```
aelkiss, line 37: This seems a little bit clunky; I'll see if I can propose an alternative.

aelkiss (changes requested):
I think the split into `OverlapTableUpdate` and `ClusterUpdate` makes sense. I think we should try to come up with a better name for the module they're in (admittedly not my forte).

I'll separately comment with thoughts on a test to add to make sure we're doing what we think we are regarding minimizing churn.

lib/etas/overlap_table_update.rb
```diff
@@ -0,0 +1,53 @@
...
+
+module ETAS
+
+  # An update of the table used for ETAS: holdings_htitem_htmember
+  class OverlapTableUpdate
+    attr_accessor :cutoff_date, :waypoint, :num_deletes, :num_adds
+
+    def initialize(cutoff_date = nil)
+      @cutoff_date = cutoff_date || Date.today - 1.5
+      @waypoint = Utils::Waypoint.new(BATCH_SIZE)
```
aelkiss, line 16: This should be a parameter to initialize, rather than a constant.

lib/etas/overlap_table_update.rb
```diff
@@ -0,0 +1,53 @@
...
+      Thread.new do
+        loop do
+          sleep(seconds)
+          session.client.command(refreshSessions: [session.session_id])
+        end
+      end
+    end
+
+    def overlap_table
+      Services.holdings_db[TABLENAME]
```
aelkiss, line 31: I would rather see (in `services.rb`):

```ruby
Services.register(:holdings_db_overlap_table) { Services.holdings_db[:holdings_htitem_htmember] }
```

or something along those lines, and getting rid of the `stub_const` calls

spec/etas/cluster_update_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+require "etas/overlap_table_update"
+
+RSpec.describe ETAS::ClusterUpdate do
+  let(:update) { ETAS::OverlapTableUpdate.new }
+  let(:h) { build(:holding, organization: "upenn") }
+  let(:ht) { build(:ht_item, :spm, ocns: [h.ocn], billing_entity: "not_same_as_holding") }
+  let(:ht2) { build(:ht_item, :spm, billing_entity: "not_same_as_holding") }
+
+  before(:each) do |_spec|
+    stub_const("BATCH_SIZE", 10)
```
aelkiss, line 16: See comment above; I'd really like to get rid of these constants.

lib/etas/overlap_table_update.rb
```diff
@@ -0,0 +1,53 @@
...
+      cutoff_str = cutoff_date.strftime("%Y-%m-%d %H:%M:%S")
+      Services.logger.info "Upserting clusters last_modified after #{cutoff_str} to #{TABLENAME}"
+      Cluster.with_session do |session|
+        session_refresh = keep_alive(session)
+        Cluster.where("ht_items.0": { "$exists": 1 },
+                      last_modified: { "$gt": cutoff_date }).no_timeout.each do |cluster|
+                        cu = ETAS::ClusterUpdate.new(overlap_table, cluster)
+                        cu.upsert
+                        waypoint.incr(cu.deletes.count + cu.adds.count)
+                        waypoint.on_batch {|wp| loggin.info wp.batch_line }
```
aelkiss, line 44: `loggin` -- typo?

lib/etas/overlap_table_update.rb
aelkiss, line 41: I'd like to see something like

```ruby
logger.debug("Processing overlap for cluster #{cluster._id}")
```

lib/etas/cluster_update.rb
aelkiss: This seems like it works:

```ruby
    def existing_overlaps
      @existing_overlaps ||= cluster.ht_items.map do |ht_item|
        @overlap_table.filter(volume_id: ht_item.item_id).map(&:to_hash)
      end.flatten
    end
```

If you wanted, you could extract the block body as a separate function, e.g.

```ruby
    def existing_overlaps
      @existing_overlaps ||= cluster.ht_items.map { |h| existing_item_overlap(h) }.flatten
    end

    private
   
    def existing_item_overlap(ht_item)
      @overlap_table.filter(volume_id: ht_item.item_id).map(&:to_hash)
    end
```

aelkiss:
Regarding testing that minimizing churn is working as expected:

The specs are a little bit hard to follow but I think are basically already doing this by checking that the delete count is smaller than the old overlap count & likewise that the add count is smaller than the new overlap count.

You could consider adding a more explicit test like:

```ruby
describe "#deletes" do
  it "does not include things in both the old & new overlaps" do
    do_some_setup
    expect(deletes).not_to contain(thing_that_shouldnt_be_deleted)
  end
end

describe "#adds" do
  it "does not include things in both the old & new overlaps" do
    do_some_setup
    expect(adds).not_to contain(thing_that_shouldnt_need_to_be_re_added)
  end
end
```

Regarding testing that it's actually reducing churn in mysql, probably the way to do that would be to stub the overlap table and then check that it receives the expected number of insert & delete calls:

```ruby
it "doesn't churn" do
  mock_overlap_table = stub(:overlap_table)
  allow(mock_overlap_table).to receive(:filter).and_return(mock_overlap_table)
  expect(mock_overlap_table).to receive(:delete).exactly(M).times
  expect(mock_overlap_table).to receive(:insert).exactly(N).times  

  do_some_setup
  described_class.new(mock_overlap_table, cluster).upsert
end
```

You could I suppose set an expectation that `filter` or `insert` receives the parameters you expect it to; that's probably overkill since we're already checking the actual contents of the table. Doing it this way just allows us to make sure that we aren't calling `insert` or `delete` more than we expected.

lib/etas/cluster_update.rb
jsteverman: I had something similar that was leaving me with nil and nil related errors which is how I ended up with my verbose clunker. Confirmed, yours works.

lib/etas/overlap_table_update.rb
jsteverman: I registered it as :relational_overlap_table. I am starting to think references to 'holdings_db' are a problem that needs addressing and don't want to perpetuate it.

bin/etas/update_overlap_table.rb
jsteverman: I gave up on giving these a module name and moved back into lib.

lib/etas/overlap_table_update.rb
jsteverman: Done

spec/etas/cluster_update_spec.rb
jsteverman: Constants killed.

aelkiss (approved):
Changes look good to me. Should be good to go once you resolve the merge conflict & squash.

spec/cluster_update_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+    end
+
+    it "does not remove the old overlap record" do
+      expect(update.overlap_table.count).to eq(2)
+      cfirst = Cluster.find_by(ocns: ht.ocns.first)
+      Clustering::ClusterHtItem.new(updated_item).cluster.tap(&:save)
+      cfirst = Cluster.where(_id: cfirst._id).first
+      # cfirst no longer has any ht_items
+      cu = described_class.new(update.overlap_table, cfirst)
+      # expect(cu.existing_overlaps.count).to eq(0)
```
mwarin, line 116: maybe don't commit these commented lines?

PR #133, opened 2021-09-20
Ht 2568 membercounts contd

mwarin:
Trying to figure out what's triggering the 

> Exceeded memory limit for $group, but didn't allow external sort. Pass allowDiskUse:true to opt in. (16945)

error when compiling a member counts report.

### DIFF
--- a/bin/reports/compile_member_counts_report.rb
+++ b/bin/reports/compile_member_counts_report.rb
@@ -14,7 +14,7 @@ def make_outf(output_dir)
   else
     FileUtils.mkdir_p(output_dir)
     ymd = Time.new.strftime("%F")
-    File.join(output_dir, "member_counts_#{ymd}.tsv")
+    fn = File.join(output_dir, "member_counts_#{ymd}.tsv")
     File.open(fn, "w")
   end
 end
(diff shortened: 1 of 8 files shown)


bin/reports/compile_member_counts_report.rb
```diff
@@ -14,7 +14,7 @@ def make_outf(output_dir)
   else
     FileUtils.mkdir_p(output_dir)
     ymd = Time.new.strftime("%F")
-    File.join(output_dir, "member_counts_#{ymd}.tsv")
+    fn = File.join(output_dir, "member_counts_#{ymd}.tsv")
```
aelkiss, line 17: Any reason why not `File.open(File.join(output_dir, "member_counts_#{ymd}.tsv"), "w")`?

aelkiss (commented):
Let's rebase & squash; then it looks good to me.

bin/reports/compile_member_counts_report.rb
mwarin: That's usually what the first iteration of my code looks like. 
I prefer more short & simple lines, where you can almost tell what it does without reading the whole thing, just by seeing the shape of it.

PR #135, opened 2021-09-24
Ht 3192 enumbug

mwarin:
Not a solution to the problem, but a change to make the warnings make more sense.

### DIFF
--- a/lib/enum_chron_parser.rb
+++ b/lib/enum_chron_parser.rb
@@ -206,7 +206,7 @@ class EnumChronParser
         matches.each do |m|
           # $stderr.puts "'#{m}'"
           add_to_chron(m[0])
-          input_str = input_str.gsub!(m[0], "")
+          input_str.gsub!(m[0], "")
         end
       end
       if input_str&.match?(MONTH_SPAN_RE)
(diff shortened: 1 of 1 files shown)


mwarin:
offending enumchrons listed in https://tools.lib.umich.edu/jira/browse/HT-3192

aelkiss (changes requested):
So this fix makes the warning more readable, but it doesn't prevent the underlying problem with the exception being thrown.

To be clear, I don't think we need to make it so that the offending enumchron is parseable in any meaningful way. However, I think we **should** try to avoid it throwing the `undefined method `gsub!' for nil:NilClass` exception internally. Looking at where that's happening (          input_str = input_str.gsub!(m[0], "")) it looks like the issue is that `input_str` is nil at that point. I don't really understand how that would happen just looking at the code (in particular I don't understand how a nil string could match the regexp, and I don't understand how the string could end up being transformed into nil), but I do think it's worth tracing that path with an offending enumchron.

aelkiss:
The problem is that:

```ruby
input.gsub!("not_in_string","replacement") 
```

returns `nil` if `input` does not contain `not_in_string`.

We should change this to either:

```ruby
input_str = input_str.gsub(m[0], "")
```

Or, probably my preference at this point given that we're already mutating `input_str` a bunch:

```ruby
input_str.gsub!(m[0], "")
```

There are several other places we're doing the same thing, so we should probably just review all uses of `gsub!` and make sure we are using it consistently and correctly.

mwarin:
> The problem is that:
> ...
> There are several other places we're doing the same thing, so we should probably just review all uses of `gsub!` and make sure we are using it consistently and correctly.

Took care of it.

aelkiss (approved):
LGTM 🌭

PR #137, opened 2021-10-06
Ht 3061 not all holdings in cost report

jsteverman:
We have been conflating "organizations" with "members". For our purposes, a "member" is a `ht_billing_member` with status = 1. 

Cost report calculations now filter out non-members.
   `organizations_with_holdings & Services.ht_organizations.members.keys`

Estimates add a "prospect_member" to overlap.matching_members for a given list of OCNs, so running estimates for non-members are unchanged. 

Overlap reports and the overlap table use any organization found in a cluster, so will include non-member HT organizations.


mwarin:
Well, there are still plenty of mentions of "member" and "institution" in the codebase, so I guess we can chip away at that some.

```
$ grep -inrHI member | wc -l
721
$ grep -inrHI institution | wc -l
16
$ grep -inrHI organization | wc -l
355
```

And with that I approve.

mwarin (approved):

PR #140, opened 2021-10-14
removed distinct ocns columns from member counts report script, updated tests

mwarin:


jsteverman:
I vote to remove, not comment out. Let git do its job.

PR #141, opened 2021-10-15
HT-3117, overlap namespacing

mwarin:


jsteverman (changes requested):
Some of the names seem a little redundant now, but better to be overdefined than ambiguous. As far as I'm concerned it just needs some commits squashed and it's good to go.

PR #143, opened 2021-10-22
HT-3213 delete export_overlap_report

jsteverman:
export_etas_overlap_report.rb generates reports for organizations in the format:

     ocn, local_id, item_type, rights, access

export_overlap_report.rb generates reports in the following format which is the structure of the holdings_htitem_htmember database table:

    lock_id, cluster_id, volume_id, n_enum, member_id, copy_count, brt_count, wd_count, lm_count, access_count

 

So those two aren't duplicative. However, the second appears to be unnecessary. The records generated are updated in the database directly by bin/update_overlap_table.rb nightly. There is also bin/reports/full_etas_overlap.rb which generates all of the records that would be inserted into holdings_htitem_htmember, but that has been used only once when we completely rebuilt the h_h_h table.

 

I am removing export_overlap_report.rb and lib/reports/overlap_report.rb.


mwarin (approved):
Looks legit.

PR #145, opened 2021-11-11
HT-3211: push job metrics to pushgateway

aelkiss:
- uses pushgateway via waypoint - this is probably not the final shape
we want, but it's a convenient entry point for now since both are
concerned with tracking progress. A better idea might be a progress
tracker object that delegates to waypoint but also reports to the
pushgateway.

- no specs for the pushgateway integration yet

- we probably want to label the job completion time with the expected
job frequency

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -166,6 +166,9 @@ Layout/EndAlignment:
 Layout/DefEndAlignment:
   EnforcedStyleAlignWith: def
 
+Metrics/ParameterLists:
+  CountKeywordArgs: false
+
 Performance/RedundantMerge:
   Enabled: false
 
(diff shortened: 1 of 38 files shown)


aelkiss:
I'd suggest looking at the two commits individually, since the second one is a bit noisy and perhaps some may be hard to follow. We can discuss if needed.

aelkiss:
This is ready for review now, including tracking the expected interval between completions. See description on the individual commits.

jsteverman (approved):
Actual changes are pretty limited. No objections.

aelkiss:
Great. @mwarin  if this looks good to you too (especially see the cleanup in autoscrub) I'll go ahead and merge, and (hopefully) we can see those metrics start showing up in prometheus. We'll need to set the expected completion interval via environment variable for each job (over in ht_tanka) before we would start getting alerts. Later down the road we can look as necessary at routing alerts for holdings jobs.

aelkiss:
@mwarin  I went ahead and merged this, but let us know if you have any feedback on the autoscrub stuff.

PR #146, opened 2021-11-18
Clusterable Shared Print Commitments and loader

jsteverman:
Doesn't address updates or deprecations, or even identity (there are a lot of duplicates in the existing database). 

My only concern before loading is the setting of `cluster.last_modified` when adding commitments, which seems unnecessary.

### DIFF
--- a/lib/clusterable/commitment.rb
+++ b/lib/clusterable/commitment.rb
@@ -4,19 +4,50 @@ require "mongoid"
 
 module Clusterable
 
-  # A commitment
+  # A shared print commitment
   class Commitment
     include Mongoid::Document
-    field :ocn, type: Integer # , type: OCLCNumber
+    field :uuid, type: String
     field :organization, type: String
+    field :ocn, type: Integer
+    field :local_id, type: String
+    field :oclc_sym, type: String
+    field :committed_date, type: DateTime
+    field :retention_date, type: DateTime
+    field :local_bib_id, type: String
+    field :local_item_id, type: String
+    field :local_item_location, type: String
+    field :local_shelving_type, type: String
+    field :policies, type: Array, default: []
+    field :facsimile, type: Boolean, default: false
+    field :other_program, type: String
+    field :other_retention_date, type: DateTime
+    field :deprecation_status, type: String
+    field :deprecation_date, type: DateTime
+    field :deprecation_replaced_by, type: String
 
     embedded_in :cluster
 
-    def move(new_parent)
-      unless new_parent.id == _parent.id
-        new_parent.commitments << dup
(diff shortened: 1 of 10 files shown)


aelkiss:
Agreed, we should not update `cluster.last_modified`.

aelkiss (commented):
This all looks good to me and in-line with what we're doing with other data types. I think we can go ahead and merge once the `last_updated` thing is taken care of. It would be useful to have feedback from @mwarin before going too much farther, but I think we can proceed for now.

lib/clusterable/commitment.rb
```diff
@@ -4,19 +4,50 @@
...
     field :organization, type: String
+    field :ocn, type: Integer
+    field :local_id, type: String
+    field :oclc_sym, type: String
+    field :committed_date, type: DateTime
+    field :retention_date, type: DateTime
+    field :local_bib_id, type: String
+    field :local_item_id, type: String
+    field :local_item_location, type: String
+    field :local_shelving_type, type: String
```
mwarin, line 20: This could be restricted to one of a small set of strings, if we care to.

lib/clusterable/commitment.rb
```diff
@@ -4,19 +4,50 @@
...
+    def matching_holdings
+      cluster = _parent
+      cluster.holdings.select {|h| h.organization == organization && h.local_id == local_id }
+    end
+
+    def batch_with?(other)
+      ocn == other.ocn
+    end
+
+    def deprecated?
```
mwarin, line 43: IMHO a commitment is deprecated if there is a deprecation_status AND a deprecated_date. Any state where one is set and the other isn't should be considered an error.

spec/loader/shared_print_loader_spec.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "loader/shared_print_loader"
+
+RSpec.describe Loader::SharedPrintLoader do
+  let(:json) do
+    '{"uuid":"e78047da-1193-43c7-aab0-492067d13f9c","organization":"ucsd","ocn":2,
```
mwarin, line 8: Break out and put in spec/fixtures/ ?

mwarin (approved):
A few comments but nothing that would prevent this PR to be merged.

lib/clusterable/commitment.rb
mwarin: "this" = local_shelving_type, which in the (human) spec doc is restricted to one of ['cloa','clca','sfca','sfcahm','sfcaasrs'].

PR #148, opened 2021-12-09
clusters now know more about commitments

mwarin:


lib/cluster.rb
```diff
@@ -120,6 +120,14 @@ def brt_counts
     @brt_counts
   end
 
+  def ch_counts
```
aelkiss, line 123: I'm not sure what status "ch" means -- consider changing the name of the method to reflect that? Likewise, consider changing `wd_counts` below to `withdrawn_counts`, and consider extracting a `counts_by_status` method perhaps?

lib/cluster.rb
```diff
@@ -169,6 +177,19 @@ def add_members_from(cluster)
...
+    if include_deprecated
+      commitments.any?
+    else
+      commitments.reject(&:deprecated?).any?
+    end
+  end
+
+  def eligible_for_commitments?
+    # be an spm with items and 1+ CH holdings
+    format == "spm" && ht_items.any? && holdings.any? && ch_counts.values.sum > 0
```
aelkiss, line 190: Is `ch_counts.values.sum > 0` redundant with `holdings.any?` -- that is, could `ch_counts.values.sum > 0` be true if `!holdings.any?`?

In order to make this easier to read, I'd consider extracting `ch_counts.values.sum > 0` as:

```ruby
def any_current_holdings?
  ch_counts.values.sum > 0
end
```

mwarin: I figured if `holdings.any? == false` we short circuit out and don't have to go into `ch_counts`. Haven't done the benchmark to back this up, tho.

aelkiss: Hmm.. If `holdings.any? == false` then it seems like `holdings_by_org` will also be empty (and likewise fast since `holdings.group_by(:&organization)` is just running on an empty array), so it shouldn't really be computing anything extra. Not really a big deal either way, but I'd still lean towards removing the check for `holdings.any?`.

mwarin: modified as suggested

lib/cluster.rb
mwarin: I think there are enough little reasons for not doing this _right now under the scope of HT-1944_.
* An empty status counts towards `ch_counts`, but not to `lm_counts` or `wd_counts`.
* `brt_counts` kind of looks the same but is counting something else.
* I could get rid of e.g. "wd" in `wd_counts` by making `withdrawn_counts` and having it pass `"wd"` as an argument to `counts_by_status`. I don't see the improvement. Is it just that the method names would make more sense?
* There are no tests for `(brt|lm|wd)_counts`.

aelkiss: That's fine -- yes, the main benefit is just that the method names would make more sense and that it would reduce duplication. Empty status counting towards `ch_counts` but not `lm_counts` or `wd_counts` is probably enough reason not to worry about extracting a method right now.

lib/reports/eligible_commitments.rb
```diff
@@ -0,0 +1,61 @@
...
+    def for_ocns(ocns = [])
+      if ocns.empty?
+        raise "No ocns given"
+      end
+
+      ocns.sort.uniq.each do |ocn|
+        cluster = Cluster.find_by(ocns: [ocn.to_i])
+
+        next if cluster.nil?
+        next if we_have_seen? cluster._id
```
aelkiss, line 31: This could happen because we might have OCNs that map to the same cluster, correct?

lib/reports/eligible_commitments.rb
```diff
@@ -0,0 +1,61 @@
...
+        next unless cluster.eligible_for_commitments?
+        next unless cluster.commitments.count.zero?
+
+        cluster.holdings.select(&:eligible_for_commitments?).each do |holding|
+          yield [
+            holding.organization,
```
aelkiss, line 37: For purposes of this report, do we care what HTID or cluster this commitment would correspond to?

spec/fixtures/organizations.rb
```diff
@@ -1,18 +1,25 @@
 # frozen_string_literal: true
 
+# To please rubocop.
+def neu(inst, country, weight, sym, status = 1)
```
aelkiss, line 4: This ends up leaking into the global namespace, which we should probably try to avoid, but you can use nested functions:

```ruby
def mock_organizations
  def neu(...)
    ...
  end
end
```

(I'd probably be inclined to call it `org` instead of `neu`, but that's just personal taste.)

lib/reports/eligible_commitments.rb
```diff
@@ -0,0 +1,61 @@
...
+      if ocns.empty?
+        raise "No ocns given"
+      end
+
+      ocns.sort.uniq.each do |ocn|
+        cluster = Cluster.find_by(ocns: [ocn.to_i])
+
+        next if cluster.nil?
+        next if we_have_seen? cluster._id
+        next unless cluster.eligible_for_commitments?
```
aelkiss, line 32: I'm not sure checking this up-front helps that much since we end up needing to iterate over holdings to determine this anyway.

lib/cluster.rb
```diff
@@ -169,11 +177,31 @@ def add_members_from(cluster)
...
+    end
+  end
+
+  def any_current_holdings?
+    @any_current_holdings ||= ch_counts.values.sum > 0
+  end
+
+  def eligible_for_commitments?
+    # be an spm with items and 1+ CH holdings
+    @eligible_for_commitments ||= format == "spm" && ht_items.any? && any_current_holdings?
```
aelkiss, line 194: I had written a comment about this, but I think it's fine as-is -- I can imagine a case where we might want to get a report of eligible clusters without finding the specific holdings; if we do find the specific holdings it's just one more iteration over the holdings (not N more, since this is memoized); and this should short-circuit if it's not an spm or there aren't any HTItems.

lib/clusterable/holding.rb
```diff
@@ -110,6 +110,18 @@ def update_key
...
+    def matches_commitment?
+      # check if there is a commitment on the same cluster
+      # with the same org & local_id
+      cluster.commitments.select do |spc|
+        spc.organization == organization && spc.local_id == local_id
+      end.any?
+    end
+
+    def eligible_for_commitments?
+      cluster.eligible_for_commitments? && (status.empty? || status == "CH")
```
aelkiss, line 122: Reaching up to the parent cluster here seems OK.

lib/reports/eligible_commitments.rb
mwarin: That's the idea.

lib/clusterable/holding.rb
```diff
@@ -110,6 +110,18 @@ def update_key
       as_document.slice(*(fields.keys - EQUALITY_EXCLUDED_FIELDS)).hash
     end
 
+    def matches_commitment?
+      # check if there is a commitment on the same cluster
+      # with the same org & local_id
+      cluster.commitments.select do |spc|
```
aelkiss, line 116: This is a little bit more questionable since here this introduces a dependency on the internal structure of `Commitment` that we didn't previously have, and reaches through the cluster to get them. You might consider something like:

```ruby

class Holding
  def matches_commitment?(commitment)
    commitment.organization == organization && commitment.local_id == local_id
  end
end

class Commitment
  def matches_holding?(holding)
    holding.organization == organization && holding.local_id = local_id
  end
end
```

and then defining a method on cluster that (essentially) joins them via merge-sort:
* sorts holdings & commitments by organization & local ID
* walks through the list of holdings & commitments to align them
* produce a combined list of holdings & commitments to operate on

(I haven't fully thought this through, obviously.)

This is a little bit of an awkward consequence of the mongodb model that would almost certainly be cleaner in a relational model.

aelkiss (commented):
Adding the various aggregations to `Cluster` seems OK conceptually to me; if we find ourselves adding a whole lot more of those we might consider whether we could factor out a `ClusterAggregations` class to handle this (perhaps with delegation from `Cluster`), just so that `Cluster` itself doesn't get extremely big. I definitely don't think that's worth it right now.

The only place that gives me pause conceptually is `Holding#matches_commitment?`; see comment in-line.

lib/reports/eligible_commitments.rb
mwarin: I don't think so, since commitments are on title level, sorta. We used to have resolved_oclc for this, but that's out now. Maybe cluster id, but that only makes sense if they are stable-ish.

aelkiss:
Might be worth thinking about pulling out the shared print business logic to a separate class that can know about the Cluster, Holdings, Commitment rather than sprinkling it between the three.

PR #151, opened 2021-12-10
Use milemarker gem instead of built-in waypoint

aelkiss:
- Changes waypoint name to marker
- Uses the milemarker API
- Verifies that `final_line` returns what milemarker returns, not what the pushgateway returns.


aelkiss:
FYI @billdueber

bin/delete_holdings_by_uuid.rb
```diff
@@ -19,11 +18,11 @@ def main
...
 
   ARGV.each do |path|
-    waypoint = Services.progress_tracker.new(1000)
+    marker = Services.progress_tracker.new(1000)
     inf = File.open(path, "r")
     inf.each_line do |line|
-      waypoint.incr
-      waypoint.on_batch do |wp|
+      marker.incr
+      marker.on_batch do |wp|
```
mwarin, line 25: maybe we don't want to call the variable wp anymore

jsteverman (approved):

bin/delete_holdings_by_uuid.rb
aelkiss: Good catch.

mwarin (approved):
But for my comment about the variable name "wp", I approve.

aelkiss:
> But for my comment about the variable name "wp", I approve.

Great; I pushed a commit changing that just to `m`; I'll merge assuming specs all pass.

PR #153, opened 2021-12-14
delete specified holdings and any empty clusters that result

mwarin:


bin/holdings_deleter.rb
```diff
@@ -0,0 +1,132 @@
...
+
+# Takes simple pairs of key=>value and constructs a update_many...pull query
+# based on those criteria.
+# Example:
+# bundle exec ruby bin/delete_matching_holdings.rb organization foo date_received 2020-01-01
+class HoldingsDeleter
+  attr_reader :matching_criteria
+
+  def initialize(args)
+    @matching_criteria = {}
```
jsteverman, line 37: I think this can be more opinionated. Primary motivation for this is deleting holdings by org right? No need to make things flexible if we don't have other use cases.

bin/holdings_deleter.rb
```diff
@@ -0,0 +1,132 @@
...
+
+  def run
+    if @noop
+      puts "noop!"
+      return nil
+    end
+
+    result = Cluster.collection.update_many(
+      {},
+      { "$pull": { "holdings": { "$and": [@matching_criteria] } } }
```
jsteverman, line 64: $pull should definitely be faster. Another option would be 
`Cluster.where( <criteria>).no_timeout.each do |cluster|
    ...delete the necessary holdings...
    ...delete the cluster if it's empty...
`

bin/holdings_deleter.rb
```diff
@@ -0,0 +1,132 @@
...
+    Cluster.collection.aggregate(
+      [
+        { "$match": { "ht_items.0":        { "$exists": 0 } } },
+        { "$match": { "holdings.0":        { "$exists": 0 } } },
+        { "$match": { "commitments.0":     { "$exists": 0 } } },
+        { "$match": { "ocn_resolutions.0": { "$exists": 0 } } }
+      ]
+    ).each do |empty_cluster|
+      putsif "Deleting empty cluster: #{empty_cluster}"
+      Cluster.find_by(ocns: empty_cluster["ocns"]).delete
```
jsteverman, line 85: `Cluster.where({"$and":[{"ht_items.0": {"$exists": 0}},

                                      {"holdings.0":{"$exists":0}},

                                      {"commitments.0":{"$exists":0}},

                                     {"ocn_resolutions.0":{"$exists":0}}]

                                 }).each do |cluster|

  cluster.delete

end`

mwarin: that's the stuff

bin/holdings_deleter.rb
mwarin: There is already need for doing this by `org+date_received`, and we're going to have to do it by `org+mono_multi_serial` (if the past is any indication of the future). At least. And at that point we'd have half a Swiss army knife.

aelkiss:
I think it would benefit from some documentation to indicate how to use this.

jsteverman (approved):
OptionParser might clean this up a little bit, but not strictly necessary. 

It could use some additional inline documentation, and a mention in the README.

spec/holdings_deleter_spec.rb
```diff
@@ -0,0 +1,131 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+require_relative "../bin/holdings_deleter"
+
+RSpec.describe HoldingsDeleter do
+  def neu(*args)
```
aelkiss, line 8: FWIW the general preference for rspec style seems to be to use `described_class.new` for this case (to avoid repeating the class name and to make it slightly easier if you want to change the class name) and not to worry too much about the extra verbosity. Another alternative if you don't like `described_class` might be something like:

```ruby
let(:hd) { HoldingsDeleter }

it "does something" do
  hd.new(...)
end
```

aelkiss (commented):
I think it would make sense to at least split out option parsing from the actual deletion (e.g. `HoldingsDeleter` should get a pre-parsed set of criteria).

While I agree `OptionParser` would simplify this I don't know if we need to do it right now if this works -- eventually I think we should consider a common CLI front-end for all our various scripts ([thor](https://rubygems.org/gems/thor) is the most popular, but there are a couple other options e.g. [gli](https://rubygems.org/gems/gli))

spec/holdings_deleter_spec.rb
mwarin: Oh I like that 2nd version.

mwarin:
Replaced homemade option parsing with optparse.

jsteverman (approved):

PR #154, opened 2021-12-22
Ht 3237 ocn reclustering avoidance

jsteverman:
It doesn't go as far as I would like and some of the tests need work, but I wanted to get this out before break.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -61,7 +61,7 @@ Metrics/ClassLength:
     - 'lib/scrub/autoscrub.rb'
     - 'lib/scrub/scrub_fields.rb'
 
-# Offense count: 6
+# Offense count: 7
 # Configuration parameters: CountComments, ExcludedMethods.
 Metrics/MethodLength:
   Exclude:
@@ -69,7 +69,7 @@ Metrics/MethodLength:
     - 'lib/scrub/autoscrub.rb'
     - 'lib/enum_chron_parser.rb'
     - 'lib/scrub/scrub_fields.rb'
-
+    - 'lib/clustering/cluster_ht_item.rb'
 
 # Offense count: 1
 Naming/AccessorMethodName:
--- a/lib/cluster.rb
+++ b/lib/cluster.rb
@@ -176,4 +176,15 @@ class Cluster
   def empty?
     ht_items.empty? && ocn_resolutions.empty? && holdings.empty? && commitments.empty?
   end
+
+  def update_ocns
+    self.ocns = [component_ocns].flatten.uniq
+    save
+  end
+
+  def component_ocns
+    @component_ocns ||= ocn_resolutions.pluck(:ocns) + ht_items.pluck(:ocns) +
+      holdings.pluck(:ocn) + commitments.pluck(:ocn)
+  end
+
 end
--- a/lib/cluster.rb
+++ b/lib/cluster.rb
@@ -178,12 +178,12 @@ class Cluster
   end
 
   def update_ocns
-    self.ocns = [component_ocns].flatten.uniq
+    self.ocns = [clusterable_ocn_tuples].flatten.uniq
     save
   end
 
-  def component_ocns
-    @component_ocns ||= ocn_resolutions.pluck(:ocns) + ht_items.pluck(:ocns) +
+  def clusterable_ocn_tuples
+    @clusterable_ocn_tuples ||= ocn_resolutions.pluck(:ocns) + ht_items.pluck(:ocns) +
       holdings.pluck(:ocn) + commitments.pluck(:ocn)
   end
 
--- a/lib/clustering/ocn_graph.rb
+++ b/lib/clustering/ocn_graph.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+require "cluster"
+
+module Clustering
+
+  # Constructs a graph from OCN tuples
+  class OCNGraph
+
+    attr_accessor :edges, :nodes
+
+    def initialize(cluster = nil)
+      @cluster = cluster
+      @nodes = Set.new
+      # root_ocn : [all matching ocns]
+      @edges = Hash.new {|g, root| g[root] = Set.new }
+      compile_graph_from_cluster unless @cluster.nil?
+    end
+
+    # Add the tuple of OCNs to the list of nodes and edges
+    #
+    # @param [Enumerable] tuple A set of one or more OCNs from a clusterable
+    def add_tuple(tuple)
+      nodes.merge(tuple)
+      tuple.map {|node| edges[node] << node }
+      tuple.sort.to_a.combination(2).each do |parent, child|
+        edges[parent] << child
+        edges[child] << parent
+      end
+    end
+
+    # root, children = edges.sort.first
+
+    # Partition the :nodes into one or more subgraphs
+    #
+    # TODO: better name. really just subsets of OCNS
+    def subgraphs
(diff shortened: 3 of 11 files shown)


.rubocop_todo.yml
```diff
@@ -59,15 +59,15 @@ Metrics/ClassLength:
...
+# Offense count: 7
 # Configuration parameters: CountComments, ExcludedMethods.
 Metrics/MethodLength:
   Exclude:
     - 'bin/**/*.rb'
     - 'lib/scrub/autoscrub.rb'
     - 'lib/enum_chron_parser.rb'
     - 'lib/scrub/scrub_fields.rb'
-
+    - 'lib/clustering/cluster_ht_item.rb'
```
aelkiss, line 70: Might be a good time to try `standardrb` instead of rubocop, which disables most of the complexity metrics.

aelkiss:
After looking a bit, I'd suggest:

- put this in a separate tagged test we don't run by default (because it will be slow)
- set the mongodb session timeout to 1 minute: https://docs.mongodb.com/manual/reference/command/startSession/ -- AFAICT the timeout can't be set below that
- without timeout avoidance set, run an artificially slow mongodb query (https://stackoverflow.com/questions/12908425/simulate-slow-query-in-mongodb) and verify that you get the missing cursor error: run an artificially slow mongodb query: 
- with timeout avoidance set, run the same query and verify that it returns successfully

aelkiss (commented):
I started looking at this; I still need to look at the graph traversal and the tests.

I suspect that if `ocns_changed?` is true (i.e. one of the OCNs in the removed clusterable was the last witness to that OCN in the cluster) that we don't need to recluster -- suppose we have a clusterable with `[a,b]` and there's nothing else with `b` in it in the cluster -- in the graph of OCNs, `b` only has an edge to `a`, so removing `b` still leaves a graph with one connected component.

To spell out the case for needing graph traversal more generally to determine whether we need to recluster: say that we have a cluster with an HTItem with OCNs `[a,d]` and concordance rules `[a,b], [b,c], [c,d]`. There aren't any other clusterables that contain `a` and `d` directly, but `a` and `d` are still part of the connected component; if we had this same cluster, but without the `[b,c]` concordance rule, then we *would* need to recluster when removing an HTItem with OCNs `[a,d]`.

lib/clustering/reclusterer.rb
```diff
@@ -31,8 +37,41 @@ def recluster
...
+      (@removed_ocns - @cluster.component_ocns).any?
+    end
+
+    # Determines if all component OCNs are connected.
+    # Uses simple tests before attempting more involved graph traversal.
+    def needs_recluster?
+      return false if @cluster.ocns.one? ||
+        resolution_includes_cluster_ocns? ||
+        htitem_includes_all_cluster_ocns? ||
+        cluster_components.one?
```
aelkiss, line 53: I'm not sure this one saves us much time - if there's only one thing in the cluster, then the graph traversal will be fast anyway; if there are more things, then it just means we're iterating over items in the cluster an extra time.

lib/cluster.rb
```diff
@@ -176,4 +176,15 @@ def large?
...
     ht_items.empty? && ocn_resolutions.empty? && holdings.empty? && commitments.empty?
   end
+
+  def update_ocns
+    self.ocns = [component_ocns].flatten.uniq
+    save
+  end
+
+  def component_ocns
+    @component_ocns ||= ocn_resolutions.pluck(:ocns) + ht_items.pluck(:ocns) +
```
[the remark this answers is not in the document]
jsteverman: I think I originally intended for this to be used by `OCNGraph.new.compile_graph_from_cluster` and then rewrote it there anyway. As written, `ocns_changed?` is bugged because it returns true if removed_ocns is a subset of a component's ocns tuple, e.g. ([3] - [1, [2, 3]]).any? is true. I wrote a failing test and then corrected it in ocns_changed?. 

"I suspect that if ocns_changed? is true (i.e. one of the OCNs in the removed clusterable was the last witness to that OCN in the cluster) that we don't need to recluster"
The implementation of ocns_changed? here implies just the opposite since false means there exists an exact duplicate tuple in the cluster. Corrected, ocns_changed? is true is insufficient to determine the need for reclustering because removed tuples can contain more than 2 OCNs. Given a removed tuple of [b, c, e], the first two OCNs can be the glue, and the third the last witness requiring an update to the cluster.ocns.

lib/clustering/reclusterer.rb
```diff
@@ -31,8 +37,41 @@ def recluster
...
+
     private
 
+    # By definition, if a cluster has an OCLC resolution and only 2 OCNs then
+    # the cluster is coherent.
+    def resolution_includes_cluster_ocns?
+      @cluster.ocns.count == 2 && @cluster.ocn_resolutions.one?
+    end
+
+    def htitem_includes_all_cluster_ocns?
```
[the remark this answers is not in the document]
jsteverman: `removed_ocn_tuple_equals_current_resolution?` || `removed_ocn_tuple_is_subset_of_ht_item?`

lib/cluster.rb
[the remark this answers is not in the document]
aelkiss: Good point, I wasn't thinking about cases of > 2 OCNs.

lib/clustering/ocn_graph.rb
```diff
@@ -0,0 +1,73 @@
...
+        edges[child] << parent
+      end
+    end
+
+    # root, children = edges.sort.first
+
+    # Partition the :nodes into one or more subgraphs
+    #
+    # TODO: better name. really just subsets of OCNS
+    def subgraphs
```
aelkiss, line 37: Is there an opportunity for extracting any graph-related functionality from https://github.com/hathitrust/holdings-backend/blob/main/lib/concordance_validation/concordance.rb#L69 ? It may also be worth taking a look at https://github.com/monora/rgl. It looks like `add_tuple` will result in strongly-connected components, so https://ruby-doc.org/stdlib-3.0.1/libdoc/tsort/rdoc/TSort.html might also work.

jsteverman: Concordance validation needs some attention on its own, if only because we don't actually do anything about the warnings it generates. It's also a directed graph making it a candidate for TSort while our clustering is not. RGL looks like a promising replacement for OCNGraph

aelkiss: "directed graph" - true, but you can model an undirected graph as a directed graph with two edges for each pair of vertices connected in the undirected graph, and I believe the notion of strongly connected component on such a directed graph corresponds to the notion of connected component in the undirected graph. I had thought we were already doing that because of

```ruby
        edges[parent] << child
        edges[child] << parent
```

aelkiss:
@mwarin  Take a look and if it's confusing we can find a time early next week to review together with @jsteverman

lib/clustering/ocn_graph.rb
```diff
@@ -0,0 +1,36 @@
...
+    # Add the tuple of OCNs to the list of edges
+    #
+    # @param [Enumerable] tuple A set of one or more OCNs from a clusterable
+    def add_tuple(tuple)
+      add_vertices(*tuple)
+      tuple.sort.to_a.combination(2).each {|parent, child| add_edge(parent, child) }
+    end
+
+    # Partition the vertices into one or more components
+    def components
```
mwarin, line 27: Maybe rename "components" to something less clashy with other uses of "component".

mwarin (approved):

PR #156, opened 2022-01-07
New class Utils::SessionKeepAlive: for keeping sessions alive

mwarin:
Broke out some code from `lib/overlap/overlap_table_update.rb` into new class `lib/utils/session_keep_alive.rb`, since the same thing was needed to keep the holdings deleter from crashing with an expired cursor.

@jsteverman and I tried to think what a (useful && non-production) test for this would look like. `overlap_table_update.rb` does not directly test this behavior.

Any ideas from @aelkiss or the Universe appreciated.

### DIFF
--- a/lib/utils/session_keep_alive.rb
+++ b/lib/utils/session_keep_alive.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+require "services"
+require "cluster"
+
+Services.mongo!
+
+module Utils
+  # Lets you wrap a session around a potentially long running query,
+  # and that session is kept alive by a refresh-loop in a separate thread.
+  # The actual test of keeping session/cursor alive is TODO.
+  # Intended usage:
+  # require "utils.session_keep_alive"
+  # ska = Utils::SessionKeepAlive.new(60)
+  # ska.run do
+  #   # Mongo operation at risk of timing out:
+  #   Cluster.where(long_q).no_timeout.each do |cluster|
+  #     {...}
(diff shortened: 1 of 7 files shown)


aelkiss:
After looking a bit, I'd suggest:

- put this in a separate tagged test we don't run by default (because it will be slow)
- set the mongodb session timeout to 1 minute: https://docs.mongodb.com/manual/reference/command/startSession/ -- AFAICT the timeout can't be set below that
- without timeout avoidance set, run an artificially slow mongodb query (https://stackoverflow.com/questions/12908425/simulate-slow-query-in-mongodb) and verify that you get the missing cursor error: run an artificially slow mongodb query: 
- with timeout avoidance set, run the same query and verify that it returns successfully

lib/utils/session_keep_alive.rb
```diff
@@ -0,0 +1,52 @@
...
+  # Intended usage:
+  # require "utils.session_keep_alive"
+  # ska = Utils::SessionKeepAlive.new(60)
+  # ska.run do
+  #   # Mongo operation at risk of timing out:
+  #   Cluster.where(long_q).no_timeout.each do |cluster|
+  #     {...}
+  #   end
+  # end
+  class SessionKeepAlive
```
aelkiss, line 20: Worth adding a comment about how we tried to test that this fixes the disappearing-cursor issue but failed to successfully test that.

jsteverman (approved):
Looks good.

PR #157, opened 2022-01-10
Validate gzip AND txt files found in conc/raw

jsteverman:


bin/concordance_validation/validate_and_delta.rb
```diff
@@ -25,9 +25,9 @@ def review_logs(log_file)
 
 conc_dir = Settings.concordance_path
 
-raw_gzip_files = Dir.glob("#{conc_dir}/raw/*txt.gz").sort
+raw_files = Dir.glob("#{conc_dir}/raw/*txt*").sort
```
aelkiss, line 28: Any concerns about this matching e.g. "this_is_not_a_txt_file"? I figure we control what goes in that directory, so it shouldn't be a big issue?

bin/concordance_validation/validate_and_delta.rb
```diff
@@ -25,9 +25,9 @@ def review_logs(log_file)
...
 
-file_names_split = raw_gzip_files.map {|fname| fname.split("/").last.split("_").first }
+file_names_split = raw_files.map {|fname| fname.split("/").last.split("_").first }
 raw_dates = file_names_split.select {|d| d =~ /^\d+$/ }
```
aelkiss, line 31: Might be more clear here to use a regex here -- if nothing else, it would be worth a comment here as to what filename format this is expecting.

mwarin (approved):

bin/concordance_validation/validate_and_delta.rb
jsteverman: No more than it matching "junk_that_shouldnt_be_in_this_directory.txt.gz". If it gets garbage it will end up erroring out anyway.

bin/concordance_validation/validate_and_delta.rb
jsteverman: Commented expected file format.

jsteverman: And now with regex instead of split("_").first

aelkiss (approved):

PR #160, opened 2022-02-01
Ht 1944 ii

mwarin:
Cleanup of what was in branch `HT-1944`.

### DIFF
--- a/lib/overlap/holding_commitment.rb
+++ b/lib/overlap/holding_commitment.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+require "cluster"
+require "services"
+
+Services.mongo!
+
+module Overlap
+  # Given an ocn, gets:
+  # * non-deprecated commitments in the matching cluster
+  # * holdings eligible for commitments in the matching cluster
+  class HoldingCommitment
+    def initialize(ocn)
+      # For holding-commitment overlaps we only care about spm clusters with ht_items
+      @cluster = Cluster
+        .where(ocns: ocn, "ht_items.0": { "$exists": 1 })
+        .find {|c| c.format == "spm" }
+    end
+
+    # Restrict holdings to ones that are eligible for commitments.
+    def eligible_holdings
+      @eligible_holdings ||= @cluster&.holdings&.select do |x|
+        eligible_holding?(x)
+      end
+      @eligible_holdings ||= []
+    end
+
+    # Restrict commitments to non-deprecated.
+    def active_commitments
+      @active_commitments ||= @cluster&.commitments&.reject(&:deprecated?)
+      @active_commitments ||= []
+    end
+
+    private
+
+    # For a holding to be considered for commitments, it must be:
+    # a currently held single part monograph that isn't brittle.
+    def eligible_holding?(hol)
+      hol.mono_multi_serial == "mono" &&
+        (hol.status.empty? || hol.status == "CH") &&
+        hol.condition.empty?
+    end
+
+  end
+end
--- a/lib/reports/eligible_commitments.rb
+++ b/lib/reports/eligible_commitments.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require "services"
+require "overlap/holding_commitment"
+
+module Reports
+  # Given criteria, pull up all clusters that  match those criteria
+  # AND have holdings,
+  # AND have ht_items,
+  # AND have no commitments,
+  # AND are eligible for commitments.
+  # Invoke via bin/reports/compile_eligible_commitments_report.rb
+  class EligibleCommitments
+    def header
+      ["organization", "oclc_sym", "ocn", "local_id"]
+    end
+
+    def for_ocns(ocns = [])
+      if ocns.empty?
+        raise "No ocns given"
+      end
+
+      ocns.sort.uniq.each do |ocn|
+        overlap = Overlap::HoldingCommitment.new(ocn)
+        if overlap.active_commitments.empty?
+          overlap.eligible_holdings.each do |h|
+            yield [
+              h.organization,
+              Services.ht_organizations.members.fetch(h.organization)&.oclc_sym || "N/A",
+              h.ocn,
+              h.local_id
+            ]
+          end
+        end
+      end
+    end
+
+  end
+end
--- a/spec/overlap/holding_commitment_spec.rb
+++ b/spec/overlap/holding_commitment_spec.rb
@@ -1,0 +1,143 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "overlap/holding_commitment"
+require "clustering/cluster_commitment"
+
+RSpec.describe Overlap::HoldingCommitment do
+  let(:ocn_1) { 111 }
+  let(:ocn_2) { 222 }
+  let(:local_id_1) { "i111" }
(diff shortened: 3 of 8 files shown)


lib/overlap/holding_commitment.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    # How many distinct orgs have reported commitments for the given cluster?
+    def commitments_h
+      @commitments.pluck(:organization).uniq.count
+    end
+
+    # Return only holdings that could have a commitment.
+    def filter_holdings(holdings)
+      holdings.select {|x| x.mono_multi_serial == "mono" && (x.status.empty? || x.status == "CH") }
```
aelkiss, line 41: Might be worth some more explanation in comments here -- if the member doesn't report it as `mono_multi_serial` then something has gone wrong with their data; `CH` status is current holding & we don't want a commitment on something lost/missing/withdrawn.

lib/overlap/holding_commitment.rb
```diff
@@ -0,0 +1,81 @@
...
+        .sort_by(&:organization)
+    end
+
+    # Return only non-deprecated commitments.
+    def filter_commitments(commitments)
+      commitments.select {|x| x.deprecation_status.nil? }.sort_by(&:organization)
```
aelkiss, line 47: `commitments.reject(:&deprecation_status)`

lib/overlap/holding_commitment.rb
```diff
@@ -0,0 +1,81 @@
...
+        # e.g. same org & ocn but diff local_id,
+        # or   same org & local_id but diff ocn
+        hol.organization == commitment.organization &&
+          hol.ocn        == commitment.ocn          &&
+          hol.local_id   == commitment.local_id
+      end
+    end
+
+    # Return matching pairs of [com, hol].
+    def matched_pairs
```
aelkiss, line 63: Not used outside specs

lib/overlap/holding_commitment.rb
```diff
@@ -0,0 +1,81 @@
...
+      Cluster.where(ocns: ocn, "ht_items.0": { "$exists": 1 }).each do |cluster|
+        if cluster.format == "spm"
+          @holdings    = filter_holdings(cluster.holdings)
+          @commitments = filter_commitments(cluster.commitments)
+        end
+      end
+    end
+
+    # How many distinct orgs have reported holdings for the given cluster?
+    def holdings_h
```
aelkiss, line 30: Not used outside specs

lib/overlap/holding_commitment.rb
```diff
@@ -0,0 +1,81 @@
...
+      @holdings.pluck(:organization).uniq.count
+    end
+
+    # How many distinct orgs have reported commitments for the given cluster?
+    def commitments_h
```
aelkiss, line 35: Not used outside specs

lib/overlap/holding_commitment.rb
```diff
@@ -0,0 +1,81 @@
...
+      @commitments.pluck(:organization).uniq.count
+    end
+
+    # Return only holdings that could have a commitment.
+    def filter_holdings(holdings)
```
aelkiss, line 40: `eligible_holdings`, perhaps?

lib/overlap/holding_commitment.rb
```diff
@@ -0,0 +1,81 @@
...
+      holdings.select {|x| x.mono_multi_serial == "mono" && (x.status.empty? || x.status == "CH") }
+        .sort_by(&:organization)
+    end
+
+    # Return only non-deprecated commitments.
+    def filter_commitments(commitments)
```
aelkiss, line 46: `active_commitments`?

lib/reports/eligible_commitments.rb
```diff
@@ -0,0 +1,45 @@
...
+      ["organization", "oclc_sym", "ocn", "local_id"]
+    end
+
+    def for_ocns(ocns = [])
+      if ocns.empty?
+        raise "No ocns given"
+      end
+
+      ocns.sort.uniq.each do |ocn|
+        overlap = Overlap::HoldingCommitment.new(ocn)
```
aelkiss, line 21: Consider merging what we need from `Overlap::HoldingCommitment` back to this class?

spec/overlap/holding_commitment_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  it "finds all matching pairs" do
+    make_cluster(ocn_1, org_1, local_id_1)
+    make_cluster(ocn_1, org_2, local_id_2)
+    overlap = described_class.new(ocn_1)
+    expect([overlap.commitments.size, overlap.holdings.size, overlap.matched_pairs.size])
+      .to eq [2, 2, 2]
+    expect([overlap.holdings_h, overlap.commitments_h]).to eq [2, 2]
+  end
+
```
aelkiss, line 73: it "ignores deprecated commitments"

spec/reports/eligible_commitments_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+    magic_number = 50
+    1.upto(magic_number) do |i|
+      Clustering::ClusterHolding.new(build_h("umich", i, "i#{i}", "CH")).cluster.tap(&:save)
+      Clustering::ClusterHtItem.new(build(:ht_item, :spm, ocns: [i])).cluster.tap(&:save)
+    end
+    rows = run((1..magic_number).to_a)
+    expect(rows.size).to be magic_number
+  end
+
+  it "Ignores clusters that already have commitments" do
```
aelkiss, line 69: Move these specs to the cluster holding class?

spec/reports/eligible_commitments_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+  end
+
+  it "Ignores clusters where format is ser" do
+    Clustering::ClusterHtItem.new(ht_ser).cluster.tap(&:save)
+    Clustering::ClusterHolding.new(h_ch).cluster.tap(&:save)
+    rows_ser = run([ocn1, ocn2])
+    expect(rows_ser.count).to eq 0
+  end
+
+  it "Translates organization to oclc_symbol" do
```
aelkiss, line 105: leave this one?

lib/reports/eligible_commitments.rb
mwarin: I think we landed on maybe not, since other similar Qs about commitments will arise outside the scope of this report.

spec/overlap/holding_commitment_spec.rb
mwarin: added, and will work when rebased with @jsteverman's deprecate() bugfix

lib/reports/eligible_commitments.rb
aelkiss: :+1:

aelkiss (approved):
I agree all the feedback from yesterday is addressed; looks good now.

I think we should go ahead and merge this, but I do think there's one additional need as stated in HT-1944, which is the ability to do this report for all items in the digital library, not just a given set of OCNs. I think that can be a separate PR, though -- I'd see that likely resulting in a small small change to `Reports::EligibleCommitments` to allow doing the report on all clusters, and perhaps changing `Overlap::HoldingCommitment` to take a cluster instead of an OCN.

PR #163, opened 2022-02-04
Ht 3095 automate etas overlap

jsteverman:
bin/add_print_holdings.rb and bin/reports/export_etas_overlap_report.rb will fail in production.

They need the environment variable RCLONE_CONFIG_PATH set to rclone.conf secret?

### DIFF
--- a/bin/add_print_holdings.rb
+++ b/bin/add_print_holdings.rb
@@ -13,6 +13,11 @@ def main
   holding_loader = Loader::HoldingLoader.for(filename)
   Loader::FileLoader.new(batch_loader: holding_loader).load(filename, skip_header_match: /\A\s*OCN/)
   holding_loader.final_line
+
+  organization = holding_loader.instance_variable_get("@organization")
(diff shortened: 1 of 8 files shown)


bin/add_print_holdings.rb
```diff
@@ -4,6 +4,7 @@
 require "loader/file_loader"
 require "loader/holding_loader"
 require "services"
+require "pry"
```
mwarin, line 7: should prolly remove?

lib/reports/etas_organization_overlap_report.rb
```diff
@@ -115,6 +116,37 @@ def write_records_for_unmatched_holdings(cluster, holdings_matched)
         write_record(etas_record)
       end
     end
+
```
mwarin, line 119: I feel this could be broken out as a general document uploader class, since more reports might want to get uploaded.

mwarin (approved):

PR #164, opened 2022-02-08
Ht 3265

mwarin:
Code for taking deprecation requests, finding the matching commitment, deprecating them and logging it all to a report that can be sent back to the originating organization.

### DIFF
--- a/spec/shared_print/deprecation_record_spec.rb
+++ b/spec/shared_print/deprecation_record_spec.rb
@@ -1,0 +1,161 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "shared_print/deprecation_record"
+
+RSpec.describe SharedPrint::DeprecationRecord do
+  let(:org) { "umich" }
+  let(:org_2) { "smu" }
+  let(:ocn) { 111 }
+  let(:ocn_that_should_fail) { -999 }
+  let(:loc) { "i123" }
+  let(:loc_2) { "i222" }
+  let(:sta) { "E" }
+  let(:status_that_should_fail) { "X" }
+  let(:line) { [org, ocn, loc, sta].join("\t") }
+  let(:empty_cluster) { create(:cluster) }
+  let(:dep) { described_class.new(organization: org, ocn: empty_cluster.ocns.first, local_id: loc, status: sta) }
+
+  def make_commitment(ocn, org, local_id)
+    # Make and cluster a commitment
+    com = build(:commitment, ocn: ocn, organization: org, local_id: local_id)
+    cluster = Clustering::ClusterCommitment.new(com).cluster.tap(&:save)
+    cluster.commitments.where(uuid: com.uuid).first
+  end
+
+  before(:each) do
+    Cluster.collection.find.delete_many
+  end
+
+  describe "#initialize" do
+    it "generates a #{described_class} given valid args" do
+      expect do
+        described_class.new(
+          organization: org,
+          ocn: ocn,
+          local_id: loc,
+          status: sta
+        )
+      end.not_to raise_error
+    end
+
+    it "raises an error if given any invalid args (currently only status can cause an error to be raised)" do
+      expect do
+        described_class.new(
+          organization: org,
+          ocn: ocn,
+          local_id: loc,
+          status: status_that_should_fail
+        )
+      end.to raise_error ArgumentError
+    end
+  end
+
+  describe "#to_s" do
+    it "stringifies the way we expect" do
+      expect(dep.to_s).to eq "ocn:#{empty_cluster.ocns.first}, organization:#{org}, local_id:#{loc}, status:#{sta}"
+    end
+  end
+
+  describe "#self.parse_line" do
+    it "parses a line into a #{described_class} if given a line that can be parsed" do
+      line = [org, ocn, loc, sta].join("\t")
+      parsed = described_class.parse_line(line)
+      expect(parsed).to be_a described_class
+      expect(parsed.ocn).to eq ocn
+      expect(parsed.organization).to eq org
+      expect(parsed.local_id).to eq loc
+      expect(parsed.status).to eq sta
+    end
+
+    it "raises an error if the given line contains elements that cause validate to raise an error" do
+      line_that_should_fail = [org, ocn, loc, status_that_should_fail].join("\t")
+      expect { described_class.parse_line(line_that_should_fail) }.to raise_error ArgumentError
+    end
+  end
+
+  describe "#cluster" do
+    it "finds a cluster if there is a cluster to find" do
+      expect(dep.cluster).to be_a Cluster
+      expect(dep.err.size).to eq 0
+    end
+
+    it "does not find a cluster if there is no matching cluster, and notes this in @err" do
+      dep = described_class.new(ocn: ocn_that_should_fail, status: sta)
+      expect(dep.cluster).to eq nil
+      expect(dep.err).to eq ["No cluster found for OCN:#{ocn_that_should_fail}"]
+    end
+  end
+
+  describe "#commitments" do
+    it "finds commitments if there are commitments on the cluster" do
+      dep =  described_class.new(organization: org, ocn: ocn, local_id: loc, status: sta)
+      make_commitment(ocn, org, loc)
+      expect(dep.commitments.size).to eq 1
+      expect(dep.commitments.first).to be_a Clusterable::Commitment
+      expect(dep.err.size).to eq 0
+    end
+
+    it "does not find commitments if there are no commitments on the cluster" do
+      dep =  described_class.new(organization: org, ocn: empty_cluster.ocns.first, local_id: loc, status: sta)
+      expect(dep.commitments.size).to eq 0
+      expect(dep.err.size).to eq 1
+    end
+  end
+
+  describe "#org_commitments" do
+    it "finds commitments that belong to its organization if there are any in its cluster" do
+      comm = make_commitment(ocn, org, loc)
+      dep = described_class.new(organization: org, ocn: ocn, local_id: loc, status: sta)
+      expect(dep.org_commitments).to eq [comm]
+      expect(dep.err.size).to eq 0
+    end
+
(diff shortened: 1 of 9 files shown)


bin/commitments_deprecater.rb
```diff
@@ -0,0 +1,190 @@
...
+    end
+  end
+
+  # Write log message to file, and STDERR if @verbose
+  def log(msg)
+    if @report.nil?
+      # To reduce the risk of accidental overwrites by a different instance,
+      # make the filename rather unique-ish.
+      iso_stamp = Time.now.strftime("%Y%m%d-%H%M%S")
+      rand_str  = SecureRandom.hex(8)
```
aelkiss, line 38: `ENV['HOSTNAME']` might work. We should also discuss the use cases for these logs & if just logging to stdout & reviewing later as needed with loki would be sufficient.

bin/commitments_deprecater.rb
```diff
@@ -0,0 +1,190 @@
...
+    @header_spec = ["organization", "ocn", "local_id", "deprecation_status"]
+  end
+
+  # Process a list of files.
+  def run(args)
+    args.each do |arg|
+      if File.file?(arg)
+        from_file(arg)
+      else
+        log "arg #{arg} not a file, skipping"
```
aelkiss, line 27: Would we normally expect this? Is it OK just to get a file not found exception and terminate if given something that isn't a file?

bin/commitments_deprecater.rb
aelkiss: Either way, we should probably consider logging via `Services.logger` rather than defining a custom logger here.

bin/commitments_deprecater.rb
```diff
@@ -0,0 +1,190 @@
...
+require "services"
+require "deprecation_record"
+
+Services.mongo!
+
+# Takes (shared print) deprecation records, and tries to match them to commitments
+# so that those commitments can be deprecated.
+# Usage: CommitmentsDeprecater.new.run(<list_of_files>)
+
+class CommitmentsDeprecater
```
aelkiss, line 13: We probably want to move this to `lib`, but I'm not sure what subdirectory to put it in (maybe a new `shared_print` directory?) This isn't exactly a loader like the others..

bin/commitments_deprecater.rb
```diff
@@ -0,0 +1,190 @@
...
+    end
+    @report.puts msg
+    warn msg if @verbose
+  end
+
+  # Read deprecation records from file and try to deprecate the matching commitments.
+  def from_file(path)
+    inf = File.open(path, "r")
+    log "Reading deprecation records from #{path}"
+    if !check_header(inf.gets.strip)
```
aelkiss, line 53: This is similar to `Loader::FileLoader.load`; consider adding an option to require the header there (rather than skipping it if it is present?)

bin/commitments_deprecater.rb
```diff
@@ -0,0 +1,190 @@
...
+
+  # Do what you can with a line, store errors in @err
+  # return true if success / false otherwise
+  def try_deprecate(line)
+    clear_err
+    dep = DeprecationRecord.parse_line(line)
+    log "Find and deprecate a commitment based on the deprecation record: <#{dep}>."
+
+    # dep must match a cluster (on ocn)
+    cluster          = check_cluster(dep.ocn)
```
aelkiss, line 88: As I understand it essentially if it fails the condition the remaining operations are no-ops. It might be easier to follow the logic if we structured this essentially as a set of guard conditions. I also wonder if it would make sense to move this method (and the `check_*` methods it calls) to `DeprecationRecord`... more thoughts/examples later

bin/commitments_deprecater.rb
mwarin: These are files that we are likely to want to upload to a member box account though.

spec/commitments_deprecater_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+
+  before(:each) do
+    Cluster.collection.find.delete_many
+    dep.clear_err
+  end
+
+  # Some of the tests below come in pairs, with one pos(itive) and neg(ative) per method.
+  # Pos = where we expect to find something and we do
+  # Neg = where we expect to find nothing and we don't
+
```
jsteverman, line 52: Grouping by `describe "#method"` might keep this organized a bit better.

spec/commitments_deprecater_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+  it "check_local_id_matches" do
+    ret = dep.check_local_id_matches(
+      [Mockmitment.new(local_id_1), Mockmitment.new(local_id_2)],
+      local_id_1
+    )
+    expect(ret.size).to eq 1
+    expect(ret.first.local_id).to eq local_id_1
+  end
+
+  it "check_multiple_matches" do
```
jsteverman, line 133: e.g.
`describe "#check_multiple_matches" do`
`   it "finds duplicate local_ids" do` # or whatever this does
`     expect...`
`   end`
`end`

bin/commitments_deprecater.rb
aelkiss: Got it -- in that case I think the approach here makes sense; that might be worth explicitly noting.

bin/commitments_deprecater.rb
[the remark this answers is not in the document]
mwarin: Oh, English. You are correct, of course.
This breaks the symmetry with `HoldingsDeleter`. :cry: whyyyy

bin/commitments_deprecater.rb
mwarin: Would I expect that a piece of code which gets file paths from a human occasionally gets something that isn't a file path? Absolutely, I would. Especially if that human is giving paths that aren't even on their local machine via Tanka or whatever.
Is a runtime error preferred? I could go either way.

bin/commitments_deprecater.rb
aelkiss: Looking through this a bit more, from a _logic_ perspective we could simplify this quite a bit to get straight to the non-deprecated commitments for this organization with matching local IDs, but we want to have error logging that helps debug what the issue is. I might be tempted still to separate out the error reporting logic from the whittling-down logic, at least for the steps that are fast.

To be honest, I've always struggled with error reporting like this -- ingest suffers a lot of from this as well; the goals of trying to achieve clear logic, have functions with a single responsibility, and at the same time avoid repeating work are often in tension. I think it would be worth spending some time pairing on this and seeing if we can refactor it some, especially if we can do so without adding a bunch of extra complexity and abstraction around validations or some such.

spec/commitments_deprecater_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+  let(:empty_cluster) { create(:cluster) }
+
+  def make_commitment(ocn, org, local_id)
+    # Make and cluster a commitment
+    com = build(:commitment, ocn: ocn, organization: org, local_id: local_id)
+    cluster = Clustering::ClusterCommitment.new(com).cluster.tap(&:save)
+    cluster.commitments.first
+  end
+
+  def make_holding(ocn, org, local_id)
```
aelkiss, line 35: This returns a cluster, not a holding, but also it doesn't seem to be used, so perhaps we can remove it?

aelkiss (commented):
In general I think this is on the right track; the tests and in particular the integration tests look good and should serve as a good base for refactoring. I do think it's worthwhile thinking about moving the `check_*` methods and considering some further refactoring there as well. I'd also suggest renaming the `log` method to make it clear this is a report that we will provide externally and to avoid confusion with what `Services.logger` is doing. I wonder if it would be worth rethinking the format of the log to make it more machine-usable (e.g. the output could be a report with the status of each input deprecation record)?

spec/commitments_deprecater_spec.rb
mwarin: yep, that was a copy-paste leftover, is gone now

bin/commitments_deprecater.rb
mwarin: I've tried to reword it so it's more clear that it's a report and we are appending messages to a report, rather than logging to a log.

bin/commitments_deprecater.rb
mwarin: I am strongly pro-headers for tabular data.

bin/commitments_deprecater.rb
mwarin: Right. You'll notice that the `Cluster.where()` query just looks for an OCN. A lot could have been done here, but I wanted to be able to log along every step.

bin/commitments_deprecater.rb
mwarin: addressed

spec/shared_print/deprecation_record_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+        described_class.new(
+          organization: org,
+          ocn: ocn,
+          local_id: loc,
+          status: sta
+        )
+      end.not_to raise_error
+    end
+
+    it "neg, status" do
```
aelkiss, line 42: `it "raises ArgumentError when given an unknown status"` (should try to avoid "good" or "bad" value judgments when possible; they're just inputs the program doesn't handle or understand)

aelkiss (changes requested):
I'm happy with the changes in `DeprecationRecord`. I would like to see the tests fully described and to write out `MockCommitment` rather than using `Mockmittment`, both to avoid the amount of work necessary in the future to understand the tests.

spec/shared_print/deprecator_spec.rb
```diff
@@ -0,0 +1,119 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "shared_print/deprecator"
+
+class Mockmitment
```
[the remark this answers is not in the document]
mwarin: yeah that wasn't even used anymore so it's gone now

jsteverman (approved):

aelkiss (approved):
Looks good now, merge when ready.

PR #165, opened 2022-02-08
HT-3095:bugfix rclone move system command

jsteverman:


aelkiss (approved):
Looks good. In the past I would have wanted to see this using `Dir.mktmpdir` instead of hardcoded temporary directory names; that still might be good to do in that it ensures better test isolation, but it should be OK as is since you are making sure to remove the temporary directory before & after each test.

PR #166, opened 2022-02-10
add less and entr to Dockerfile

mwarin:
I've been using `entr` for developing, Sandi Metz style, but have had to keep it in a local modified Dockerfile.
Also I find it super frustrating to not have `less` available.
Can we just add them to the Dockerfile?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ARG UID=1000
 ARG GID=1000
 
 RUN apt-get update -yqq && apt-get install -yqq --no-install-recommends \
-  nodejs netcat
+  nodejs netcat less entr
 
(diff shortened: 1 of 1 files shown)


aelkiss:
Let's discuss this a bit -- I might have some other options besides adding it to the `Dockerfile` that work for my local development workflow. If those don't seem like options that would work for you I don't object to this, but then we should use a separate `Dockerfile.prod` for production/deployment use.

mwarin:
Don't we already have a `Dockerfile.prod`?

aelkiss:
So we do. I do at least want to show a bit of my local development workflow, and then you can decide if you want to merge this or not. I should have a little time tomorrow to at least give some examples.

PR #168, opened 2022-02-11
Test SessionKeepAlive and ways to prevent CursorNotFound

jsteverman:
I came up with a way to test SessionKeepAlive. It is not effective in preventing CursorNotFound, at least when it is caused by MongoDB waiting on the client to make the next batch request.  Setting `Cluster.max_time_ms` was also ineffective. One or both may be useful in preventing a timeout caused by a slow database query.

Reducing batch size was effective, at least in the contrived test.

All of these tests are `xit` because they are designed to cause a database timeout (10 minutes?), but are probably good documentation for a problem we are bound to run into again.

### DIFF
--- a/spec/utils/session_keep_alive_spec.rb
+++ b/spec/utils/session_keep_alive_spec.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "spec_helper"
 require "utils/session_keep_alive"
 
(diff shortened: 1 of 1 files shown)


mwarin:
BTW, tests tagged with `:slow` will be ignored by spec_helper.

> it 'is a slow test', :slow = > true do
> ...
> end

PR #169, opened 2022-02-11
ETAS Overlap: Limit batch size to avoid CursorNotFound

jsteverman:
I put `50` in the settings mostly on a whim.


aelkiss (approved):
Worth giving it a try.

PR #172, opened 2022-02-15
Deprecation fix mw 20220215

mwarin:
When the deprecation code met with reality, a few bugs revealed themselves.
* Report::EligibleCommitments did not output correct format.
* Report::EligibleCommitments was too restrictive, throwing away whole clusters because they had commitments.
* SharedPrint::DeprecationRecord had no quotes around the rejected status, leading to headscratching when the status was "".
* SharedPrint::Deprecator could find itself appending to @err before @err was defined as Array.
* Spruced up tests.


aelkiss (approved):
With a quick look it looks like the tests cover the issues you mention, so I'm happy with this.

PR #173, opened 2022-02-16
HT-1944

mwarin:
Allows for a combination of ocn and organization search terms, or all.
For all matching clusters, return holdings.
"Matching clusters" have at baseline holdings, htitems but no (active) commitments, and then filtered further based on ocn and/or organization. 

Thrown in is also a renaming of "eligible_commitments" to "commitment_replacements", as I kept getting annoyed at the inaccuracy of the name.

### DIFF
--- a/bin/reports/compile_uncommitted_holdings.rb
+++ b/bin/reports/compile_uncommitted_holdings.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require "reports/uncommitted_holdings"
+require "optparse"
+require "services"
+
+Services.mongo!
+
+def main(args)
+  criteria = parse_opts(args)
+  report = Reports::UncommittedHoldings.new(**criteria)
+  puts report.header.join("\t")
+  report.run do |record|
+    puts record.to_s
+  end
+end
+
+def parse_opts(args)
+  criteria = {}
+  help_str = "Restrict search to clusters matching"
+  OptionParser.new do |opts|
+    opts.on("--ocn [INT]", Array, "#{help_str} OCN(s)") do |ocns|
+      ocns.map!(&:to_i)
+    end
+    opts.on("--organization [STR]", Array, "#{help_str} org(s)")
+    opts.on("--all", "Search across all clusters.")
+    opts.on("--verbose")
+    opts.on("--noop")
+    opts.on("--help", "You're looking at it.") do |help|
+      puts opts
+      exit
+    end
+  end.parse!(args, into: criteria)
+
+  criteria
+end
+
+main(ARGV) if __FILE__ == $PROGRAM_NAME
--- a/lib/reports/uncommitted_holdings.rb
+++ b/lib/reports/uncommitted_holdings.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+require "services"
+require "reports/uncommitted_holdings_record"
+require "overlap/holding_commitment"
+
+module Reports
+  # Get holdings from spm clusters where there are items & holdings but no active commitments.
+  # Allow to optionally filter by OCN(s) and/or organization(s).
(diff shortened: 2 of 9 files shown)


lib/reports/uncommitted_holdings.rb
```diff
@@ -0,0 +1,53 @@
...
+  class UncommittedHoldings
+    attr_reader :criteria
+    def initialize(criteria = {})
+      @criteria = criteria
+      @verbose = criteria.delete(:verbose)
+      @noop = criteria.delete(:noop)
+      raise ArgumentError.new("No criteria given") if @criteria.empty?
+    end
+
+    def run
```
jsteverman, line 22: Maybe extract a `def query` here.

lib/reports/uncommitted_holdings.rb
```diff
@@ -0,0 +1,53 @@
...
+
+module Reports
+  # Get holdings from spm clusters where there are items & holdings but no active commitments.
+  # Allow to optionally filter by OCN(s) and/or organization(s).
+  # "I want to know which items in HTDL lack commitments,
+  #  so I can seek new titles to secure commitments on"
+  # Invoke via bin/reports/compile_uncommitted_holdings.rb
+  class UncommittedHoldings
+    attr_reader :criteria
+    def initialize(criteria = {})
```
jsteverman, line 15: Some named parameters here instead of a generic `criteria` would make this easier to follow.

bin/reports/compile_uncommitted_holdings.rb
```diff
@@ -0,0 +1,38 @@
...
+def main(args)
+  criteria = parse_opts(args)
+  report = Reports::UncommittedHoldings.new(criteria)
+  puts report.header.join("\t")
+  report.run do |record|
+    puts record.to_s
+  end
+end
+
+def parse_opts(args)
```
jsteverman, line 18: https://github.com/hathitrust/holdings-backend/tree/HT-3273-consolidate-bin-scripts-with-cli
can take care of most of this so don't sweat the details in here.

spec/reports/uncommitted_holdings_spec.rb
```diff
@@ -0,0 +1,146 @@
...
+  let(:com2) { build(:commitment, ocn: ocn2, local_id: loc2, organization: org2) }
+
+  let(:mpm) { build(:ht_item, :mpm, ocns: [ocn1]) }
+  let(:ser) { build(:ht_item, :ser, ocns: [ocn1]) }
+
+  # Clusters and saves each element in an array of clusterables
+  def cluster_tap_save(clusterables)
+    clusterables.each do |clusterable|
+      case clusterable.class
+      when Clusterable::Holding
```
aelkiss, line 32: I think this approach might be helpful to apply elsewhere in the tests. (It also makes it tempting to apply some kind of Rails-style metaprogramming to the `Clusterable` class name to get the corresponding `Clustering` class, which I think is best avoided even when straightforward...)

It's also worth noting that we don't really ever need this correspondence in the code itself (it really only shows up in the `Loader` classes, which know about the corresponding `Clusterable` and `Clustering` classes), but it shows up over and over in setup for the specs.

lib/reports/uncommitted_holdings.rb
aelkiss: Agreed; I went in thinking this was going to be an arbitrary set of mongo where clauses or something and got confused later on, but I agree we should have a set of named parameters here instead, since there is a limited set we actually handle.

aelkiss (commented):
See comments inline.

spec/reports/uncommitted_holdings_spec.rb
mwarin: Moved.

jsteverman (approved):

PR #175, opened 2022-02-18
update shared print commitments

mwarin:
New code for updating shared print commitments:
Give a file to `SharedPrint::Updater` and it'll go through each line, turn line into a `SharedPrint::UpdateRecord`, use `SharedPrint::Finder` to find a matching `Clusterable::Commitment` and update it if conditions are right.

### DIFF
--- a/lib/shared_print/finder.rb
+++ b/lib/shared_print/finder.rb
@@ -1,0 +1,79 @@
+# frozen_string_literal: true
+
+require "cluster"
+require "services"
+
+Services.mongo!
+
+module SharedPrint
+  # Pass criteria to new(), and
+  # * call clusters()    to yield the matching clusters,
+  # * call commitments() to yield the matching commitments.
+  class Finder
+    attr_reader :organization, :ocn, :local_id, :deprecated, :query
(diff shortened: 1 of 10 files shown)


mwarin:
There is no bin script for this, as I assumed it would at best become immediately outdated given `bin/phctl.rb`.

lib/shared_print/finder.rb
```diff
@@ -0,0 +1,86 @@
...
+    # Yield matching clusters.
+    def clusters
+      Cluster.where(@query).no_timeout.each do |cluster|
+        yield cluster
+      end
+    end
+
+    # Yield matching commitments in matching clusters.
+    def commitments
+      clusters do |cluster|
```
jsteverman, line 33: `return enum_for(:commitments) unless block_given?`

Then you can just do:
`finder.commitments.to_a`
obviating the need for `def gather`

Same with `def clusters`.

I think.

jsteverman (approved):

lib/shared_print/finder.rb
mwarin: Yeah, nice.

PR #177, opened 2022-02-22
removed automatic reporting

mwarin:
Automatic reporting is not desired in bin/add_print_holdings.rb, at least not the way it is currently being used.


aelkiss (approved):
Makes sense; we can consider some other way of orchestrating this at holdings load time when we return to fully automating holdings loads.

PR #178, opened 2022-02-23
Ht 3355 - organization status bugs

mwarin:
Turning `HTOrganization.status` from `1/0` to `true/false` brought a whole lot of test cases crashing down. 
This is that fix plus the resulting test fixes.


lib/reports/cost_report.rb
```diff
@@ -60,7 +60,11 @@ def pd_cost
     end
 
     def pd_cost_for_member(member)
-      (pd_cost / total_weight) * active_members[member.to_s].weight
```
jsteverman, line 63: "hathitrust" already has a weight of 0.0 in `002._data.sql` ?

jsteverman (changes requested):
Looks good other than my question about "hathitrust" weight.

lib/reports/cost_report.rb
mwarin: Ah good catch, this was something I added while everything was broken and didn't remove.
Removing.

jsteverman (approved):

PR #179, opened 2022-02-25
Ht 3273 consolidate bin scripts with cli

jsteverman:


mwarin (approved):
I could do with a quick "invoke me like this" comment in phct.rb, but see nothing else worth commenting on.

bin/phctl.rb
```diff
@@ -0,0 +1,112 @@
...
+      require_relative "concordance_validation/validate_and_delta.rb"
+      main
+    end 
+  end
+
+
+  class Report < Thor
+    desc "costreport ORG", "Run a cost report"
+    def costreport(org=nil)
+      require_relative "reports/compile_cost_reports.rb"
```
aelkiss, line 53: We will want to get these `main` functions out of the global namespace so we don't need to do the `require_relative` in the block; otherwise, this is looking pretty good.

jsteverman:
Now with support for the SP commitment operations. Testing the cli operations is still a bit of a question mark.

mwarin (approved):

aelkiss (approved):
Looking good to me at this point.

PR #180, opened 2022-03-03
Ht 3351 replace commitments

mwarin:
draft


lib/shared_print/replace_record.rb
```diff
@@ -0,0 +1,80 @@
...
+    def apply
+      # Sadly (?) this makes it so that all replaced commitments are deprecated with
+      # the same status "E". Even if previously deprecated with another status.
+      # Status "R" for Replaced was rejected because it would be meaningless.
+      matches.first.deprecate(status: "E", replacement: @replacement)
+      matches.first.save
+    end
+
+    # Original implementation that creates duplicate records with the same _ids in Mongo.
+    def apply_broken
```
jsteverman, line 45: There a reason for this, or just forgot to rm?

jsteverman (approved):
See `apply_broken`, otherwise good to go.

lib/shared_print/replace_record.rb
mwarin: This was intentionally left as a nod to @aelkiss  saying "Worth having a failing test for this case." on https://tools.lib.umich.edu/jira/browse/HT-3351, but I forgot to actually put the test in.

mwarin: Test added.

PR #181, opened 2022-03-08
update keio ht_items

mwarin:
A one-off script for updating ht_items that have `collection_code: "KEIO", billing_entity: "hathitrust"` to `billing_entity: "hathitrust"`.
Can be removed once used, assuming we never need it again.

### DIFF
--- a/lib/keio_updater.rb
+++ b/lib/keio_updater.rb
@@ -1,0 +1,41 @@
+require "cluster"
+require "services"
+
+Services.mongo!
+
+class KeioUpdater
+  def initialize(limit = nil)
+    @limit = limit
+  end
+
+  def limit_query(query)
+    if @limit.nil?
+      Cluster.where(**query)
+    else
+      Cluster.where(**query).limit(@limit)
+    end.no_timeout
+  end
+
+  def run
+    query = {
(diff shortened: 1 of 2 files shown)


lib/keio_updater.rb
```diff
@@ -0,0 +1,28 @@
...
+    query = {
+      "ht_items.0": {"$exists": 1},
+      "ht_items.collection_code": "KEIO",
+      "ht_items.billing_entity": "hathitrust",
+    }
+
+    Cluster.where(**query).no_timeout.each do |cluster|
+      cluster.ht_items.each do |ht_item|
+        if ht_item.collection_code == "KEIO" && ht_item.billing_entity == "hathitrust"
+          ht_item.billing_entity = "keio"
```
jsteverman, line 17: `ht_item.set_billing_entity` would also work, but same diff.

jsteverman (approved):

lib/keio_updater.rb
mwarin: Nope, not without changing `set_billing_entity` to public, which I didn't feel was called for.

jsteverman (approved):

PR #186, opened 2022-04-08
add uuid when creating a new commitment object

mwarin:
this may not be the smartest way of doing this, but it seems to work?


jsteverman (changes requested):
A test would be helpful as well as a comment on `def uuid_validation` describing its intent. The following might be a cleaner approach
```
def initialize
    super
    uuid = SecureRandom.uuid if (uuid.nil? || uuid.empty?)
end
```

jsteverman (approved):

PR #193, opened 2022-05-09
Set a cursor batch size for overlap table update

jsteverman:


aelkiss (approved):
Based on dac187821430da6aae87fbe2fdb2b02d319d661b I think this should be fine. Is there anywhere else we'd want to apply that and then perhaps abstract something out? Some possibilities:

https://github.com/hathitrust/holdings-backend/blob/baee2a64a11ddd3d7206b2164dd2308e2b829699/lib/overlap/cluster_overlap.rb#L47

https://github.com/hathitrust/holdings-backend/blob/90f6c73731473e86336b9af72c93653ccb2f8672/lib/reports/cost_report.rb#L105

https://github.com/hathitrust/holdings-backend/blob/baee2a64a11ddd3d7206b2164dd2308e2b829699/lib/clustering/cluster_holding.rb

https://github.com/hathitrust/holdings-backend/blob/baee2a64a11ddd3d7206b2164dd2308e2b829699/lib/clustering/cluster_holding.rb

For example we could do

```ruby
class Cluster
  def self.big_where(**kwargs)
    self.batch_size(Settings.large_select_batch_size).where(**kwargs).no_timeout
  end
end
```

mwarin (approved):

jsteverman:
The ETAS overlap reports and the overlap table updates have one thing in common that the other large queries don't have.  I *think* their main weakness is less the time it takes to compute records for a given batch of clusters and more because they both write to external services that can have high latency. 

A more performant solution may be to write these reports/updates to the pod locally, then transfer to htprep/maria in bulk at the end. I might be thinking this only because I don't know what kind of file system is backing the pods; writing across network to Great Value brand spinning disk would be worse.

aelkiss:
For transient storage in pods it's just using (by default) RAM on the local node. We could also use SSD local to the worker nodes if we need somewhat more space (that's what mongo is using) or NFS to the Isilon if we need more space (which is still better than Great Value (TM)-style spinning disk)

PR #194, opened 2022-05-12
Use pushgateway in dev

aelkiss:
Some scripts in dev failed because the pushgateway was not started by
default or referenced by development config. This starts and enables it.


mwarin (approved):
Works!

PR #197, opened 2022-05-27
Write ETAS rpts local, cp .gz to persistent storage

jsteverman:
Adds a step to the generation of the ETAS overlap reports. It now writes to a local directory by default. Once completed, the gz files are copied to the /htprep and dropbox locations.

### DIFF
--- a/bin/phctl.rb
+++ b/bin/phctl.rb
@@ -108,7 +108,7 @@ module PHCTL
     def etas_overlap(org=nil)
       rpt = Reports::EtasOrganizationOverlapReport.new(org)
       rpt.run
-      rpt.move_reports_to_remote
+      rpt.move_reports
     end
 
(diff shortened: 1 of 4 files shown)


jsteverman:
I don't get it. 
`cronJob.emptyVolumeMount('tmp','/tmp',volumeMixin=volume.emptyDir.withMedium('Memory'))` 
seems like writing to /usr/src/app/tmp with extra steps.

aelkiss:
Discussed via slack; the `emptyVolume` mount is necessary to use RAM rather than the k8s worker node local disk.

jsteverman:
I added `local_report_path: "/tmp_reports"` to the holdings configmap. 

And added the tmp_reports volume in tanka:
https://github.com/hathitrust/ht_tanka/pull/35

aelkiss:
:+1: :ship:

PR #199, opened 2022-06-20
Refactor autoscrub 2022

mwarin:
Based on suggestions in https://hathitrust.atlassian.net/browse/DEV-296.

The "files changed" diff looks like a lot, but I broke down in here https://docs.google.com/spreadsheets/d/1-Q1DqLhDrszsTL2ww23S_9zHTrOm4glUa6c-pmCA994/edit?usp=sharing which kinds of changes were made (hard to capture otherwise). Very little functional change, almost all reorganization.


lib/scrub/ocn.rb
```diff
@@ -0,0 +1,116 @@
...
+    # Take a string, presumed to be one or more OCNs, and get the actual OCNs out as numbers.
+    def initialize(str)
+      @ok_ocns = []
+
+      if str.nil?
+        raise ArgumentError, "ocn is nil"
+      end
+
+      # Get current max oclc
+      @current_max_ocn = Scrub::MaxOcn.new.ocn
```
jsteverman, line 17: Could we have Scrub::MaxOcn.new raise the error instead? And/or put this in a `def current_max_ocn`

lib/scrub/ocn.rb
```diff
@@ -0,0 +1,116 @@
...
+      if @current_max_ocn.nil?
+        raise "failed to set current_max_ocn"
+      end
+
+      count_x(:rec_with_ocn)
+      candidates = str.split(OCN_SPLIT_DELIM).map(&:strip)
```
jsteverman, line 23: Could use an additional OCNCandidate class we could hide stick all the validation in, but this is pretty easy to understand as is.

lib/scrub/ocn.rb
mwarin: Yep, added `Scrub::MaxOcn.current_max_ocn` which does the raising.

lib/scrub/ocn.rb
mwarin: Yeah, that makes it neater. Done.

PR #203, opened 2022-08-01
sketching out how to really run autoscrub at scale

mwarin:
here be placeholders and broken code

### DIFF
--- a/lib/scrub/scrub_runner.rb
+++ b/lib/scrub/scrub_runner.rb
@@ -14,89 +14,93 @@ require "loader/holding_loader"
 require "scrub/autoscrub"
 require "fileutils"
 
-# Scrubs and loads any new member-uploaded files.
-class ScrubRunner
-  def initialize
-    @members = DataSources::HTOrganizations.organizations
+# Example:
+# runner = Scrub::ScrubRunner.new
+# runner.run_all_members
+# runner.run_some_members(["umich", ..., "harvard"])
+# runner.run_one_member("umich")
 
-    if Settings.scrubbed_files_dir.nil?
-      raise "Need Settings.scrubbed_files_dir to be set"
+module Scrub
+  # Scrubs and loads any new member-uploaded files.
+  class ScrubRunner
+    def initialize
+      @members = DataSources::HTOrganizations.organizations
+      if Settings.scrubbed_files_dir.nil?
+        raise "Need Settings.scrubbed_files_dir to be set"
+      end
+    end
+    # see lib/reports/etas_organization_overlap_report.rb for rclone stuff,
+    # should be broken out into its own class?
+
+    def run_all_members
+      puts "Run all members."
+      run_some_members(@members)
+    end
+
+    def run_some_members(members)
+      puts "Run members: #{members.join(", ")}."
+      members.each do |member|
+        run_one(member)
+      end
+    end
+
+    def run_one_member(member)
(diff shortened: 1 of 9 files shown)


lib/scrub/scrub_runner.rb
```diff
@@ -0,0 +1,111 @@
...
+    # This should ideally spin off another (set of?) worker thread(s).
+    def run_file(member, file)
+      puts "running file #{file} for member #{member}"
+      download_to_work_dir(member, new_file)
+      scrubber = Scrub::Autoscrub.new(new_file)
+      scrubber.run
+      # figure out the output files that was just generated
+      # todo: implement scrubber.reports and/or do it with Scrub::ScrubOutputStructure
+      upload_to_member(member, scrubber.reports.latest)
+      holding_loader = Loader::HoldingLoader.for(scrubber.output.latest)
```
aelkiss, line 65: should hand off to chunker at this point (as another async job); chunker would then queue one job for each chunk

aelkiss: Probably simplifies tracking if we load all chunks for a given member rather than trying to sort members together

lib/scrub/scrub_runner.rb
```diff
@@ -0,0 +1,111 @@
...
+      Loader::FileLoader.new(holding_loader).load(scrubber.output.latest)
+    end
+
+    # Check member-uploaded files for any not previously seen files
+    def check_new_files(member)
+      puts "check new files for member #{member}"
+      remote_files = [] # todo: something something rclone
```
aelkiss, line 72: Should implement a class for rclone that implements the functionality we need (upload, download, list files)

Can use something like https://github.com/hathitrust/holdings-backend/blob/main/spec/reports/etas_organization_overlap_report_spec.rb#L8-L19 to use a local path for test purposes rather than using dropbox

aelkiss: Is a file showing up in dropbox sufficient cause for starting a scrub & load job? @mwarin says yes

aelkiss: A problem is that members can make their own directories inside their own dropbox. We can't control what they name things. Anything that doesn't match should probably trigger a warning.

Information from file name:
- member id - if on dropbox, should be determinable from the folder, but doesn't hurt to add it (since we have to add a file naming scheme)
- update type (full or partial) which we aren't using yet
- mono/multi/serial which we may not use
- date string - Dropbox has proved unreliable with file timestamping, so we should ask for this
- whether it's compressed - can only work with uncompressed or .gz

aelkiss: What happens if somebody renames a file? More generally, how do we want to track where things are in the process?

lib/scrub/scrub_runner.rb
mwarin: `Scrub::Chunker` is essentially ready to plug in here.

lib/scrub/scrub_runner.rb
mwarin: `Utils::FileTransfer` in combination with `DataSources::DirectoryLocator` are pretty much doing this now.

mwarin:
Broke out the `Utils::FileTransfer` code as PR#206.

aelkiss:
Superceded by `oct-17-dev` branch

PR #204, opened 2022-08-02
phctl starts sidekiq jobs

aelkiss:
Posting as a draft PR to give some sense of the shape of how this will work.

In terms of Kubernetes resources:
- we'll have a Deployment that runs a set of Sidekiq workers that can run any of these jobs
- to enqueue jobs, the pattern will be to do something like:

```bash
kubectl exec -n holdings deployment/holdings-client -- bundle exec ruby bin/phctl.rb ...
```
We can adapt the wrapper scripts currently at https://github.com/hathitrust/ht_tanka/tree/main/environments/holdings/jobs to do this instead and then (I think it would be simplest) move those scripts to this repository.

I have a similar pattern for ingest (see https://github.com/hathitrust/ht_tanka/blob/main/environments/ingest/cmdline/main.jsonnet; the stuff to exec into it is here:

https://github.com/hathitrust/ingest_helpers/blob/main/perl_wrappers/enqueue (for example)
https://github.com/hathitrust/ingest_helpers/blob/main/perl_wrappers/k8s_helpers.sh
https://github.com/hathitrust/ingest_helpers/blob/main/k8s/kubeexec

For ingest, we even have `kubectl` installed on `macc-ht-ingest-000` so Jackson can prep material and queue it up for ingest. We could do something similar on sidecar for holdings, if desired.


aelkiss:
The one piece still missing here is that some reports output to stdout instead of a file. I think we can begin reviewing this though and wait to merge until we have that additional piece.

Gemfile
```diff
@@ -17,6 +17,8 @@ gem "rgl"
 gem "sequel"
 gem "thor"
 gem "zinzout"
+gem "puma"
```
aelkiss, line 20: Added so that we can run the sidekiq web front-end.

README.md
```diff
@@ -33,6 +33,16 @@ docker volume rm holdings-backend_gem_cache # to clear out gems
 bash bin/setup/setup_dev.sh
 ```
 
+## Generating and loading fake data
```
aelkiss, line 36: This is pretty bare-bones instructions -- do we want more?

bin/generate_fake_data.rb
```diff
@@ -0,0 +1,111 @@
...
+  if(File.exist?(outfile))
+    puts "#{outfile} already exists, not creating"
+    true
+  else
+    false
+  end
+end
+
+
+# For now, only single-part monographs
```
aelkiss, line 30: The general idea here is to generate enough data to be somewhat realistic; using a maximum OCN of the same magnitude as the number of items we're generating ensures that we have htitems and holdings that are actually in the same cluster. We could reduce the maximum number of OCNs to get more holdings/htitems in a cluster, if we wanted. Note that we're only currently generating spms -- I am mostly interested in testing data loading, and most of the logic around data loading doesn't care at all about format.

bin/generate_fake_data.rb
```diff
@@ -0,0 +1,111 @@
...
+
+def generate_holdings(count = 50000, inst = "umich", date = Date.parse("2020-01-01"))
+  # holdings
+  outfile = File.join(OUTDIR, "umich_fake_holdings.ndj")
```
aelkiss, line 34: Might want to parameterize the file output so that we can load holdings from multiple members.

It was much easier to build holdings in JSON format than to deal with the TSV format. Eventually we may want to be able to generate fake data in the format members are expecting to send us so that we can do end-to-end testing with autoscrub.

bin/generate_fake_data.rb
```diff
@@ -0,0 +1,111 @@
...
+        ocn: rand(MAX_OCN),
+        date_received: date)
+      out.puts(holding.as_document.except("_id").to_json)
+    end
+  end
+end
+
+def htitem_fields(max_ocn: MAX_OCN)
+  htitem = FactoryBot.build(:ht_item, ocns: [rand(max_ocn).to_s])
+  collection = Services.ht_collections[["PU", "MIU", "KEIO", "UCM"].sample]
```
aelkiss, line 50: Most of this data just gets thrown away. I don't know if it's worth using Faker here, or if it would be better to make it explicit that we don't actually use it.

bin/generate_fake_data.rb
```diff
@@ -0,0 +1,111 @@
...
+  outfile = File.join(OUTDIR, "htitems_fake.tsv")
+  return if warn_exists(outfile)
+  File.open(outfile, "w") do |out|
+    count.times do
+      out.puts(htitem_fields(max_ocn: MAX_OCN).join("\t"))
+    end
+  end
+end
+
+def generate_ocn_resolutions(max_ocn = MAX_OCN, percent_deprecated = 0.10)
```
aelkiss, line 92: We can increase `perecent_deprecated` if we want to model having a greater average # of OCNs per cluster. I don't know what the stats "for real" are.

bin/sidekiq_web.ru
```diff
@@ -0,0 +1,19 @@
...
+
+if !File.exist?(SESSION_KEY)
+  require "securerandom"
+  File.open(SESSION_KEY, "w") do |f|
+    f.write(SecureRandom.hex(32))
+  end
+end
+
+Sidekiq.configure_client do |config|
+  config.redis = {size: 1}
```
aelkiss, line 14: I believe this is ensuring that the sidekiq web interface uses only one connection to Redis, which seems fine.

lib/loader/file_loader.rb
```diff
@@ -39,7 +39,10 @@ def load(filename, filehandle: Zinzout.zin(filename), skip_header_match: nil)
       skip_matching_header(filehandle, skip_header_match).each.lazy
         .map { |line| log_and_parse(line) }
         .chunk_while { |item1, item2| item1.batch_with?(item2) }
-        .each { |batch| batch_loader.load(batch) }
+        .each do |batch|
+          batch_loader.load(batch)
+          Thread.pass
```
aelkiss, line 44: Sidekiq complains if your jobs run for too long without passing. I'm not convinced they need to pass after every batch, and the threading in general might be a source of the bad performance I was seeing in dev (i.e. it might be better just to start a bunch of processor containers, rather than a single one with many threads)

lib/loader/holding_loader.rb
```diff
@@ -27,12 +27,6 @@ def item_from_line(_line)
     def load(batch)
       Clustering::ClusterHolding.new(batch).cluster
     end
-
-    def final_line
```
aelkiss, line 31: AFAICT this was never getting called (there was no way for @update to be set). In any case it wouldn't have worked with the chunking, and regardless we need to have another method of indicating when all chunks are done and then doing the deletes as a separate job.

lib/reports/cost_report.rb
```diff
@@ -9,6 +9,8 @@ class CostReport
     attr_accessor :organization, :logger, :maxlines, :target_cost
 
     def initialize(org = nil, cost: Settings.target_cost, lines: 1_000_000, logger: Services.logger)
+      cost ||= Settings.target_cost
```
aelkiss, line 12: I'm not sure why I needed this. Maybe something is passing in `cost: nil` and this is to try to get around that? Seems like it would be worth fixing whatever was sending `cost: nil`.

lib/reports/etas_organization_overlap_report.rb
```diff
@@ -18,10 +18,10 @@ def initialize(organization = nil)
       @local_report_path = Settings.local_report_path || "local_reports"
       Dir.mkdir(@local_report_path) unless File.exist?(@local_report_path)
       # persistent storage
-      @persistent_report_path = Settings.etas_overlap_reports_path || "tmp_reports"
+      @persistent_report_path = Settings.etas_overlap_reports_path
```
aelkiss, line 21: Should use the settings mechanism rather than setting it in a surprising way here. Should probably update `@local_report_path` above as well.

lib/sidekiq_jobs.rb
```diff
@@ -0,0 +1,180 @@
...
+require "reports/compile_estimated_IC_costs"
+require "reports/compile_member_counts_report"
+require "shared_print/updater"
+require "shared_print/replacer"
+require "shared_print/deprecator"
+
+# Don't want to do this by default when we aren't running under sidekiq
+Services.register(:logger) { Sidekiq.logger }
+
+module Jobs
```
aelkiss, line 24: I don't love the parallel class hierarchy with slightly different names, BUT the advantage to this is that we don't have to really change our existing code or tie it to sidekiq in any way -- only phctl and this need to know about sidekiq.

spec/factories/ht_item.rb
```diff
@@ -6,7 +6,7 @@
 FactoryBot.define do
   factory :ht_item, class: "Clusterable::HtItem" do
     ocns { [rand(1_000_000)] }
-    item_id { rand(1_000_000).to_s }
+    item_id { "test." + rand(1_000_000).to_s }
```
aelkiss, line 9: htids should have a namespace

spec/fixtures/cluster_2503661.json
```diff
@@ -37,7 +37,7 @@
       "n_chron": "",
       "ocn": 2503661,
       "local_id": "000238264",
-      "organization": "amherst",
+      "organization": "upenn",
```
aelkiss, line 40: These needed to be organizations that were mocked in the fixture data

spec/reports/compile_estimated_ic_costs.rb
```diff
@@ -1,4 +0,0 @@
-# frozen_string_literal: true
```
aelkiss, line 1: This did nothing (since it wasn't named like `_spec.rb anyway) and now we aren't running the report from a separate bin script anyway

spec/reports/cost_report_spec.rb
```diff
@@ -17,7 +17,7 @@
   let(:holding2) { build(:holding, ocn: c.ocns.first, organization: "smu") }
 
   before(:each) do
-    Settings.target_cost = 10
+    @old_cost = Settings.target_cost
```
aelkiss, line 20: I don't recall specifically what was breaking here, but in general if we're monkeying with settings in tests we should ensure they get set back to their original values so that we don't end up with behavior that depends on the order we run the tests.

spec/reports/etas_organization_overlap_report_spec.rb
```diff
@@ -5,15 +5,11 @@
 require "reports/etas_organization_overlap_report"
 
 RSpec.describe Reports::EtasOrganizationOverlapReport do
-  let(:tmp_local) { "tmp_local_reports" }
-  let(:tmp_pers) { "tmp_pers_path" }
-  let(:tmp_rmt) { "tmp_remote_path" }
+  let(:tmp_local) { Settings.local_report_path }
```
aelkiss, line 8: again here we can rely on the settings environment mechanism rather than hardcoding it. (Ultimately it might end up being better to use `Dir.mktmpdir`, though.)

spec/shared_print/deprecator_spec.rb
```diff
@@ -25,15 +27,10 @@ def lookup_commitment(ocn)
 
   before(:each) do
     Cluster.collection.find.delete_many
+    FileUtils.rm(Dir.glob("/tmp/deprecation_report/*"))
```
aelkiss, line 30: Another case where just using `mktmpdir` might make things less brittle.

spec/sidekiq_jobs_spec.rb
```diff
@@ -0,0 +1,151 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "sidekiq_jobs"
+require "sidekiq/testing"
+
+RSpec.describe "Jobs" do
```
aelkiss, line 7: In general I tried to do the minimum here to ensure that the job was actually running and doing what was expected without being dependent on a lot of data. I'm not sure how much integration testing to do for the reporting here, since testing anything for real depends on enough data being loaded.

spec/spec_helper.rb
```diff
@@ -13,7 +13,11 @@
 require "pry"
 require "settings"
 require "services"
+require "sidekiq/testing"
+
 SimpleCov.start
+Sidekiq.strict_args!
+Sidekiq::Testing.inline!
```
aelkiss, line 20: By default,  run the jobs inline as if we just called `perform` directly.

spec/spec_helper.rb
```diff
@@ -81,6 +85,12 @@
       Services.register(:logger) { old_logger }
     end
   end
+
+  config.around(:each, type: "sidekiq_fake") do |example|
+    Sidekiq::Testing.fake! do
```
aelkiss, line 90: Tests tagged with `type: "sidekiq_fake"` that use `Sidekiq::Testing.fake!` here allow us to test that the correct job gets queued with the correct arguments (i.e. for testing phctl itself) without running the job.

aelkiss:
@mwarin  I added a bunch of comments to try to explain certain areas.

While there are a lot of files that have changed here, most of the meaningful work is here:

* lib/phctl.rb: https://github.com/hathitrust/holdings-backend/pull/204/files#diff-caee7444d5af165d983dabbb7c95eaf804ade33b67e27dbdbde8ddf6a6477d28
* lib/sidekiq_jobs.rb: https://github.com/hathitrust/holdings-backend/pull/204/files#diff-7ccec67dc6d4f5856389b57d31a0ed17e46bef643cbc6ac4abe699efb9ffd769
* spec/phctl_spec.rb: https://github.com/hathitrust/holdings-backend/pull/204/files#diff-70a2bb9e11143cc8a1117d3bb8bc3ce9e36a9c52c3a6cf2f5edc42fc549caa46
* spec/sidekiq_jobs_spec.rb: https://github.com/hathitrust/holdings-backend/pull/204/files#diff-0bd3b45786bbb64da4a937fb44bc4b76b0067ae82f67c9f264052b6c048df447

aelkiss:
Looks like holdings deletion isn't working (not working for me currently in dev either) so I'll see what's going on with that.

aelkiss:
@mwarin  Let's talk about the best way to review this -- I think there are a couple separable pieces here (in particular the fake data generation & loading is separate from from the sidekiq jobs & refactoring to support that)

docker-compose.yml
```diff
@@ -11,6 +11,30 @@ services:
...
+      - gem_cache:/gems
+    environment:
+      REDIS_URL: redis://redis/
+      MYSQL_CONNECTION_STRING: "mysql2://ht_repository:ht_repository@mariadb/ht_repository"
+      PUSHGATEWAY: http://pushgateway:9091
+    entrypoint: bundle exec ruby bin/phctl.rb
+
+  processor:
+    build: .
+    restart: always
```
mwarin, line 27: Should the phctl instance have restart:always as default? 
Seems potentially harmful.

lib/loader/holding_loader.rb
mwarin: Yeah I think this was ever only aspirational.

lib/sidekiq_jobs.rb
mwarin: I don't understand the nested modules here.

spec/reports/etas_organization_overlap_report_spec.rb
mwarin: I would like to see a cleanup PR where we implement a good temp dir policy for all tests that generate output files.

spec/spec_helper.rb
```diff
@@ -6,14 +6,27 @@
...
 require "webmock/rspec"
 require "fixtures/organizations"
 require "fixtures/collections"
 require "fixtures/large_clusters"
 require "pry"
 require "settings"
 require "services"
+require "sidekiq/testing"
+
+SimpleCov::Formatter::LcovFormatter.config do |c|
```
mwarin, line 19: Is this related to a particular issue or did you just want to have better coverage analysis?

docker-compose.yml
aelkiss: I think this is correct. Sidekiq should manage retrying individual jobs, and if the processor (f/k/a "worker") crashes over and over again Kubernetes will mark it as `CrashLoopBackOff`; `docker-compose` has similar functionality to prevent a restart loop.

spec/spec_helper.rb
aelkiss: Not related to any particular issue -- mainly just wanted to be able to see coverage reports in pull requests and to be able to see coverage over time.

Gemfile.lock
```diff
@@ -58,7 +60,11 @@ GEM
       byebug (~> 11.0)
       pry (~> 0.13.0)
     public_suffix (4.0.7)
+    puma (5.6.4)
+      nio4r (~> 2.0)
+    rack (2.2.4)
     rainbow (3.1.1)
+    redis (4.7.1)
```
mwarin, line 67: I guess it's time to learn about redis.

lib/sidekiq_jobs.rb
aelkiss: Question of how to add additional jobs -- worth documenting. Probably worth additional cleanup tickets to get the remaining jobs in a common shape (probably either a 'load' shape or a 'run' shape)

aelkiss: Worth documenting some of the intention here.

lib/loader/file_loader.rb
aelkiss: See DEV-405

lib/reports/etas_organization_overlap_report.rb
aelkiss: Will finish cleaning this up in DEV-406 / DEV-407

lib/shared_print/deprecator.rb
```diff
@@ -108,9 +92,24 @@ def try_deprecate(line)
...
       @err = []
     end
-  end
 
-  if __FILE__ == $PROGRAM_NAME
-    SharedPrint::Deprecator.new.run(ARGV)
+    private
+
+    # Append messages to a report file, and to STDERR if @verbose
+    def append_report(msg)
```
aelkiss, line 99: Could be made more consistent with the other reports.

spec/reports/etas_organization_overlap_report_spec.rb
aelkiss: DEV-406

mwarin (approved):

PR #205, opened 2022-08-10
fixes for when there is no matching cluster

mwarin:
was getting errors running spo-20, having to do with expecting `[]`s but getting `nil`s

### DIFF
--- a/lib/shared_print/deprecation_record.rb
+++ b/lib/shared_print/deprecation_record.rb
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
+require "cluster"
+require "shared_print/deprecation_error"
+
 module SharedPrint
   # Represents a line in a file of (shared print) deprecation requests,
   # and can try to find a Commitment that matches it.
   class DeprecationRecord
-    attr_reader :organization, :ocn, :local_id, :status, :err
+    attr_reader :organization, :ocn, :local_id, :status
 
     def initialize(organization: nil, ocn: nil, local_id: nil, status: nil)
       @ocn = ocn.to_i
@@ -12,7 +15,6 @@ module SharedPrint
       @local_id = local_id
       @status = status
       @allowed_status = ["C", "D", "E", "L", "M"]
-      @err = []
       validate!
     end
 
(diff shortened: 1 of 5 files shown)


mwarin:
@aelkiss I need this in production in order to complete https://hathitrust.atlassian.net/browse/SPO-20

aelkiss (changes requested):
I'm finding the logic here pretty difficult to follow, and I'm not following the choices around where to use `&.` In particular I think it's difficult to follow because sometimes we're using the `org_commitments` method and sometimes using the `@org_commitments` instance variable, and mutating `@org_commitments` in multiple places. I think we can probably clean that up by using temporary variables in https://github.com/hathitrust/holdings-backend/blob/f2a6d21c5eb53cfc228d2f2962bd96c0d70fdacd/lib/shared_print/deprecation_record.rb#L113 rather than mutating `@org_commitments`; let's find some time to pair on working through this.

lib/shared_print/deprecation_record.rb
```diff
@@ -77,8 +79,8 @@ def org_commitments
     # If commitments contain deprecated ones, add to @err and remove them.
     # Return only non-deprecated
     def reject_deprecated
-      already_deprecated = org_commitments.select(&:deprecated?)
-      @org_commitments.reject!(&:deprecated?)
+      already_deprecated = org_commitments&.select(&:deprecated?)
```
aelkiss, line 82: It looks like `org_commitments` could return nil if there was a previous error; I think maybe some of my confusion here is that it's trying to follow through parts of the process if there was an error? Is there a reason to do that rather than baling out after the first error?

lib/shared_print/deprecation_record.rb
```diff
@@ -77,8 +79,8 @@ def org_commitments
...
+      @org_commitments&.reject!(&:deprecated?)
       if @err.empty? && org_commitments.empty?
```
aelkiss, line 84: so I guess we don't need `&.` here because if `@err.empty?` is true then `org_commitments` will be an array?

lib/shared_print/deprecation_record.rb
mwarin: No more @err, just raising at the first sign of trouble.

lib/shared_print/deprecation_record.rb
mwarin: No more &.

mwarin:
Changed to using exceptions and that makes it much cleaner, with more uniformity between the whittle-down methods.

aelkiss:
I agree this is much clearer, and actually suggests to me something that might be even clearer about what's happening in find_commitment by moving the validation separate from the logic of getting the data -- what would you think of something like:

```ruby
  # Raise if ocn does not lead to a cluster.
    def cluster
      @cluster ||= Cluster.where(ocns: @ocn).first
    end

    # If cluster does not have commitments by dep.organization, raise.
    # Return the commitments that _do_ match dep.organization.
    def org_commitments
      @org_commitments = cluster.commitments.select do |c|
        c.organization == @organization
      end
    end

    # If there are only deprecated commitments, raise.
    # Otherwise return the undeprecated commitments.
    def undeprecated_commitments
      @undeprecated_commitments ||= org_commitments.select do |c|
        c.deprecated? == false
      end
    end

    # Raise if no commitment matches local_id.
    # Return the one(s) that do(es) match.
    def local_id_matches
      @local_id_matches = undeprecated_commitments.select do |c|
        c.local_id == @local_id
      end

      if @local_id_matches.empty?
        raise SharedPrint::DeprecationError, "No commitment with local_id:#{@local_id} found"
      end

      @local_id_matches
    end

    # Multiple local_id matches is an error maybe?
    def single_match
      if local_id_matches.size > 1
        err = ["Multiple matches found:"]
        local_id_matches.each do |match|
          err << match.inspect
        end
        raise SharedPrint::DeprecationError, err.join(" / ")
      end

      @single_match = local_id_matches.first
    end

    # Start at cluster level, given ocn, and whittle down.
    # Raise error if at any point we have an unexpected number of things.
    def find_commitment
      raise SharedPrint::DeprecationError, "No cluster found for OCN:#{ocn}" if cluster.nil?
      raise SharedPrint::DeprecationError, "No commitments found in cluster" if cluster.commitments.empty?
      raise SharedPrint::DeprecationError, "No commitments by organization:#{organization} in cluster." if org_commitments.empty?
      raise SharedPrint::DeprecationError, "Only deprecated commitments found." if undeprecated_commitments.empty?
      raise SharedPrint::DeprecationError, "No commitment with local_id:#{@local_id} found" if local_id_matches.empty?
      single_match
    rescue SharedPrint::DeprecationError => e
      raise SharedPrint::DeprecationError,
        "Could not match deprecation record to exactly one commitment. #{e.message}"
    end
```

`single_match` has a little more complicated validation logic, so not sure what to do about that one.

aelkiss (commented):
See comment above (https://github.com/hathitrust/holdings-backend/pull/205#issuecomment-1216970056) -- I would be OK merging as-is, but see what you think of the proposed change.

aelkiss (approved):

PR #206, opened 2022-08-16
file transfer class for DEV-358

mwarin:
For https://hathitrust.atlassian.net/browse/DEV-358, a child task of DEV-299.
A file transfer class that simplifies uploading/downloading to/from local storage and dropbox.
Essentially just a wrapper for `rclone` system calls.

### DIFF
--- a/lib/utils/file_transfer.rb
+++ b/lib/utils/file_transfer.rb
@@ -31,7 +31,7 @@ module Utils
       make_call("#{call_prefix} ls \"#{dir}\"")
     end
 
-    def mkdir(dir)
+    def mkdir_p(dir)
       make_call("#{call_prefix} mkdir \"#{dir}\"")
     end
 
--- a/lib/utils/file_transfer.rb
+++ b/lib/utils/file_transfer.rb
@@ -36,7 +36,7 @@ module Utils
     end
 
     # Return parsed JSON ls output
-    def ls_remote_dir(remote_dir)
+    def lsjson(remote_dir)
       # Never parse ls output, let rclone do it for us.
       call = "#{call_prefix} lsjson \"#{remote_dir}\""
       puts "call #{call}"
--- a/lib/utils/file_transfer.rb
+++ b/lib/utils/file_transfer.rb
@@ -43,8 +43,8 @@ module Utils
       response = `#{call}`
       puts "response #{response}"
       JSON.parse(response)
-    rescue JSON::ParserError => _err
-      raise Utils::FileTransferError, "could not ls #{remote_dir}"
+    rescue JSON::ParserError => err
+      raise Utils::FileTransferError, "Could not ls #{remote_dir}... #{err}"
     end
 
     private
--- a/lib/utils/file_transfer.rb
+++ b/lib/utils/file_transfer.rb
@@ -1,0 +1,76 @@
+require "services"
+require "utils/file_transfer_error"
+
+module Utils
+  # Use rclone system calls to transfer files between
+  # local and remote storage (and vice versa).
+  class FileTransfer
+    def initialize
+      # Check that config is set...
+      if Settings.rclone_config_path.nil?
+        raise "Settings.rclone_config_path missing from Settings"
+      end
+      # ... and points to an actual file.
+      unless File.exist?(Settings.rclone_config_path)
+        raise "Settings.rclone_config_path points to " \
+              "#{Settings.rclone_config_path}, which does not exist"
+      end
+    end
+
+    # Really just an alias for transfer
+    def upload(local_file, remote_dir)
+      transfer(local_file, remote_dir)
+    end
+
+    # Really just an alias for transfer
+    def download(remote_file, local_dir)
+      transfer(remote_file, local_dir)
+    end
+
+    def exists?(dir)
+      make_call("#{call_prefix} ls \"#{dir}\"")
+    end
+
+    def mkdir(dir)
+      make_call("#{call_prefix} mkdir \"#{dir}\"")
+    end
+
+    # Return parsed JSON ls output
+    def ls_remote_dir(remote_dir)
+      # Never parse ls output, let rclone do it for us.
+      call = "#{call_prefix} lsjson \"#{remote_dir}\""
+      puts "call #{call}"
+      response = `#{call}`
+      puts "response #{response}"
+      JSON.parse(response)
+    rescue JSON::ParserError => _err
+      raise Utils::FileTransferError, "could not ls #{remote_dir}"
+    end
+
+    private
+
+    def make_call(sys_call)
+      puts sys_call
+      # returns true/false based on exit code of the executed system call
+      system sys_call
+    end
+
+    # Any call will start with:
+    def call_prefix
+      "rclone --config #{Settings.rclone_config_path}"
+    end
+
+    def transfer(file, dir)
+      unless exists?(file)
+        raise Utils::FileTransferError, "file #{file} does not exist"
+      end
+
+      unless exists?(dir)
+        raise Utils::FileTransferError, "dir #{dir} does not exist"
+      end
+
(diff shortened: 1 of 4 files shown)


lib/utils/file_transfer.rb
```diff
@@ -0,0 +1,76 @@
...
+    # Return parsed JSON ls output
+    def ls_remote_dir(remote_dir)
+      # Never parse ls output, let rclone do it for us.
+      call = "#{call_prefix} lsjson \"#{remote_dir}\""
+      puts "call #{call}"
+      response = `#{call}`
+      puts "response #{response}"
+      JSON.parse(response)
+    rescue JSON::ParserError => _err
+      raise Utils::FileTransferError, "could not ls #{remote_dir}"
```
moseshll, line 47: The custom exception class loses the context of the JSON error. I think I might pass `_err` on to the `Utils::FileTransferError` initializer as a custom field and just tuck it away in an ivar for the unlikely situation there's a need to debug unexpected output from `lsjson`. Or more simply, append `_err` in some form a la `"could not ls #{remote_dir} (#{_err})"` (should remove the underscore on `_err` of course yadda yadda)

spec/utils/file_transfer_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+  let(:remote_file_name) { "test_file_b.txt" }
+  let(:local_file_path) { "#{local_dir}/#{local_file_name}" }
+  let(:remote_file_path) { "#{remote_dir}/#{remote_file_name}" }
+  let(:conf) { "config/rclone.conf" }
+  let(:missing_dir) { "/dev/null/foo" }
+  let(:missing_file) { "/dev/null/foo.txt" }
+
+  # Clean up temporary files and directories
+  before(:each) do
+    Settings.rclone_config_path = conf
```
aelkiss, line 32: I think we can probably avoid setting this here and just do it in `settings/test.yml`. If we do change it here, we should change it back to the previous value, e.g:

```ruby
before(:each) do
  @old_rclone_config_path = Settings.rclone_config_path
  Settings.rclone_config_path = "/mashed/potatoes"
end

after(:each) do
  Settings.rclone_config_path = @old_rclone_config_path
end
```

spec/utils/file_transfer_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+    end
+    it "throws an error if dir does not exist" do
+      expect(ft.exists?(remote_dir)).to be true
+      expect { ft.ls_remote_dir(remote_dir) }.not_to raise_error
+      expect(ft.exists?(missing_dir)).to be false
+      expect { ft.ls_remote_dir(missing_dir) }.to raise_error Utils::FileTransferError
+    end
+  end
+
+  context "#mkdir" do
```
aelkiss, line 77: Consider calling this `mkdir_p` or `mkpath` for parity with `FileUtils`?

lib/utils/file_transfer.rb
```diff
@@ -0,0 +1,76 @@
...
+    def exists?(dir)
+      make_call("#{call_prefix} ls \"#{dir}\"")
+    end
+
+    def mkdir(dir)
+      make_call("#{call_prefix} mkdir \"#{dir}\"")
+    end
+
+    # Return parsed JSON ls output
+    def ls_remote_dir(remote_dir)
```
aelkiss, line 39: `exists?` and `mkdir` are both (I assume) primarily for remote directories; consider calling this just `ls` rather than `ls_remote_dir` for parallelism?

aelkiss (changes requested):
Some comments inline to consider; one thing I definitely do want to do is to make sure we're describing our tests well, though -- in particular:

- use `describe` when we're talking about a particular piece of functionality; `context` when there's differing setup/teardown -- primarily because that's the rspec convention, and it (should) make reading the tests (and potentially the output) easier

- use descriptive language in the "it" description so that we can more easily understand later what the test is testing / understand the output when the test fails

I'll add some additional comments inline by way of example.

spec/utils/file_transfer_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+    FileUtils.touch remote_file_path
+    FileUtils.touch conf # if it does not exist FileTransfer complains
+  end
+
+  after(:each) do
+    FileUtils.rm_rf local_dir
+    FileUtils.rm_rf remote_dir
+  end
+
+  context "initialize" do
```
aelkiss, line 47: `describe #initialize`

spec/utils/file_transfer_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+      expect { described_class.new }.to raise_error RuntimeError
+    end
+    it "requires conf file to exist" do
+      expect { described_class.new }.not_to raise_error
+      FileUtils.rm conf
+      expect { described_class.new }.to raise_error RuntimeError
+    end
+  end
+
+  context "listing remote files" do
```
aelkiss, line 60: `describe #ls_remote_files`

spec/utils/file_transfer_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+    it "ls_remote_dir" do
```
aelkiss, line 61: `it "provides the expected json"`

spec/utils/file_transfer_spec.rb
aelkiss: also, `describe "#mkdir"`

spec/utils/file_transfer_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+    end
+    it "makes the whole path even if missing, like mkdir -p" do
+      new_dir = "#{local_dir}/nope1/nope2/nope3"
+      expect(ft.exists?(new_dir)).to be false
+      ft.mkdir(new_dir)
+      expect(ft.exists?(new_dir)).to be true
+    end
+  end
+
+  context "transferring files" do
```
aelkiss, line 92: I'd break this into separate blocks for upload & download, e.g:

```ruby
describe "#upload" do
  it "puts a file in the expected location" do
      expect(ft.ls_remote_dir(remote_dir).count { |h| h["Name"] == local_file_name }).to eq 0
      ft.upload(local_file_path, remote_dir)
      expect(ft.ls_remote_dir(remote_dir).count { |h| h["Name"] == local_file_name }).to eq 1  
  end

  it "raises if the file does not exist" do
    expect { ft.upload(missing_file, remote_dir) }.to raise_error Utils::FileTransferError
  end

  it "raises if dir does not exist" do
    expect { ft.upload(local_file_path, missing_dir) }.to raise_error Utils::FileTransferError
  end
end
```

and likewise for download

moseshll (approved):
Looks like Aaron's got a bevy of rspec suggestions, I just have the one suggestion for exception handling.

spec/utils/file_transfer_spec.rb
mwarin: yep and yep

lib/utils/file_transfer.rb
mwarin: renamed it lsjson, which is what it does

spec/utils/file_transfer_spec.rb
mwarin: yup

lib/utils/file_transfer.rb
mwarin: Yeah that's better.

spec/utils/file_transfer_spec.rb
mwarin: I cannot get this to work and it's really bugging me, but I have to quit for the day.

mwarin: yup

aelkiss (approved):
Looks good to go!

PR #207, opened 2022-08-16
code for chunking in ruby (ish)

mwarin:
A ruby wrapper around system calls to sort & split, for the specific purpose of splitting scrubbed files into several sorted files so they can be loaded in parallel. 

For https://hathitrust.atlassian.net/browse/DEV-359, a child task of https://hathitrust.atlassian.net/browse/DEV-299.

### DIFF
--- a/spec/scrub/chunker_spec.rb
+++ b/spec/scrub/chunker_spec.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "scrub/chunker"
+
+Settings.scrub_chunk_work_dir = "/tmp/scrub_chunks"
+RSpec.describe Scrub::Chunker do
+  let(:pre_chunked_file) { "spec/fixtures/pre_chunk.tsv" }
+  let(:input_line_count) { 16 }
+  let(:chunk_count) { 4 }
+  let(:expected_chunk_line_count) { 4 }
+  it "splits input file into the desired number of chunks" do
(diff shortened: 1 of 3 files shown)


aelkiss (commented):
Looks pretty good! A couple of thoughts for the tests:

- Do we care what the names of the chunks are, or only that they exist?
- If we don't care what the file names are -- would it perhaps be worth checking that each chunk has a different file name (as opposed to the chunker saying there are N chunks, but they all refer to the same file?)
- Would it be worth testing that the chunks have the expected size and/or whether the complete set of chunks contains the complete input?

aelkiss (approved):
Nice! A couple small suggestions in-line but otherwise this is looking good to go.

spec/scrub/chunker_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    expect(chunker.chunks.size).to eq 0
+
+    # Run chunker.
+    chunker.run
+
+    # Get the expected number of files.
+    expect(chunker.chunks.size).to eq chunk_count
+
+    # -1 because we skip the header line
+    full_line_count = `wc -l #{pre_chunked_file}`.split.first.to_i - 1
```
aelkiss, line 23: I'd suggest hard-coding this based on what's actually in `pre_chunk.tsv` - which is to say, I think it's fine to make the expected values in the tests dependent on the data going in.

mwarin: yeah that simplifies it a bit

PR #210, opened 2022-09-01
OCLC Registration Report for Shared Print

mwarin:
Generate a report of all commitments given an organization.
Some fields are based on commitment data, some fields are static, and the OCLC collection ID is looked up in a file given oclc_sym.

To run:
`bundle exec ruby bin/phctl.rb report oclc-registration <ORGANIZATION>`

To test:
`bundle exec rspec spec/reports/oclc_registration_report_spec.rb`


aelkiss:
We will need to add the collection id map, probably via configmap for now.

aelkiss (approved):
Looks good - follows the pattern of the other reports and should be easy to integrate in the machinery of sidekiq.

PR #216, opened 2022-09-16
Use shared github actions

aelkiss:
Compare to e.g.

https://github.com/hathitrust/oai_solr/tree/main/.github/workflows
https://github.com/hathitrust/otis/tree/main/.github/workflows

(We don't yet have any deploy actions here.)


aelkiss:
This also removes the build-v1.0.x action which for unclear reasons didn't actually work.

I think with the shared actions, the following will work as a temporary release workflow:

* Use the manual build action with the branch, tag, or commit you want to release
  https://github.com/hathitrust/holdings-backend/actions/workflows/build-manual.yml

* Make a new release pointed to that same commit

Still not ideal, but maybe more tractable than before.

I also talked some to @niquerio about the possibilities of using the github API to coordinate actions around deployment. She pointed out that it might be possible to do some of that directly from the release action rather than needing an external shell script to coordinate, but I think either way will work.

aelkiss:
@moseshll  This is doing basically the same thing as https://github.com/hathitrust/otis/pull/183 (which I merged while you were out). I think it's worth taking a look and understanding what the shared github actions are doing here on at least one of the PRs. The shapes of github actions between different repositories are gravitating towards something closer together.

moseshll (approved):
Glad to see a shared set of facilities take shape. I'm mentally noting two items to think about in future: docker/build-push is up to v3 as of this May (worth looking at whether there are differences of interest), and also multi-arch builds will prolly need to be supported for some projects, like db-image (I think).

PR #217, opened 2022-09-20
Configuration for sidekiq

aelkiss:
* Add location for sidekiq/redis configuration
* Expose more configuration for mongodb -- holdings-testing / mongo 5 needed some different options here

sidekiq-web still isn't working, I suspect because the inline config overrides the provided config in the sidekiq initializer.

### DIFF
--- a/bin/load_test_data.sh
+++ b/bin/load_test_data.sh
@@ -7,7 +7,7 @@
 # Run this outside docker; it will start all required services.
 # Running this script assumes you have already run --rm bin/setup_dev.sh
 
-docker-compose up -d sidekiq_web processor redis
+docker-compose up --scale processor=3 -d sidekiq_web processor redis 
 docker-compose run --rm dev bin/setup/wait-for redis:6379 -- echo "redis is ready"
 
 echo "generating fake data"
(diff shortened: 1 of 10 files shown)


bin/load_test_data.sh
```diff
@@ -7,7 +7,7 @@
 # Run this outside docker; it will start all required services.
 # Running this script assumes you have already run --rm bin/setup_dev.sh
 
-docker-compose up -d sidekiq_web processor redis
+docker-compose up --scale processor=3 -d sidekiq_web processor redis
```
aelkiss, line 10: This allows testing multiple sidekiq processes in development

docker-compose.yml
```diff
@@ -29,11 +28,10 @@ services:
       - .:/usr/src/app
       - gem_cache:/gems
       - ./example/datasets:/tmp/datasets
-    command: bundle exec sidekiq -r ./lib/sidekiq_jobs.rb
+    command: bundle exec sidekiq -c 1 -r ./lib/sidekiq_jobs.rb
```
aelkiss, line 31: This ensures sidekiq uses only a single thread (no benefit to multiple threads, since MRI can only use one CPU core and our jobs tend to be CPU bound)

aelkiss:
Will not resolve the sidekiq-web issue right now; see DEV-424

PR #209, opened 2022-08-26
Dev 299 refresh, draft

mwarin:
draft

PR #218, opened 2022-09-26
Add sidekiq-batch, rspec-sidekiq

aelkiss:
- Enables getting a callback when a batch of jobs completes
- Uses rspec-sidekiq to test callbacks & the batching implementation

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ gem "thor"
 gem "zinzout"
 gem "puma"
 gem "sidekiq"
+gem "sidekiq-batch"
 
 group :development, :test do
   gem "pry"
@@ -29,6 +30,7 @@ group :test do
   gem "factory_bot"
   gem "faker"
   gem "rspec"
+  gem "rspec-sidekiq"
   gem "simplecov"
   gem "simplecov-lcov"
   gem "webmock"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -83,6 +83,9 @@ GEM
     rspec-mocks (3.11.1)
       diff-lcs (>= 1.2.0, < 2.0)
       rspec-support (~> 3.11.0)
+    rspec-sidekiq (3.1.0)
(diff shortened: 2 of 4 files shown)


aelkiss:
@mwarin I recommend getting #209  to the point where we can merge it, then doing something like this here https://github.com/hathitrust/holdings-backend/pull/209/files#diff-ca99b54ef7e5ac35271f7440dc7cf9ee2214ddd251430e6c57d7e38138feee8cR71-R77

```ruby
batch = Sidekiq::Batch.new
batch.description = "Holdings load for #{scrubber_out_file}"
batch.on(:success, Loader::HoldingLoader,
         'tmp_chunk_dir' => @tmp_chunk_dir
         'member' => member
         'logger_path' => scrubber.logger_patch)
batch.jobs do
  chunker.chunks.each do |chunk|
    puts "Load chunk #{chunk}"
    Loader::FileLoader.new(batch_loader: Loader::HoldingLoader.for(chunk)).load(chunk)
  end
end
``` 

and then:

```
class Loader::HoldingLoaderCleanup
  def on_success(status, options)
    FileUtils.rm_rf(options['tmp_chunk_diri'])
    upload_to_member(options['member'], options['logger_path'])
  end

  # move upload_to_member here
end
```

and then testing with something like:
```ruby
it "cleans up old holdings after loading new ones" do
  # load a few 'old' records
  ...
  # load a new thing with scrubrunner that doesn't completely match
  batch = ScrubRunner.new(...).run_file(...)
  batch.status.join

  # check that the old records are gone
  ...
end

```


Note all the options for `batch.on` need to be JSON-serializable, since they all end up serialized/deserialized as a sidekiq job via redis, but you may not get warned in that in testing because of the way rspec-sidekiq mocks the batching system. In particular I'm not sure what "member" is (is it just an institution ID string, or the actual member data structure?)

Alternatively, we could first pull this commit in over on that branch (once I see whatever isn't working in CI) and add the batching stuff straightaway.

mwarin (approved):

PR #219, opened 2022-10-12
fix for DEV-454

mwarin:
Problem: You cannot update `local_id` or `ocn` with the existing shared print update-script.
Solution: Now, in order to update `local_id` and/or `ocn`, add the columns `new_local_id` and/or `new_ocn` to the update file, like:

```
organization  |ocn       |local_id           |new_ocn
bu            |8061382   |99193019080001161  |898915865
bu            |6785135   |99191292180001161  |263616332
bu            |4738078   |99177938200001161  |76925351
```
The "new_" prefix helps us distinguish the update field (`new_local_id`, `new_ocn`) from the lookup field (`local_id`, `ocn`).
Pass to `phctl sp update <file>` as before.

### DIFF
--- a/lib/shared_print/update_record.rb
+++ b/lib/shared_print/update_record.rb
@@ -85,7 +85,9 @@ module SharedPrint
         local_item_location: :string,
         local_shelving_type: :string,
         oclc_sym: :string, # might be called oclc_symbol in places, TODO: check that.
-        retention_date: :date_time
+        retention_date: :date_time,
+        new_local_id: :string,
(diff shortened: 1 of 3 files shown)


aelkiss (approved):
This feels like a workaround for something maybe more fundamental in terms of the interface for updating records. Perhaps the issue is that we don't have a persistent identifier as such for the commitment itself rather than by reference to something else. This is certainly OK for now, but it might be worth working through the implications for the data model more generally at some future point in time.

PR #221, opened 2022-10-18
DEV-405, DEV-424: sidekiq/redis configuration

aelkiss:
- inline configuration when using redis sentinel (rather than putting in ht_tanka)
- ensure sidekiq web doesn't clobber existing redis configuration

Basically, we put the configuration we want in `Services.redis_config` and can get at it / modify it there as needed; the default if we try to use it without configuring it is to raise an exception. The `Sidekiq.configure_server` / `Sidekiq.configure_client` blocks are basically setting up callbacks to tell Sidekiq where to get the configuration from when it needs it rather than providing it directly.. this is just the style of configuration that Sidekiq (and many other Ruby gems like rspec) use, but it does make things a bit more confusing

### DIFF
--- a/bin/sidekiq_web.ru
+++ b/bin/sidekiq_web.ru
@@ -11,9 +11,7 @@ if !File.exist?(SESSION_KEY)
   end
 end
 
-Sidekiq.configure_client do |config|
-  config.redis = {size: 1}
-end
+Services.redis_config[:size] = 1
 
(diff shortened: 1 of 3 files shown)


aelkiss:
See corresponding change in ht_tanka: https://github.com/hathitrust/ht_tanka/pull/44

aelkiss:
I would like sidekiq to comment about coverage but not fail the build. I think I changed the correct setting ("use status API") to do that, but we'll see :shrug:

PR #222, opened 2022-10-18
DEV-461 Add option to phctl to run inline

aelkiss:
Having a global option for thor is a little bit clunky, but using the "class_option" on the parent class and then handling the option in a single place that the subcommands inherit from seems to work. If there is a better way to do this with Thor, I couldn't find it.

### DIFF
--- a/lib/phctl.rb
+++ b/lib/phctl.rb
@@ -8,98 +8,114 @@ $LOAD_PATH.unshift(File.dirname(__FILE__))
 # or on the cluster with
 # `ht_tanka/environments/holdings/jobs/run_generic_job.sh bin/phctl.rb <command>`
 module PHCTL
-  class Load < Thor
+  class JobCommand < Thor
+    no_commands do
+      def run_job(klass, *args, **kwargs)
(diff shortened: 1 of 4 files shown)


lib/phctl.rb
```diff
@@ -122,16 +138,23 @@ def uncommitted_holdings
...
-      Jobs::Common.perform_async("Reports::OCLCRegistration", options, organization)
+      run_common_job(Reports::OCLCRegistration, options, organization)
     end
   end
 
   class PHCTL < Thor
+    # Run inline instead of with sidekiq
+    class_option :inline, type: :boolean
+
+    def self.exit_on_failure?
```
aelkiss, line 154: This suppresses a warning from Thor about not having this here.

aelkiss: And probably prevents cron jobs etc from falsely thinking they're successful.

aelkiss:
Note that coveralls still posted its comment but did not include the status check (which would have been OK in this case)

mwarin (approved):
That's really cool.

PR #223, opened 2022-10-19
Dev 299 final

mwarin:
Connecting pieces so that `bundle exec ruby bin/phctl.rb scrub <org>` works.
It: 
* checks dropbox for <org>'s files and compares them to local files.
* downloads any new files
* scrubs them
* chunks them
* loads them
* uploads report back in member dropbox for each successfully processed file


lib/scrub/autoscrub.rb
```diff
@@ -26,12 +26,16 @@ def initialize(path)
...
       @output_dir = @output_struct.date_subdir!("output")
       @log_dir = @output_struct.date_subdir!("log")
+      @out_files = []
 
       # Once we have @member_id and @item_type,
       # build a log path and re-register the service logger to log to that path
-      logger_path = File.join(@log_dir, "#{@member_id}_#{@item_type}.log")
+      ymd = Time.new.strftime("%F")
+      @logger_path = File.join(@log_dir, "#{@member_id}_#{@item_type}_#{ymd}.log")
+      puts "autoscrub logging to #{@logger_path}"
```
aelkiss, line 35: Extraneous/duplicate -- we're logging it via `Services.logger` at line 48 already

lib/scrub/autoscrub.rb
```diff
@@ -84,13 +92,18 @@ def run
...
+        File.split(out_file_path).last
+      )
     end
 
     private
 
     def count_file_lines
       # zcat -f works as plain cat if @path is not in gzip format.
-      `zcat -f #{@path} | wc -l`.match(/^(\d+)/)[0].to_i
+      `zcat -f "#{@path}" | wc -l`.match(/^(\d+)/)[0].to_i
```
aelkiss, line 106: Consider doing this in Ruby -- not as elegant for opening a maybe-gzipped file, although we could extract that out as a helper somewhere (we're doing it in a couple other places in the code base already e.g. in lib/concordance_validation). I think this should do it:

```ruby
require 'zlib'

if @path.end_with?(".gz") then
  Zlib::GzipReader.open(@path)
else
  File.open(@path)
end.count
```

aelkiss (changes requested):
Started looking at this, will look at it more later. I definitely want to remove the `puts`; up to you on the gzip thing.

lib/scrub/chunker.rb
```diff
@@ -9,31 +10,47 @@ module Scrub
...
       # `split --number=l/x` only works on files, not on STDIN.
       # TODO: we might want another dir for sort -T .
-      tmp_file = "#{@work_dir}/#{@uuid}/tmp_sorted.txt"
+      tmp_file = File.join(@tmp_chunk_dir, "tmp_sorted.txt")
+      add_uuid_call = @add_uuid ? "| bundle exec ruby bin/add_uuid.rb" : ""
       sort_call = "egrep -vh '^OCN' #{@glob} | " \
-                  "sort -s -n -k1,1 -T ./ | " \
-                  "bundle exec ruby bin/add_uuid.rb > #{tmp_file}"
+                  "sort -s -n -k1,1 -T ./ " \
+                  "#{add_uuid_call} " \
```
aelkiss, line 47: I waffled a little bit about how I felt about the external calls here. Ultimately, I think it's OK; otherwise, there would have to be more temporary files or complicated pipe logic, and add_uuid.rb is pretty simple.

lib/scrub/chunker.rb
```diff
@@ -43,29 +60,30 @@ def run
...
+      FileUtils.mv(fn, "#{fn}.#{@out_ext}")
       # Return new name
-      "#{fn}.tsv"
+      "#{fn}.#{@out_ext}"
     end
 
     # These can start taking up a lot of space and are ~worthless once loaded.
     def cleanup!
-      FileUtils.rm_rf("#{@work_dir}/#{@uuid}")
+      puts "rm -rf #{@tmp_chunk_dir}"
```
aelkiss, line 85: Remove puts or use Services.logger

lib/scrub/scrub_runner.rb
```diff
@@ -0,0 +1,125 @@
...
+      if Settings.remote_member_data.nil?
+        raise "Need Settings.remote_member_data to be set"
+      end
+      if @organization.nil?
+        raise "Need @organization to be set"
+      end
+    end
+
+    def run
+      puts "Running org #{@organization}."
```
aelkiss, line 46: We should use `Services.logger` in preference to `puts` everywhere here

lib/scrub/scrub_runner.rb
```diff
@@ -0,0 +1,125 @@
...
+      end
+    end
+
+    def run_file(file)
+      puts "Running file #{file}"
+      downloaded_file = download_to_work_dir(file)
+      scrubber = Scrub::AutoScrub.new(downloaded_file)
+      scrubber.run
+      scrubber.out_files.each do |scrubber_out_file|
+        puts "Ready to split #{scrubber_out_file} into chunks"
```
aelkiss, line 60: Consider extracting the contents of the block into a method on Chunker, perhaps like:

```ruby
scrubber.out_files.each do |scrubber_out_file|
  Scrub::Chunker.new(glob: file, chunk_count: 4, out_ext: "ndj").chunk_and_load
  upload_to_member(scrubber.logger_path)
end

class Scrub::Chunker
  def chunk_and_load(file)
   Services.logger.info("Ready to split #{file} into chunks")
   run
   load_chunks
   cleanup!
  end

  def load_chunks
    chunks.each do |chunk|
     Services.logger.info("Load chunk #{chunk}")
     Loader::FileLoader.new(batch_loader: Loader::HoldingLoader.for(chunk)).load(chunk)
    end
  end
end
```

lib/scrub/scrub_runner.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    # Check org-uploaded files for any not previously seen files
+    def check_new_files
+      # Return new (as in not in old) files
+      remote_files = @ft.lsjson(remote_dir)
+      old_files = check_old_files
+
+      # Include in new_files only those remote_files whose name is not in old_files.
+      new_files = []
+      remote_files.each do |f|
```
aelkiss, line 85: I think the intention is that `new_files` should contain anything that is:
- in `remote_files`
- does not end in `.log`
- is not in `old_files`

in which case you might be able to do something like:

```ruby
remote_files.reject { |f| f["Name"].end_with?(".log") } - old_files
```

That would compare more fields than just the name, but perhaps you could do:

```ruby
new_file_names = remote_files.map {|f| f["Name"] }.reject { |f| f.ends_with?(".log") } - old_files.map { |f| f["Name"] }
```

spec/integration_spec.rb
```diff
@@ -155,4 +155,23 @@ def cleanup(output)
...
+      local_d = "/tmp/local_member_data/umich-hathitrust-member-data/print\ holdings/#{Time.new.year}/"
+      remote_d = "/tmp/remote_member_data/umich-hathitrust-member-data/print\ holdings/#{Time.new.year}/"
+      logfile = File.join(remote_d, "umich_mono_#{Date.today}.log")
+      FileUtils.rm_rf(local_d)
+      FileUtils.mkdir_p(remote_d)
+      FileUtils.touch("/tmp/rclone.conf")
+      FileUtils.rm_rf(logfile)
+      FileUtils.cp("spec/fixtures/umich_mono_full_20220101.tsv", remote_d)
+      # Actual tests:
+      expect(File.exist?(File.join(remote_d, "umich_mono_#{Date.today}.log"))).to be false
```
aelkiss, line 171: This is just verifying a precondition, right? Might be worth putting that before the "actual tests" comment to clarify that nothing has happened yet

spec/scrub/auto_scrub_spec.rb
```diff
@@ -1,14 +1,8 @@
 # frozen_string_literal: true
 
-require "spec_helper"
 require "scrub/autoscrub"
 
 RSpec.describe Scrub::AutoScrub do
-  before(:each) do
```
aelkiss, line 7: It was unclear to me whether the specs will hit `OCLC_URL` now or not, but I'm thinking they still would when running in Github Actions (i.e. we aren't committing the cached max_ocn, right?) If so, it might be worth continuing to stub this request in the specs.

lib/scrub/chunker.rb
```diff
@@ -9,31 +10,47 @@ module Scrub
   # #{chunk_count} equal-sized (roughly) files from them.
   class Chunker
     attr_reader :glob, :chunks
-    def initialize(glob, chunk_count = 16)
+    def initialize(glob: "/dev/null", chunk_count: 16, add_uuid: false, out_ext: nil)
```
aelkiss, line 13: Under what circumstances would we call `Chunker.new` with no glob? Likewise, we check that `out_ext` isn't nil on line 34, so could we just make that a required parameter?

lib/scrub/chunker.rb
mwarin: Yeah I did not feel great writing this.

lib/scrub/autoscrub.rb
mwarin: Line 48 is `Services.logger` ($stderr) saying that we logging some specifics to the file `@logger_path`. These are 2 separate logs. Could perhaps be done clearer, but it is intentional.

lib/scrub/autoscrub.rb
mwarin: Added and plugged in `Utils::LineCounter`.

lib/scrub/scrub_runner.rb
mwarin: I see what you mean, but I like it better having the coordination done by (the already necessarily more complex) `ScrubRunner`, and leaving `Chunker` with a more limited scope.

lib/scrub/scrub_runner.rb
mwarin: I don't see how this improves upon or makes easier to read what's already there.

lib/scrub/chunker.rb
mwarin: good point. done

spec/scrub/auto_scrub_spec.rb
mwarin: I don't understand what your concern is here.
The stub is moved to `spec_helper.rb:85`, and it seems to be working just fine?
And no, we are not committing the max ocn.

aelkiss: I missed that the stub was moved to `spec_helper`, so I think it's fine as-is. Without the stub, I was concerned that 1) we introduce an external dependency into the specs and 2) every time we run the specs in github it would hit `OCLC_URL` for real. But we still have the stub, so it doesn't matter.

lib/scrub/autoscrub.rb
aelkiss: I'm not sure I understand, but regardless we shouldn't be using `puts`

lib/scrub/scrub_runner.rb
aelkiss: :+1:

lib/scrub/scrub_runner.rb
aelkiss: I found it difficult to understand the intention of what `new_files` was doing without tracing the code (although it is noted in the spec). That said, I'm fine with leaving what we have here for the time being. Perhaps if we find that we need to change the logic here in the future, we could consider havingremote_files` return an array of objects instead of an array of hashes. Then we have some additional options for encapsulating the logic here and making the operations more clear.

lib/scrub/autoscrub.rb
aelkiss: And now we aren't, so everything is fine

spec/scrub/auto_scrub_spec.rb
aelkiss: Actually, I'm not seeing it in `spec_helper` - missed committing it?

aelkiss (changes requested):
See comment about `OCLC_URL` and `spec_helper.rb` - the stub does not appear to be there in this branch. Once that's taken care of, I think this is good to squash & merge.

spec/spec_helper.rb
```diff
@@ -82,6 +82,7 @@
     # stub external APIs
     Services.register(:pushgateway) { instance_double(Prometheus::Client::Push, add: true) }
     Services.register(:slack) { instance_double(Utils::SlackWriter, write: true) }
+    stub_request(:get, "https://www.oclc.org/apps/oclc/wwg").to_return(body: '{ "oclcNumber": "1000000000" }')
```
mwarin, line 85: @aelkiss here it is

aelkiss (approved):
All looks good now. Go ahead and merge when ready.

PR #224, opened 2022-10-25
Update dependencies

aelkiss:
- update gems
- update mongo to mongodb 6.0.2
- remove unused prep_loadfiles.sh


spec/loader/cluster_loader_spec.rb
```diff
@@ -16,7 +16,7 @@
   end
 
   def count_clusters
-    Cluster.find_by(ocns: ocn).to_a.size
+    Cluster.for_ocns([ocn]).count
```
aelkiss, line 19: `Cluster.find_by` returns a single cluster or nil; with the updated `mongoid` gem, `to_a` on a single cluster did not work. The change seems to better reflect the intention expressed in the method name.

aelkiss:
@mwarin  Any concerns?

spec/loader/cluster_loader_spec.rb
mwarin: Do we gotta change:
```
clustering/cluster_holding.rb:51:        c = Cluster.find_by(ocns: holding.ocn)
reports/estimate.rb:31:        cluster = Cluster.find_by(ocns: ocn.to_i,
reports/etas_organization_overlap_report.rb:99:        cluster = Cluster.find_by(_id: c)
```

as well then?

mwarin (approved):
Looks fine, just one q that doesn't need to block merging.

spec/loader/cluster_loader_spec.rb
aelkiss: I checked for that as well, and I don't think so -- the problem was the `to_a`, not the `find_by` as such.

PR #225, opened 2022-10-26
getting scrubrunner to batch load with sidekiq

mwarin:
This PR adds `Sidekiq::Batch` to `Scrub::ScrubRunner`.
When a batch completes it triggers some cleanup (done by `Loader::HoldingLoader::Cleanup`), and we have to fake this in testing (where sidekiq is not running), hence the `force_holding_loader_cleanup_test` stuff.

### DIFF
--- a/lib/loader/holding_loader.rb
+++ b/lib/loader/holding_loader.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "clustering/cluster_holding"
+require "utils/file_transfer"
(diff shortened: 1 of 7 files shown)


aelkiss (approved):
Looks good. If we at some point change `ScrubRunner` to fully honor `--inline` (i.e. have it run its holdings load jobs inline rather than async) we can use the same shape and just remove the language about "for testing only".

PR #226, opened 2022-10-28
Dev 324

mwarin:
Adds a check of the record count of the most recent comparable loadfile, and warns (so far only in the logs) if the abs diff% against the current scrub is greater than some value. That value is `Settings.scrub_line_count_diff_max` and is a percentage expressed as a  float (where 0.0 is 0% and 1.0 is 100%). 

`scrub_line_count_diff_max: 0.05` means a submitted file with a scubbed record diff (compared to the last loaded file with the same org & item_type) greater than +5% or smaller than -5% will be rejected with a warning.

This check is done after scrubbing but before chunking and loading, so we can bail out early. It also follows the old style approach of counting scrubbed records rather than submitted records.

If you want to force through an update anyways, set `@force=true` either by
`Scrub::ScrubRunner.new(org, {"force" => true})`
or via phctl
`$ bundle exec ruby bin/phctl.rb scrub <org> --force`

### DIFF
--- a/lib/scrub/scrub_runner.rb
+++ b/lib/scrub/scrub_runner.rb
@@ -155,6 +155,8 @@ module Scrub
       File.join(local_dir, File.split(remote_file).last)
     end
 
+    private
+
     def remote_dir
       DataSources::DirectoryLocator.for(:remote, @organization).holdings_current
     end
--- a/lib/scrub/scrub_runner.rb
+++ b/lib/scrub/scrub_runner.rb
@@ -15,9 +15,12 @@ require "loader/file_loader"
 require "loader/holding_loader"
 require "scrub/autoscrub"
 require "scrub/chunker"
+require "scrub/malformed_file_error"
+require "scrub/record_counter"
 require "sidekiq_jobs"
 require "sidekiq/batch"
 require "utils/file_transfer"
+require "utils/line_counter"
 
 # Example:
 # runner = Scrub::ScrubRunner.new(ORG)
@@ -28,9 +31,11 @@ module Scrub
   class ScrubRunner
     def initialize(organization, options = {})
       @organization = organization
+      # @force: force loading a file even if it exceeds diff limit
+      @force = options["force"] || false
+      # @force_holding_loader_cleanup_test: only set to true in testing.
+      @force_holding_loader_cleanup_test = options["force_holding_loader_cleanup_test"] || false
       @ft = Utils::FileTransfer.new
-      # Only set force_holding_loader_cleanup_test to true in testing.
-      @force_holding_loader_cleanup_test = options["force_holding_loader_cleanup_test"]
       validate
     end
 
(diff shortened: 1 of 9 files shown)


moseshll (approved):
Looks good here.

lib/scrub/scrub_runner.rb
```diff
@@ -60,17 +53,36 @@ def run_file(file)
       downloaded_file = download_to_work_dir(file)
       scrubber = Scrub::AutoScrub.new(downloaded_file)
       scrubber.run
+
+      rc = Scrub::RecordCounter.new(@organization, scrubber.item_type)
+      unless rc.acceptable_diff? || @force
+        raise MalformedFileError, [
```
aelkiss, line 59: A couple questions here:

- Should `MalformedFileError` know how to generate this message itself (given a `Scrub::RecordCounter`?)
- Should `Scrub::RecordCounter` raise the exception itself (perhaps something like `rc.ensure_acceptable_diff!`)?

lib/scrub/scrub_runner.rb
```diff
@@ -86,12 +98,29 @@ def run_file(file)
...
+    rescue => err
       # If the scrub failed, remove the file from local storage, that we may try again.
+      Services.logger.error err
       FileUtils.rm(downloaded_file)
       raise "Scrub failed, removing downloaded file #{downloaded_file}"
     end
 
+    # If we just scrubbed a umich mono we want to know what the line count was
+    # in the most recent umich mono so we can warn if that number is way +/-.
+    def count_lines_last_loaded(scrubber)
```
aelkiss, line 110: Doesn't appear to be called anywhere except in the tests? I'd also advocate for moving this to `Scrub::AutoScrub` - it looks like all the data it's using is from `scrubber`

aelkiss (commented):
The tests look good to me. I'm confused in that `count_last_lines_loaded` only appears to be called in the tests. Is there more forthcoming that would use it, or should we remove it?

lib/scrub/scrub_runner.rb
mwarin: Yeah I don't know why I did that. Not needed & removed.

lib/scrub/scrub_runner.rb
mwarin: * I don't think so
* I don't think so

But I could be convinced otherwise.

aelkiss: `run` is definitely starting to get a bit on the lengthy side with the logic for several different things here happening. The two options I see that would help with that are as described above (at least in terms of `MalformedFileError` knowing how to generate this message) and the innards of the block starting here https://github.com/hathitrust/holdings-backend/pull/226/commits/5d5aab2edba1c0ee5efcd50cebb8e1298973a5be#diff-ca99b54ef7e5ac35271f7440dc7cf9ee2214ddd251430e6c57d7e38138feee8cL64 (as previously suggested). I'd certainly be hesitant to make more changes to `run` here without extracting some of the pieces.

PR #227, opened 2022-11-02
fix push metrics

aelkiss:
With the old version of prometheus client, metrics that we had registered but not actually set a value for did not get sent to the push gateway. With the newer version, they do, which reset last success time & expected success interval to '0' at every job start, which caused spurious alerts.

This version substantially changes how metrics & the specs work. Now, metrics are lazily registered by the push metrics marker when they are actually needed. This also removes some of the stubbing in the specs for prometheus stuff while adding integration tests to ensure that pushing to the push gateway does actually work.

In doing this I found that the behavior in rspec with various hooks running at spread-out times got pretty confusing esp. regarding stubbing the push gateway and at what time that happened. I'm not sure what to do about that, but I still think that stubbing the pushgateway in tests in general makes sense.

Any way, let me know what you think. I also think this code is ripe for merging with https://github.com/billdueber/milemarker (and moving to hathitrust/mlibrary) but I'm not quite ready to do that yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
         run: docker-compose run --rm dev bundle exec standardrb
 
       - name: Run tests
-        run: docker-compose run --rm -e MONGOID_ENV=test dev bin/setup/wait-for mariadb:3306 -- bundle exec rspec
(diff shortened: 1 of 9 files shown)


mwarin (approved):
I don't get everything that is going on but I didn't before either. 
I see nothing that particularly raises my eyebrows, and it all seems to work, so that's good enough for me.

PR #228, opened 2022-11-02
Update dependencies

aelkiss:
- ruby 3.1
- gems incl sidekiq 7


spec/spec_helper.rb
```diff
@@ -25,7 +25,6 @@
   SimpleCov::Formatter::HTMLFormatter,
   SimpleCov::Formatter::LcovFormatter
 ])
-SimpleCov.start
```
mwarin, line 28: Is this not needed anymore / is it being started somewhere else?

mwarin (approved):
Approved with one questionmark re: simplecov in spec_helper

spec/spec_helper.rb
aelkiss: This is included in https://github.com/hathitrust/holdings-backend/blob/main/.simplecov

PR #230, opened 2022-11-03
DEV-500, enable deprecation and update of commitments based on uuid

mwarin:
When updating and/or deprecating commitments, we use `organization + ocn + local_id` as the identifying fields.
That's just by convention, though. So it is possible, usually because of legacy data, that 2+ commitments share the same `organization + ocn + local_id`. The updater and deprecator scripts are aware of this issue, and won't let you update/deprecate a commitment if they find more than 1 commitment given a `organization + ocn + local_id`.

Because we still might want to update/deprecate commitments that have "duplicates" we need to expand the search params to include `uuid` which should (all but) guarantee unique commitments.


mwarin:
this wasn't going anywhere

PR #232, opened 2022-11-10
DEV-505: Only use resolutions & htitems when updating OCNs

aelkiss:
This change ensures that when OCNs in a cluster change, we only use the OCNs for the resolutions and the htitems. In normal usage, holdings should never end up in the "wrong" cluster. However, a small number of shared print commitments had their OCNs changed without moving clusters, leading to inconsistent clusters. This update mitigates the effect of that. DEV-504 will address the underlying issue.

### DIFF
--- a/lib/cluster.rb
+++ b/lib/cluster.rb
@@ -178,7 +178,7 @@ class Cluster
   end
 
   def update_ocns
-    self.ocns = [clusterable_ocn_tuples].flatten.uniq
+    self.ocns = [ocn_resolutions.pluck(:ocns) + ht_items.pluck(:ocns)].flatten.uniq
     save
   end
 


aelkiss:
Note that I did *not* change `clusterable_ocn_tuples` because that was being used in other places and it was unclear to me whether removing holdings & commitments from that list could result in problems when reclustering - my sense was that it could a,if a cluster loses its "glue" and then we need to recluster the holdings/commitments.

Another option could be to at cluster save time validate that all the holdings & commitments in the cluster have OCNs matching `cluster.ocns`. That may be worth considering as a follow-up once DEV-504 is implemented to help maintain referential integrity.

mwarin (approved):
I had the same question as raised by @aelkiss in https://github.com/hathitrust/holdings-backend/pull/232#issuecomment-1310494003 .

No test but that's fine.

aelkiss:
Right. I could add a test, but since clusters shouldn't end up in this state to begin with it seems better to test that it won't continue to happen (on DEV-504) and it's sufficient here that it doesn't break any existing tests.

PR #233, opened 2022-11-11
Potentially move commitment to another cluster if updating commitment OCN

mwarin:


aelkiss:
@mwarin Probably worth looking at the new uncovered lines in shared print updater https://coveralls.io/builds/55292645/source?filename=lib%2Fshared_print%2Fupdater.rb#L74 -- there are several cases we could/should test there

mwarin:
100% coverage for `/lib/shared_print/updater.rb` ought to do it.

spec/shared_print/updater_spec.rb
```diff
@@ -97,4 +96,80 @@
...
+      # set up a cluster with ocn1 and ocn2
+      resolution = build(:ocn_resolution, deprecated: ocn1, resolved: ocn2)
+      cluster = build(:cluster, ocns: [ocn1, ocn2])
+      cluster.save
+      cluster.add_ocn_resolutions(resolution)
+      expect(cluster.ocns).to eq [ocn1, ocn2]
+
+      # add an spc that has ocn1 (and a matching ht_item)
+      cluster_tap_save [spc1, ht1]
+      expect(Cluster.count).to eq 1
```
aelkiss, line 136: I waffle a bit about these expectations -- on the one hand I think it's worthwhile as a kind of documentation for what the precondition is especially given the somewhat complex setup, but it also maybe muddies what the test is actually testing (lines 149-151). If nothing else, I think it would help to be more explicit with comments about what's setup, what's preconditions, and what the actual test is. It also might be worth refactoring some of the setup in these tests and in other similar tests - if we can find any common patterns or abstract the variation in setup some way it might in the end make the tests easier to understand. I think we have the

aelkiss: https://hathitrust.atlassian.net/browse/DEV-546

aelkiss (approved):
The additional tests look good; the comment about potentially abstracting test setup in the future (https://hathitrust.atlassian.net/browse/DEV-546) doesn't need to block merging.

PR #234, opened 2022-11-23
fail more gracefully if ocn.nil?, DEV-513

mwarin:
Autoscrub would previously crash if given an OCN without a numeric part.


aelkiss (approved):
Looks good. You might consider moving `out_file.close` into an `ensure` block to clarify the intention, but I think it should work fine as is.

PR #236, opened 2022-11-23
fail gracefully when scrubbing an empty line

mwarin:
autoscrub would previously crash out when encountering an empty file


aelkiss (approved):
Looks good; I see the existing `Scrub::MalformedRecordError` handling in `each_holding`.

PR #237, opened 2022-11-28
DEV-515, handle scrubbing file that has a bigger diff than we like without crashing

mwarin:
It used to be the case that if one of the files being scrubbed broke the "no greater/smaller diff than x" rule, the whole scrub session crashed out because exceptions were not caught properly. This fixes that.

### DIFF
--- a/lib/scrub/scrub_runner.rb
+++ b/lib/scrub/scrub_runner.rb
@@ -62,8 +62,9 @@ module Scrub
           "The scrubbed file (#{rc.last_ready}) has #{rc.count_ready} records.",
           "Diff is #{rc.diff}, which is greater than Settings.scrub_line_count_diff_max",
           "... which is #{Settings.scrub_line_count_diff_max}.",
-          "Run again with --force to load anyways."
+          "This file will not be loaded."
         ].join("\n")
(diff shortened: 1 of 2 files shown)


lib/scrub/scrub_runner.rb
```diff
@@ -98,11 +99,11 @@ def run_file(file)
           Loader::HoldingLoader::Cleanup.new.on_success(:success, cleanup_data)
         end
       end
-    rescue => err
+    rescue MalformedFileError => err
```
aelkiss, line 102: Took me a minute to understand this -- now we are catching specifically `MalformedFileError` rather than logging & re-raising an error here. But, we are also changing the behavior not to catch unexpected errors & remove the downloaded file -- is that OK?

aelkiss (approved):
I like the change in the test to checking that the expected side-effect happened or not. See inline comment about no longer catching unexpected errors, if that's what we want.

lib/scrub/scrub_runner.rb
mwarin: Yeah you are right... this `rescue` _does_ do the job but it is too specific. Changing it.

PR #240, opened 2022-12-02
attempt to fix the sidekiq callback issue in DEV-522

mwarin:


aelkiss (approved):
Looks fine to me, especially if you tried it in dev & verified that the callback ran

PR #243, opened 2022-12-13
Don't pass --inline option through to common job

aelkiss:
Most don't know how to handle it


mwarin:
Will probably be great once it passes standardrb.

mwarin (approved):
Looks alright and works as intended.

PR #244, opened 2022-12-19
Adding class for mass-update of embedded documents in mongo

mwarin:
To address cases where we want to do a mass-update and don't want to deserialize-serialize every document.


aelkiss (approved):
I looked at this with @mwarin and it all makes sense to me. Although the tests aren't necessarily the most realistic cases I think they still demonstrate the operation and cover the functionality sufficiently.

PR #245, opened 2022-12-22
DEV-523: shared print policies; commitment tsv loading

aelkiss:
@mwarin  I think this covers the needs of DEV-523 as I understand them:

- validate policies being a specific list
- allow TSV loading of commitments, including policies

It does NOT:
- normalize other fields coming in to lowercase (either for TSV or JSON), in particular local shelving type
- handle the alleged "Do not reproduce" policy (I'm not sure how that would be formatted in practice)

Please feel free either to kick it back to me or take it over if there are additional needs.


lib/loader/file_loader.rb
```diff
@@ -36,7 +36,8 @@ def skip_matching_header(filehandle, skip_match)
     # @return [Integer] Number of records loaded
     def load(filename, filehandle: Zinzout.zin(filename), skip_header_match: nil)
       logger.info "Loading #{filename}, batches of #{ppnum @batch_size}"
-      skip_matching_header(filehandle, skip_header_match).each.lazy
+
+      skip_matching_header(filehandle.enum_for(:each), skip_header_match).lazy
```
aelkiss, line 40: this change ensures we get an enumerator back regardless of what filehandle was to begin with -- all we need is for it to respond to `:each`

lib/utils/tsv_reader.rb
```diff
@@ -40,12 +42,12 @@ def records
       # For each line, get a copy of header_hash and populate it using header_index
       # so that cols_hash = {ocn: 11, local_id: i11}
       @inf.each_line do |line|
-        yield line_to_hash(line.strip)
+        yield line_to_hash(line.chomp)
```
aelkiss, line 45: the changes here are to ensure we can handle trailing empty columns (not sure whether we'd expect to see that in practice or not)

spec/clusterable/commitment_spec.rb
```diff
@@ -19,16 +19,16 @@
   end
 
   it "validates local_shelving_type" do
-    expect(comm.valid?).to be true
+    expect(comm).to be_valid
```
aelkiss, line 22: with all these changes, rspec will better describe failures - rspec can dynamically generate matchers for any predicate ending in `?` that returns a boolean

aelkiss:
Decrease in coverage is primarily filename exception branches in shared print loader - I'm not too worried about that

lib/utils/tsv_reader.rb
mwarin: yeah that's a change for the better

spec/clusterable/commitment_spec.rb
mwarin: :exploding_head:

lib/utils/tsv_reader.rb
```diff
@@ -40,12 +42,12 @@ def records
...
       end
     end
 
     def line_to_hash(line)
-      cols = line.split(@delim)
+      cols = line.split(@delim, -1)
```
mwarin, line 50: Why the -1 here? I don't understand.

mwarin (approved):

PR #246, opened 2022-12-23
DEV-412, shared print overlap groups

mwarin:
Added shared print groups and related overlap column in the rare_uncommitted report.

Now if you run a rare_uncommitted report for an organization that is one of the `SharedPrint::Groups`, you get the additional column "overlap_group" which counts how many of the other members in your group(s) have holdings on the same cluster.

### DIFF
--- a/lib/shared_print/groups.rb
+++ b/lib/shared_print/groups.rb
@@ -1,0 +1,88 @@
+module SharedPrint
+  # A lookup object for which orgs are members of which shared print program,
+  # and vice versa.
+  class Groups
+    attr_reader :groups
+
+    def initialize
+      @groups = {
+        "coppul" => [
+          "ucalgary",
+          "ualberta"
+        ],
+        "east" => [
+          "brandeis",
+          "brynmawr",
+          "bu",
+          "colby",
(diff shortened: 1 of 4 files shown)


lib/shared_print/groups.rb
```diff
@@ -0,0 +1,88 @@
+module SharedPrint
+  # A lookup object for which orgs are members of which shared print program,
+  # and vice versa.
+  class Groups
+    attr_reader :groups
+
+    def initialize
+      @groups = {
```
moseshll, line 8: If it all possible I would extract this `@groups` data into a JSON or YAML config-ish file and read it in.

moseshll (approved):
See single comment about extracting data. Otherwise looks and runs fine for me.

lib/shared_print/groups.rb
mwarin: In general I'd agree with this, but the SharedPrint::Groups is such a specific bit of data that I would like to have committed anyways, and I don't foresee a lot of changes to it. I'm perfectly happy hardcoding this (famous last words).

PR #247, opened 2023-01-09
DEV-523, `NON-REPRO` policy for commitments

mwarin:
Adding new shared print policy `NON-REPRO` (formerly hypothesized as `do not reproduce`) and some associated tests.
For context: `policies` is an array field on `Clusterable::Commitments` that may take a number of different values.
We expect to get it from members in the form of a comma-separated text field in a `.tsv` file.

### DIFF
--- a/lib/utils/tsv_reader.rb
+++ b/lib/utils/tsv_reader.rb
@@ -1,3 +1,5 @@
+require "English"
+
 module Utils
   class TSVReader
     # Takes a .tsv file with a header and turns it into hashes,
@@ -49,7 +51,7 @@ module Utils
     def line_to_hash(line)
       cols = line.split(@delim, -1)
       if cols.size != @header_hash.keys.size
-        raise IndexError, "Line cols: #{cols.size}, header cols: (#{@header_hash.keys.size})."
+        raise IndexError, "Line #{$INPUT_LINE_NUMBER} has #{cols.size} cols, header has #{@header_hash.keys.size} cols."
       end
       col_hash = @header_hash.clone
       cols.each_with_index do |col_val, i|
--- a/spec/clusterable/commitment_spec.rb
+++ b/spec/clusterable/commitment_spec.rb
@@ -122,7 +122,11 @@ RSpec.describe Clusterable::Commitment do
   end
 
   describe "policies" do
-    ["blo", "digitizeondemand", "non-circ"].each do |policy|
+    it "accepts empty policy" do
+      expect(build(:commitment, policies: [])).to be_valid
+    end
+
+    ["blo", "digitizeondemand", "non-circ", "non-repro"].each do |policy|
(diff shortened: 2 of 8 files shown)


lib/utils/tsv_reader.rb
```diff
@@ -49,7 +49,7 @@ def records
     def line_to_hash(line)
       cols = line.split(@delim, -1)
       if cols.size != @header_hash.keys.size
-        raise IndexError, "Line cols: #{cols.size}, header cols: (#{@header_hash.keys.size})."
+        raise IndexError, "Line #{$.} has #{cols.size} cols, header has #{@header_hash.keys.size} cols."
```
aelkiss, line 52: I'm surprised standardrb didn't complain about the perlism - we should prefer `require 'English'` and then we can say `$INPUT_LINE_NUMBER` instead of `$.`

spec/loader/shared_print_loader_spec.rb
```diff
@@ -68,4 +69,22 @@
...
+      Loader::FileLoader.new(batch_loader: described_class.for(fixt))
+        .load(fixt, filehandle: described_class.filehandle_for(fixt))
+      # Get policies on all loaded commitments
+      all_policies = Cluster.each.map do |cluster|
+        cluster.commitments.map do |commitment|
+          commitment.policies.class
+        end
+      end
+      # Check that the loaded policies were saved as arrays
+      expect(all_policies.flatten.uniq).to eq [Array]
```
aelkiss, line 87: This was confusing to me, probably because `all_policies` has the _class_ of each policy rather than the actual policy. 

The intention might be clearer with this:

```ruby
all_policies = Cluster.each.map do |cluster|
  cluster.commitments.map(&:policies)
end

expect(all_policies.flatten).to all ( be_a(Array) )
```

aelkiss (changes requested):
I would for sure like to use `$INPUT_LINE_NUMBER` instead of `$.`; up to you whether you think my suggestion for checking that policies are arrays is better or not.

spec/clusterable/commitment_spec.rb
```diff
@@ -122,7 +122,11 @@
   end
 
   describe "policies" do
-    ["blo", "digitizeondemand", "non-circ"].each do |policy|
+    it "accepts empty policy" do
```
aelkiss, line 125: Good catch.

lib/utils/tsv_reader.rb
mwarin: Thought I could get away with the occasional perlism but I will concede.

spec/loader/shared_print_loader_spec.rb
mwarin: Yeah that's better and less confusing. Implemented.

aelkiss (approved):

PR #248, opened 2023-01-10
DEV-464, a validator for phase 3 sp commitments

mwarin:
Implementing the phase 3 commitments validator that DEV-464 implies.
`SharedPrint::Phase3Validator` can take a file of commitments and load the ones that meet the criteria set forth in `pass_validation?`.

### DIFF
--- a/lib/phctl.rb
+++ b/lib/phctl.rb
@@ -87,6 +87,11 @@ module PHCTL
     def deprecate(*infile)
       run_job(Jobs::SharedPrintOps::Deprecate, options[:verbose], [*infile])
     end
+
+    desc "phase3load INFILE", "Load Phase 3 commitments, if valid, from file"
+    def phase3load(infile)
+      run_common_job(SharedPrint::Phase3Validator, options, infile)
+    end
   end
 
   class Report < JobCommand
--- a/lib/shared_print/phase_3_validator.rb
+++ b/lib/shared_print/phase_3_validator.rb
@@ -1,0 +1,132 @@
+# frozen_string_literal: true
+
+require "cluster"
+require "loader/shared_print_loader"
+require "services"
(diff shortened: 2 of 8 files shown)


aelkiss:
Regarding naming & positioning --- I think it might make sense to remove the `run` method here and perhaps to allow `SharedPrintLoader` to take an object as a callback to do validation on the loaded commitment? Then we could kind of cleanly extract the validation logic from the loading part. In such a case I think it's fine to just keep this class under `SharedPrint`. I'm not sure about the `Phase3` part. If it's completely clear and obvious to everyone involved in the project what it means to be a `Phase3` commitment then it's probably fine. I know for me personally if I look at this at some future point in time that I'm going to have a hard time remembering, and I wonder if might be better to name it something like `RareUncommittedValidator` by way of parity with the report that generates these?

aelkiss:
Or looking at it again, maybe thinking about splitting it into a basic shared print commitment validator - i.e. what we'd expect any commitments at any phase to have - vs. the phase 3 specific `require_phase_3_policies` part. Then perhaps that could just be named `require_blo_or_non_circ` or something (not that I can remember what `blo` is.._)

aelkiss:
I'm confused because I did review this - maybe not in the right place - but @mwarin I think you were going to make at least some changes based on the feedback? So feel free to request re-review once that's done.

mwarin:
I think you only made some comments, but didn't actually review it in the github sense of the word. 
I think I got your verbal go-ahead just now to merge it as is, and to do the cleanup on a rainy day. 
I put the cleanup task in https://hathitrust.atlassian.net/browse/DEV-571 .
Please confirm.

aelkiss:
Works for me. Please feel free to merge when ready.

PR #252, opened 2023-01-27
Added policy validation rule from DEV-465

mwarin:
Per https://hathitrust.atlassian.net/browse/DEV-465 we should reject a commitment if it has both `NON-REPRO` & `DIGITIZEONDEMAND`, as these are mutually exclusive.

This goes both for adding new commitments as for updating existing ones.

### DIFF
--- a/lib/shared_print/update_record.rb
+++ b/lib/shared_print/update_record.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "clusterable/commitment"
+
 module SharedPrint
   # An individual update record. Consists of find-fields and update-fields
   class UpdateRecord
@@ -54,7 +56,13 @@ module SharedPrint
       when :bool
         {"true" => true, "false" => false}[val]
       when :array
-        val.split(",").map(&:strip)
+        split_vals = val.split(",").map(&:strip)
+        if key == :policies
+          if Clusterable::Commitment.incompatible_policies?(split_vals)
+            raise ArgumentError, "Policies contain mutually exclusive commitments."
+          end
+        end
+        split_vals
       else
         raise ArgumentError, "Did not know how to cast #{key} (= #{val})"
       end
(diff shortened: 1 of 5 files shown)


mwarin:
Didn't see that part of the requirements on the ticket were in the comments, so this is NOT done yet, converting to draft and removing reviewer.

lib/shared_print/update_record.rb
```diff
@@ -54,7 +56,13 @@ def cast(key, val)
       when :bool
         {"true" => true, "false" => false}[val]
       when :array
-        val.split(",").map(&:strip)
+        split_vals = val.split(",").map(&:strip)
+        if key == :policies
+          if Clusterable::Commitment.incompatible_policies?(split_vals)
```
aelkiss, line 61: There might be something to extract here about commonalities between update & load validation, but maybe that's only incidental? Like, is this something that should *always* be true about *any* commitment, or is it something particular to phase 3? If the former, maybe it makes sense to put this on `Commitment` itself rather than here.. ? In any case, if in the future we find ourselves making parallel updates here and in `phase_3_validator` we can rethink.

spec/shared_print/phase_3_validator_spec.rb
```diff
@@ -49,6 +49,19 @@
...
+      expect { p3v.require_compatible_policies(spc) }.to_not raise_error
+      # add non-repro, should still be good
+      spc.policies = ["non-repro"]
+      expect { p3v.require_compatible_policies(spc) }.to_not raise_error
+      # add blo, should still be good
+      spc.policies = ["non-repro", "blo"]
+      expect { p3v.require_compatible_policies(spc) }.to_not raise_error
+      # add digitizeondemand should blow it up
+      spc.policies = ["non-repro", "blo", "digitizeondemand"]
+      expect { p3v.require_compatible_policies(spc) }.to raise_error SharedPrint::Phase3Error
```
aelkiss, line 63: I wonder if this could or should be somewhat more generic like `SharedPrint::InvalidCommitmentError` -- is there a reason to tie it to Phase 3?

aelkiss (approved):
I think there might be something to extract here about things that are *always* supposed to be true about shared print commitments vs. things that are only true about commitments from phase 3 at the time we load them vs. things that should be true about phase 3 commitments at any time - but I don't know if those requirements are clear enough right now or the existing data is consistent enough to enforce those. For now let's merge, but (as with most things) if we find ourselves making changes in that area it's likely ripe for a little refactoring.

spec/shared_print/phase_3_validator_spec.rb
mwarin: Yeah, that could be made more generic.

PR #253, opened 2023-01-31
Before scrubbing file, make sure it has no illegal utf8 stuff

mwarin:
New class `Utils::Encoding` for checking if a file contains illegal utf8 sequences.
`Scrub::AutoScrub` uses it to check incoming files before subjecting them to further analysis.
`Utils::Encoding` has the ability to force_utf8 and remove illegal sequences (but that's not in use yet, if ever).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ARG UID=1000
 ARG GID=1000
 
 RUN apt-get update -yqq && apt-get install -yqq --no-install-recommends \
-  nodejs netcat rclone less entr
+  nodejs netcat rclone less entr uchardet
 
 WORKDIR /usr/src/app
 ENV BUNDLE_PATH /gems
--- a/lib/scrub/autoscrub.rb
+++ b/lib/scrub/autoscrub.rb
@@ -4,6 +4,7 @@ require "services"
 require "scrub/member_holding_file"
 require "scrub/scrub_output_structure"
 require "utils/line_counter"
+require "utils/encoding"
 
 module Scrub
   # Usage:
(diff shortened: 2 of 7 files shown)


aelkiss:
I'm interested in @moseshll 's thoughts on the best way to deal with this. Right now, if we encounter utf-16 with no byte order marker and no characters outside ASCII, it won't provoke an error -- it will just appear to have null characters in between every byte of text. For example, try:

```
echo "this is a test" | iconv -f utf8 -t utf16le | hexdump
echo "this is a test" | iconv -f utf8 -t utf16le | iconv -f utf8 -t utf8 | hexdump
```

No errors, no difference - just a bunch of extra NULLs.

My inclination as to order of operations would be:

1. check for byte order marker for utf-8, utf16le, utf16be & interpret according to that encoding if one is found (perhaps warn if it isn't utf-8)
2. strip any null characters (warn if any are found)
3. if it still can't be handled, error

aelkiss:
For reference @moseshll  we got an input file from Université de Montréal encoded as UTF-16LE (with byte order marker) (which I believe is the default character set used by the Windows API). @mwarin were there any other apparent different encodings that we got?

mwarin:
@aelkiss I don't really know what other encodings we are/been getting. Still working on umontreal.

What prompted me to get back into this was that the mpm file looks like utf8 except some enum_chrons with strangely encoded accented e:s, see `spec/fixtures/non_valid_utf8.txt`. I have just enough encoding rope to do something bad with, so any input from you or @moseshll is much appreciated.

aelkiss (commented):
I guess the question at this point is if/where we want to use `force_utf8` - but everything that's here seems fine to me currently.

PR #254, opened 2023-02-01
DEV-586: Use push metrics gem

aelkiss:
Interface to using push metrics changes slightly esp. since we want to make sure we can inject/override parameters for registry & pushgateway via Services.

See also https://github.com/hathitrust/push_metrics/pull/2, https://github.com/hathitrust/hathifiles/pull/30


mwarin (approved):
Looks fine, runs fine. 
I'm only confused about where `ppnum` went, which is still used in 5 places.

aelkiss:
`ppnum` is included as a top-level helper in milemarker. Probably not the best, but better than defining it twice...

aelkiss:
I'm confused why coveralls is complaining about new missed lines in the bin scripts, but I don't think anything in the specs should have changed that. I will merge once CI passes.

PR #255, opened 2023-02-03
MarcXML parser for Exlibris Alma holdings output

mwarin:
Using this branch to work out how to deal with Exlibris's Alma MarcXML files.

See e.g. https://hathitrust.atlassian.net/browse/DEV-415, https://hathitrust.atlassian.net/browse/DEV-691 , https://hathitrust.atlassian.net/browse/DEV-692, but perhaps most of all DEV-833


bin/ex_libris_holdings_xml_parser.rb
```diff
@@ -0,0 +1,105 @@
+require 'nokogiri'
+
+class ExLibrisHoldingsXmlParser
+  def main(files)
+    yield HTRecord.header_tsv
+    files.each do |file|
+      File.open(file) do |io|
+        Nokogiri::XML(io).xpath("//record").each do |xml_record|
```
aelkiss, line 8: Definitely fine as a first pass, but if we do this for real we should use rubymarc - if nothing else that parses the XML in streaming mode rather than trying to load the whole file into memory

bin/ex_libris_holdings_xml_parser.rb
```diff
@@ -0,0 +1,105 @@
...
+  end
+
+  def item_type
+    @item_type ||= map_item_type(path_text "datafield[@tag='ITM']/subfield[@code='m']")
+  end
+
+  # Todo: Figure out the structure of enum_chrons.
+  def enum_chron
+    @enum_chron ||= [
+      path_text("datafield[@tag='ITM']/subfield[@code='a']"),
```
aelkiss, line 60: with rubymarc we can do basically `record.fields("ITM").first["a"]` - should be easy enough just to use that.

mwarin:
Failing because this branch needs to rebase or merge in the wait-for/netcat fixes.

mwarin:
> Failing because this branch needs to rebase or merge in the wait-for/netcat fixes.

Rebasing

aelkiss:
I handled some of the item type stuff so we can use the output more directly.

On the file I have only 5 records appear to be govdocs with the given logic (out of 965,749 holdings records), so we'll definitely want to check that.

aelkiss:
Although I don't think we're doing anything with govdoc information right now anyway in holdings, so worst case maybe we can just omit that column.

aelkiss:
I also fixed the govdoc parsing.

@mwarin I figure before merging we should probably have some tests -- anything else you think we should do here?

mwarin:
This work continues in DEV-415-exlibris-sample2. Closing this PR.

PR #256, opened 2023-02-08
added retention_condition field to commitments

mwarin:
As per https://hathitrust.atlassian.net/browse/DEV-595, Phase 3 commitments require a retention_condition value of "ACCEPTABLE" or "EXCELLENT".


aelkiss (approved):
Looks good -- tests look fine; adds validation in the places I'd expect, doesn't have to change or remove anything else for this to work, etc.

PR #257, opened 2023-02-08
make sure log is uploaded even when error has been raised

mwarin:
Simplified error handling and made it so that any error caught by `ScrubRunner.run_file` results in the log file being uploaded to the member.


aelkiss (approved):
nice simple change & clear test.

PR #259, opened 2023-02-13
DEV-632: changed scrub logging to prevent seepage, changed timestamp format on scrub log filename

mwarin:
Scrub logs were seeping into each other. 

Uploaded log files had a different timestamp format than we require from the input files, so I harmonized that to all be YYYYMMDD (no dashes) so as to be pleasurable to the eye while I was at it.

### DIFF
--- a/lib/scrub/autoscrub.rb
+++ b/lib/scrub/autoscrub.rb
@@ -22,6 +22,15 @@ module Scrub
     def initialize(path)
       @path = path
 
+      # @member_id and @item_type are used in the path to scrub_logger, but we
+      # also need somewhere to log to before we know @member_id and @item_type
+      scrub_dir = File.dirname(@path)
+      early_scrub_logger_path = File.join(scrub_dir, "scrub.log")
+      Services.register(:scrub_logger) do
+        Logger.new(early_scrub_logger_path)
+      end
+      Services.scrub_logger.info("Getting basic info from #{@path}")
+
       @member_holding_file = Scrub::MemberHoldingFile.new(@path)
       @member_id = @member_holding_file.member_id
       @output_struct = Scrub::ScrubOutputStructure.new(@member_id)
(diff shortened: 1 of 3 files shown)


aelkiss (commented):
I'd consider `Date.strftime("%Y%m%d")` instead of the `delete('-')` - IMO that's more clear as to what will happen / what the format is

lib/scrub/autoscrub.rb
```diff
@@ -22,6 +22,15 @@ class AutoScrub
     def initialize(path)
       @path = path
 
+      # @member_id and @item_type are used in the path to scrub_logger, but we
+      # also need somewhere to log to before we know @member_id and @item_type
```
aelkiss, line 26: What is the purpose of the early logging? If it's primarily for our own debugging, could it only go to STDOUT and we look at it via loki, or are there reasons to want it in a file? (I know there are many unknowns with loki, plus it only currently retains logs for I think 14 days, but I'd still like to understand what we end up doing with this 'early' log file)

lib/scrub/autoscrub.rb
```diff
@@ -30,14 +39,17 @@ def initialize(path)
...
-      # build a log path and re-register the service logger to log to that path
-      ymd = Time.new.strftime("%F")
+      # build a new log path and re-register the service logger to that path
+      ymd = Time.new.strftime("%F").delete("-")
       @logger_path = File.join(@log_dir, "#{@member_id}_#{@item_type}_#{ymd}.log")
       Services.logger.info "autoscrub logging to #{@logger_path}"
 
       Services.register(:scrub_logger) do
+        # Get the early log in there.
+        FileUtils.mv(early_scrub_logger_path, @logger_path)
```
aelkiss, line 52: Well, I guess I answered my own question - the "early" stuff ends up getting moved to the "real" log once we know its name

aelkiss:
I'm not following from the PR what exactly was happening to cause the seepage - could you give a concrete example here of what was happening?

mwarin:
I did describe the seepage issue in the associated ticket (DEV-632), so I didn't feel I needed to be more explicit in the pull request, but in retrospect I should at least have said "see the problem as detailed in ticket [link]" or something.

mwarin:
> I'd consider `Date.strftime("%Y%m%d")` instead of the `delete('-')` - IMO that's more clear as to what will happen / what the format is

Total brainfart. Yes, of course.

PR #261, opened 2023-02-17
DEV-639: ignore UTF8 BOM in holdings files

mwarin:
The file header line validation was not OK with there being non-printable characters in there. 
This change makes the validation ignore UTF-8 BOMs.


aelkiss (changes requested):
It looks like there is a mechanism to automatically read the BOM or fall back to UTF-8:

https://ruby-doc.org/core-2.5.0/IO.html#method-c-new-label-IO+Encoding

```ruby
File.open("whatever.txt","r:BOM|UTF-8")
```

This should handle utf-16 without much rigmarole as well.

aelkiss:
It looks like `Zinzout` should honor an "encoding" parameter - https://github.com/billdueber/zinzout/blob/master/lib/zinzout.rb - but worth verifying that it works.

mwarin:
Can confirm that it is indeed possible to pass `encoding:"BOM|UTF-8"` via `ZinZout.zin` all the way to the underlying `IO`.

billdueber (approved):
All looks straightforward to me. TIL about `BOM|UTF-8`. Makes me wonder if I should put that as the default encoding for Zinzout, although apparently `StringIO` objects just ignore the BOM and you have to do some messing around. But so long as you're working with file-like objects this should be solid.

aelkiss (approved):
Looks good to me now.

aelkiss:
Superceded by #263

mwarin:
All important ideas here were incorporated into pr 263

PR #262, opened 2023-02-17
handle lines with only \r for line breaks

mwarin:
Some holdings files we see only have `\r` for a line break (old mac style), which needs to be read differently or the whole file looks like a one liner. 

This workaround consists of reading the file with `file.gets("\r")` and writing those lines to a temporary file with proper line breaks, and then overwriting the old file with the new file.

### DIFF
--- a/lib/scrub/autoscrub.rb
+++ b/lib/scrub/autoscrub.rb
@@ -3,6 +3,7 @@
 require "services"
 require "scrub/member_holding_file"
 require "scrub/scrub_output_structure"
+require "securerandom"
 require "utils/line_counter"
 require "utils/encoding"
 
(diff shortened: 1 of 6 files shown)


aelkiss:
How much have you looked into other methods for dealing with this? There might be a way to use "universal newlines" mode: see the bottom of https://ruby-doc.org/3.2.1/encodings_rdoc.html . There is also the `$INPUT_RECORD_SEPARATOR` special global a.k.a. `$/` but as far as I can tell unlike in perl there's no way to scope that locally, so I would be hesitant to use that.

mwarin:
My first impulse was to do it the Perl way, messing with record separators, but that seemed messy and not very Ruby-y.

My impression of the universal_newline is that that's a string encoding method, and I need to get the string out of the file one line at a time before I can do that, and so I'm back at square one, no? Hence `file.gets("\r")`.

If there is a cleverer method I have not perceived it.

aelkiss:
I can't find a good way to deal with this that also takes into account that files might be gzipped.

```ruby
File.open("file_with_mac_style_newlines.txt",universal_newline: true).readlines
```  works

```
Zlib::GzipReader.open("file_with_mac_style_newlines.txt.gz",universal_newline: true).readlines
``` doesn't work as desired, because `Zlib::GzipReader` (which `Zinzout` uses) doesn't honor the `universal_newline` option -- `Zlib::GzipReader` doesn't actually subclass `IO`.

**Comment by aelkiss** [comment]
So, I see a few options here:

* gunzip the file before reading it
* use `IO::popen` à la 

```ruby
IO.popen("zcat foo_old_mac_nl.txt.gz",universal_newline: true, encoding: "BOM|UTF-8")
```

or 

* rewrite the file like this

I personally would lean towards the `IO.popen`. I guess that isn't too different from shelling out to `tr`, although shelling out regardless means taht we need to deal with the maybe-gzippedness. I will see what I can come up with.

mwarin:
All important ideas here were incorporated into pr 263.

aelkiss:
Superceded by #263

PR #264, opened 2023-02-22
DEV-406: Use temporary directories for test

aelkiss:
- creates a temporary directory just for that test; puts it in ENV["TEST_TMP"]. This means we can avoid cleanup in each test

- reloads settings to point to that tmpdir; reloading settings before each test also means we don't have to worry about saving/restoring settings in individual specs

- Enable disabled autoscrub JSON spec

### DIFF
--- a/config/environments/test.yml
+++ b/config/environments/test.yml
@@ -1,27 +1,27 @@
-concordance_path: /tmp/concordance
-cost_report_freq_path: /tmp/cost_report_freq
-cost_report_path: /tmp/cost_reports
-deprecation_report_path: /tmp/deprecation_report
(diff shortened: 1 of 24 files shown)


aelkiss:
Generally speaking this means that tests will run in a more isolated way. There were a few specs that needed to change to ensure that each test ran in isolation.

mwarin (approved):
Looks good. Did a `watch -n 0.5 -d ls -lt /tmp/` while tests were running and could observe the temporary dirs popping in and out of existence. Very relaxing! 

Can also verify that the temp dirs do get cleaned up even if the test suite is broken (e.g. by inserting a `raise "oopsie"` in a random test.

PR #266, opened 2023-02-24
DEV-569: Adding `Thread.pass` to get rid of Sidekiq warnings when loading holdings

mwarin:
As far as I can tell, `Loader::HoldingLoader.item_from_line` is _the_ thing being called once for each line in the input file being loaded. If `Thread.pass` needs to be sprinkled in more places, I'm not seeing it.

### DIFF
--- a/lib/clustering/cluster_holding.rb
+++ b/lib/clustering/cluster_holding.rb
@@ -64,6 +64,7 @@ module Clustering
         c.holdings
           .select { |h| h.organization == org && h.date_received < date }
           .map { |h| ClusterHolding.new(h).delete }
+        Thread.pass
       end
     end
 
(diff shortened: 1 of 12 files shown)


aelkiss:
hmm.. `Thread.pass` is already being called in `lib/loader/file_loader.rb` after each batch is loaded. I wonder if we should pass there https://github.com/hathitrust/holdings-backend/blob/main/lib/loader/file_loader.rb#L64 in addition to after loading each batch? Like, how frequently do these things need to yield? I would have thought once per batch would be sufficient.

aelkiss:
@mwarin  See what you think or if you want to refactor some of this. Either way, I'm OK with whatever - I _think_ it should be harmless (in general we're only running sidekiq workers with one thread anyway, so this is just yielding back to the 'control' thread rather than something that would mess with running work) but I guess we can see - it doesn't seem to have caused an issue for the loaders

mwarin (commented):
Well it's my own pull request so GitHub doesn't let me formally approve.
Let it however be known, throughout all the lands, that I hereby approve this PR.

PR #271, opened 2023-04-17
PHO-20: adding report for basic holdings based overlaps

mwarin:
New report script that is based on one `organization`'s holdings, where we list data from the member holdings and add 1-3 different overlaps:
* `ph` for the count of members with `holdings` in the same `Cluster`
* `sp` for the count of (non-deprecated) `commitments` in the same `Cluster`
* `htdl` for the count of `ht_items` in the same `Cluster`

The `ph`, `sp` and `htdl` all default to `nil`, and one or more must be set to non-`nil` for the report to run.

Example:
```
$ phctl report organization-holdings-overlap --organization=ucmerced --ph --sp --htdl
```
... will generate a report with all the holdings by `ucmerced` along with their `ph`, `sp` and `htdl` overlap counts.


lib/reports/overlap_report.rb
```diff
@@ -0,0 +1,149 @@
...
+    def initialize(organization: nil, ph: false, htdl: false, sp: false)
+      @organization = organization
+      @ph = ph
+      @htdl = htdl
+      @sp = sp
+      validate!
+    end
+
+    # Check inputs
+    def validate!
```
aelkiss, line 22: I'm always leery of any code that checks a specific class in Ruby. What problems would happen if any of these were not literal `true` or `false`? I'd lean towards just doing

```ruby
def validate!
   unless [@ph, @htdl, @sp].any?
      raise "At least one of ph/htdl/sp must be true"
   end
end
```

What I suspect might be happening is concern about sidekiq & serialization, and worries about getting the string "true" or "false". I think that would be a good thing to verify in an integration test -- but not necessarily the responsibility of this class to check.

lib/reports/overlap_report.rb
```diff
@@ -0,0 +1,149 @@
...
+          cluster_sp = count_cluster_sp(cluster)
+          cluster.holdings.select { |h| h.organization == @organization }.each do |hol|
+            outf.puts row(hol, cluster_ph, cluster_htdl, cluster_sp).join("\t")
+          end
+        end
+      end
+    end
+
+    # Get output location
+    def outf_path
```
aelkiss, line 70: I think we have an open ticket already to abstract this kind of thing; might now be the time to fix it rather than duplicating the code for generating output file paths again?

lib/reports/overlap_report.rb
```diff
@@ -0,0 +1,149 @@
...
+      ]
+      output << cluster_ph if @ph
+      output << cluster_htdl if @htdl
+      output << cluster_sp if @sp
+
+      output
+    end
+
+    # If @ph=true, count how many orgs have holdings in the cluster
+    def count_cluster_ph(cluster)
```
aelkiss, line 118: I kind of wonder if each of these (ph, htdl, sp) should have their own little classes; then you instantiate whichever ones you want based on the provided options, and this becomes kind of a wrapper class that can include several of these kinds of counts? I'm not sure what exactly that might look like, just noticing the repetition several times in this class of the variation of ph/htdl/sp

aelkiss (commented):
I'm a little bit leery of the defensive coding in `OverlapReport#validate!` which I suspect might better be covered by an integration test checking that Sidekiq will round-trip booleans via its job json encoding.

The rest is maybe more of a philosophical question to think about how much refactoring is worth doing ahead of or behind things like this -- I might be more inclined now to take the time to extract out the report path generation (DEV-407); I would likely not spend the time now to extract out the different columns each as a separate class, but it might be worth thinking about doing before adding a fourth potential column -- right now the overlap report isn't really open for extension, so if in the future we need to modify it, first we should make it open to extension...

lib/reports/overlap_report.rb
mwarin: Changed the phctl params to strings and then just checking for non-`nil`-edness.

PR #274, opened 2023-05-11
Pho 22

mwarin:
A holdings report. 
List the last time (which year) an org got holdings loaded, grouped by org and mono_multi_serial.


moseshll (approved):
Looks good, although for my version of Docker I had to merge the two `volumes` sections under the `mongo_dev` service, may be worth changing to avoid errors.
```
volumes:
      - data_db:/data/db
      - ./bin:/tmp/bin
```

PR #275, opened 2023-05-12
DEV-550: Report newly ingested volumes for Shared Print purposes

mwarin:
The main payload here is the report script (`lib/reports/shared_print_newly_ingested.rb`) which will read ingested volumes (and their dates) from a file if given one, otherwise it will try connecting to mysql and run a query to get that data. Then based on that look up clusters in mongodb and present data accordingly.

The rest is plumbing and testing.


lib/reports/shared_print_newly_ingested.rb
```diff
@@ -0,0 +1,172 @@
...
+      if @ht_item_ids_file.nil?
+        :matching_item_ids_from_db
+      else
+        :matching_item_ids_from_file
+      end
+    end
+
+    # Get item_id and "min_time" from mysql, these are the volumes that
+    # were ingested or got their rights assigned since @start_date.
+    def matching_item_ids_from_db
```
aelkiss, line 138: Perhaps worth nothing that this method introduces a new dependency on the rights database and also is not tested (because instead of reading from the DB you can also just get this from a file). I _think_ this is fine to leave for now, but we may want to add a comment that it should be a target for getting via an API or something in the future...

aelkiss (commented):
Looks good; tests look comprehensive; I like being able to provide a file with the ingest dates rather than relying solely on the rights database. I would like to see a somewhat more descriptive comment for `Reports::SharedPrintNewlyIngested`.

lib/reports/shared_print_newly_ingested.rb
mwarin: It's not a brand new dependency, as in we already have production code that uses the rights db in similar ways. But yes, I'll put a sentence there to that effect.

moseshll: \<sniffle\> Everyone forgets poor old https://github.com/hathitrust/rights_database

(Probably not worth trying to shoehorn it in now, but worth a TODO)

moseshll (approved):

PR #276, opened 2023-05-26
dealing with holdings on large clusters

mwarin:
PR to understand/prevent/fix the issue of super bloated holdings on large clusters.

Contains some new tests and a bin script that forces the specified clusters to behave as `LargeClusters` (i.e. max one holding per org).


moseshll (approved):
FYI I had only one problem with this (and I think maybe I have seen this before so it's not related to the PR and should not be an impediment):
The `mongo_dev` service in `docker-compose.yml` has two `volumes` entries, on my Macs this causes `docker-compose build` to gripe
```
parsing /Users/moseshll/hathitrust/holdings-backend/docker-compose.yml: yaml: unmarshal errors:
  line 41: mapping key "volumes" already defined at line 38
```
The easy fix was just to combine the two `volumes` entries and tally-ho.

PR #277, opened 2023-06-05
Dev-550: (bugfix) Making sure variable is used properly in query

mwarin:
Diff, essentially:

```
- w
+ W
```
That is to say, what has functionally changed is a non-interpolating `%w[...]` was changed into an interpolating `%W[...]`. The changed code was broken out into its own method (`query`) so I could write a simple test to make it obvious that interpolation happened.

### DIFF
--- a/lib/reports/shared_print_newly_ingested.rb
+++ b/lib/reports/shared_print_newly_ingested.rb
@@ -142,20 +142,24 @@ module Reports
       return enum_for(:matching_item_ids_from_db) unless block_given?
       db = Services.holdings_db
       db.run("USE ht")
-      query = %w[
+      db.fetch(query, @start_date) do |row|
+        yield row[:ht_id], row[:min_time]
+      end
+    end
+
+    # The query string, if going via matching_item_ids_from_db
+    def query
+      <<~HEREDOC
         SELECT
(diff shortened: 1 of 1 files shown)


aelkiss (changes requested):
A couple thoughts: while this appears correct, it's an unusual construction (I don't know that I've ever seen `%W` before) and is insecure - given that `start_date` is passed in from the outside this is in theory vulnerable to sql injection. Best practice is to use a SQL placeholder for start_date. If we want a multi-line string I would prefer using a heredoc (mariadb shouldn't care about the extra whitespace):

```ruby
QUERY = <<~SQL
        SELECT
        CONCAT(namespace, '.', id) AS ht_id,
        MIN(DATE_FORMAT(time, "%Y-%m-%d")) AS min_time
        FROM
        rights_log
        GROUP BY
        namespace, id
        HAVING
        MIN(time) >= ?
SQL
```
along with:

```ruby
  db.fetch(QUERY,@start_date) do |row|
    ...
  end
```

 (see https://www.rubydoc.info/gems/sequel/4.38.0/Sequel%2FDatabase:fetch)

aelkiss (approved):
Looks good now - thanks

moseshll (approved):
Three cheers for here docs.

PR #278, opened 2023-06-09
DEV-779: weeding decision report

mwarin:
This pull request adds the new Weeding Decision report to the reports that can be started via `phctl`.

The Weeding Decision Report is mainly for use by shared print members, so this is something they would request from us to run and upload on their behalf.

The main query is based on a member's holdings, restricted to clusters with (1+ open) items and (1+ non-deprecated) commitments, and the output records show numbers based on those clusters.

The report script itself is `lib/reports/weeding_decision.rb`, with tests in `spec/reports/weeding_decision_spec.rb` and `spec/phctl_spec.rb`.

The changes to `lib/phctl.rb` and `lib/sidekiq_jobs.rb` make the report possible to run from inside a running container using:

```
bash phctl.sh report weeding-decision <org>
```


moseshll (approved):
Looks good. I thought I could DRY out WeedingRecord a bit but that's easier said than done.

PR #281, opened 2023-08-21
Dev-906: in Print Holdings cost report, make items w/ rights:icus count towards the public domain items.

mwarin:
Items with `rights:icus` used to be counted towards the IC part of the Print Holding cost report (meaning only the members who report holding ocns matching those items are charged for them). 

This change makes them count towards the PD part (meaning all members are charged a fraction of the cost).

Most of the lines changed have to do with the getting the tests into shameless green: changing them so that they specify rights when building ht_items (otherwise a test item built with e.g. `access:deny` could be assigned `rights:ic` as likely as `rights:icus` and that's not right).

### DIFF
--- a/lib/reports/cost_report.rb
+++ b/lib/reports/cost_report.rb
@@ -3,6 +3,10 @@
 require "overlap/ht_item_overlap"
 require "services"
 
+# In Aug 2023 we decided that items with rights:icus should behave as if
+# they were access:allow (PD, everybody pay), instead of the traditional
+# access:deny (IC, holders pay).
+
 module Reports
   # Generates reports based on h_share
   class CostReport
@@ -81,11 +85,27 @@ module Reports
       @num_pd_volumes ||= Cluster.collection.aggregate(
         [
           {"$match": {"ht_items.0": {"$exists": 1}}},
-          {"$group": {_id: nil,
-                      items_count: {"$sum": {"$size": {
-                        "$filter": {input: "$ht_items", as:    "item",
-                                    cond: {"$eq": ["$$item.access", "allow"]}}
-                      }}}}}
+          {
+            "$group": {
+              _id: nil,
+              items_count: {
+                "$sum": {
+                  "$size": {
+                    "$filter": {
+                      input: "$ht_items",
+                      as: "item",
+                      cond: {
+                        "$or": [
+                          {"$eq": ["$$item.access", "allow"]},
+                          {"$eq": ["$$item.rights", "icus"]}
+                        ]
+                      }
+                    }
+                  }
+                }
+              }
+            }
+          }
         ]
       ).first[:items_count]
     end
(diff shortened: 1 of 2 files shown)


spec/reports/cost_report_spec.rb
```diff
@@ -32,13 +38,40 @@
...
+  describe "making sure that access and rights come out the way they go in" do
+    it "pd == allow" do
+      cluster_tap_save [build(:ht_item, access: alo, rights: pd, ocns: [111])]
+      cluster = Cluster.find_by(ocns: 111)
+      expect(cluster.ht_items.count).to eq 1
+      expect(cluster.ht_items.first.rights).to eq pd
+      expect(cluster.ht_items.first.access).to eq alo
+    end
+    it "icus == allow" do
+      cluster_tap_save [build(:ht_item, access: alo, rights: icus, ocns: [222])]
```
aelkiss, line 50: This would not happen with the current logic in the hathifiles; that said the behavior here is what we would expect in the holdings system

lib/reports/cost_report.rb
```diff
@@ -145,14 +159,22 @@ def add_ht_item_to_freq_table(ht_item)
     end
 
     def matching_clusters
+      # We don't apply the icus rule here, since we're getting clusters not items.
+      # Any items gotten from these clusters need to be checked for icus though.
       if @organization.nil?
-        Cluster.where("ht_items.0": {"$exists": 1},
-          "ht_items.access": "deny").no_timeout
+        Cluster.where(
```
aelkiss, line 165: appears to be only a formatting change rather than a logic change in `matching_clusters`, so presumably existing tests cover this?

lib/reports/cost_report.rb
```diff
@@ -81,11 +85,21 @@ def num_pd_volumes
...
+          {"$group": {
+            _id: nil,
+            items_count: {"$sum": {"$size": {
+              "$filter": {
+                input: "$ht_items",
+                as: "item",
+                cond: {
+                  "$or": [
+                    {"$eq": ["$$item.access", "allow"]},
+                    {"$eq": ["$$item.rights", "icus"]}
```
aelkiss, line 97: is there a test that fails without this change?

aelkiss (commented):
The logic in `CostReport` appears as I expected. I see one test I was expecting to see (that icus is included in the pd part of the fee) but it wasn't clear if one of the tests ensures that `icus` is *not* counted in the in-copyright part of member fees.

lib/reports/cost_report.rb
mwarin: Covered, according to coveralls: https://coveralls.io/builds/62122116/source?filename=lib%2Freports%2Fcost_report.rb#L170

See also: https://github.com/hathitrust/holdings-backend/blob/782a666ab9c252005367e7cc09e0741055f1fbe4/spec/reports/cost_report_spec.rb#L126

lib/reports/cost_report.rb
mwarin: No, good point.

spec/reports/cost_report_spec.rb
mwarin: Right, and the `ht_items` we get from `build(:ht_item, ...)` (the factorybot-defined /spec/factories/ht_item.rb) set rights and access independently from each other. If I just set `rights:icus` it's a coin toss whether that comes back `access:allow` or `access:deny`.

lib/reports/cost_report.rb
mwarin: Wait, I'm still not addressing the thing in your feedback, am I.

mwarin: Broke out `ic_volumes` so I can actually count them, in order to address this bit of feedback.

PR #288, opened 2023-09-18
DRAFT PR for Dev 905

mwarin:
https://hathitrust.atlassian.net/browse/DEV-905

This is not necessarily a fix for something broken... I just felt like I had dangerously low levels of understanding how the splitting/merging of clusters actually happen, since I didn't write that code and haven't worked with it.

So in this PR I am focusing on writing tests around splitting and merging as an exercise for myself, in order to gain understanding, and see if anything can be done along the way.


mwarin:
I'm going to open this for review, in case @aelkiss or @liseli are interested.

Generally though, I think this falls under "does not need code review" since it's going from green to green, and only really changing tests. Thoughts?

liseli (approved):
@mwarin your changes make sense for me and it seems you are not impacting the current logic. I will approve it

aelkiss (approved):
The tests make sense to me. As far as the observed issue on SPO-50 it sounds like this is primarily an issue of how items are matched to commitments. From the tests I am convinced that:

* commitments move between clusters appropriately when reclustering
* changes to item data can affect eligibility and matching for commitments, and that we understand the ways in which that happens.

In reading SPO-50 I *think* I'm convinced that these tests describe the behavior that happened and we sufficiently understand things. Assuming you agree I'm good with merging this.

PR #289, opened 2023-09-19
DEV-952: Add report platform to Print Holdings/Shared Print to simplify generating basic reports

mwarin:
## Context

There has been a need for making report generation easier in Print Holdings / Shared Print. There are currently a few tickets for "generate a shared print report for x" that I intend to solve using this new code. And just to be clear, this PR does not add any new reports, just a "platform" on which new reports can (if they are simple enough) rapidly be built and run with little new code needed for each new report.

This work started here on this PR, and a ticket was added later: https://hathitrust.atlassian.net/browse/DEV-952

Since this is all new code, it doesn't have much in terms of complicated history/dependencies, which is nice.

In the Print Holdings / Shared Print system, the main organizing unit is a `Cluster`. A typical `Cluster` has a couple of ocns, and any number of `Clusterable`s. A `Clusterable` is just a module with no class definition of its own, but organizes the classes `Commitment`, `Holding`, `HtItem` or `OCNResolution` which are the main information-carrying units that the `Cluster` is made up of. Those, then, are also the level at which most queries are made.

## Solution

This PR adds the class `Reports::Dynamic`. 

This class knows which `Clusterable`s exist and how to read their field definitions. It lets you specify a query on a class in the `Clusterable` module and use its fields to restrict the record set, as well as to determine which fields to display.

So, for instance, you can get the ocns of all colby's withdrawn holdings:
```ruby
Reports::Dynamic.new(
  decorations: ["holdings.ocn"], # ~ SELECT X
  base: "holdings", # ~ FROM Y
  restrictions: {"holdings.status" => "WD", "holdings.organization" => "colby"} # ~ WHERE Z1 AND Z2 
).run
```

That code will take each `Clusterable::Holding` that matches `restrictions`, and output their `decorations` (with header) to a `.tsv`-file.

The other addition is a `.spec` file with tests for `Reports::Dynamic`.

## Testing

In order to run tests locally:

```
# SETUP
$ cd "$to_wherever_you_wanna_do_this"
$ git clone https://github.com/hathitrust/holdings-backend.git
$ cd holdings-backend
$ git checkout report-platform
$ bash bin/setup/setup_dev.sh
$ docker-compose up -d pushgateway
$ docker-compose run --rm -e MONGOID_ENV=test dev bash
# You are now in a print holdings dev bash session

# To run ALL the tests:
$ bundle exec rspec --fail_fast

# To run just the tests added in this PR:
$ bundle exec rspec spec/reports/dynamic_spec.rb

# To start up a `pry` shell with all the classes loaded:
$ bash phctl.sh pry
pry(#<PHCTL::PHCTL>)> Reports::Dynamic.new(query_params_go_here).run
pry(#<PHCTL::PHCTL>)> # ...
pry(#<PHCTL::PHCTL>)> exit

# when you are done testing:
$ exit
$ docker-compose down 
```

## Limitations

Since I did not want to reinvent SQL, there are no higher-order functions. The only axis of comparison is equality, and the only logical operator is AND (which is implied when you pass multiple restrictions). So there's only really one kind of query you can perform and it's `SELECT x FROM y WHERE z;`. If you want to do more complicated stuff, do it with something else, or on the output from this object.

## Urgency

There is at least one critical report that I want to run before the end of October 2023, so this cannot wait very long.


lib/reports/dynamic.rb
```diff
@@ -0,0 +1,266 @@
...
+
+    # Ensure that the model class, e.g. Clusterable::XY, has the field Z
+    def validate_model_field(model_class, field)
+      unless Dynamic.model[model_class].fields.key?(field)
+        raise ArgumentError, "Model #{model_class} does not have field #{field}"
+      end
+    end
+
+    # Given a string, figure out model_class. ("x.y"->"x", "y"->@base)
+    def get_model_class(str)
```
aelkiss, line 160: Not covered in tests, doesn't appear to be used

lib/reports/dynamic.rb
```diff
@@ -0,0 +1,266 @@
...
+        # Add a truth value to array.
+        matches << (doc_field_val == restriction.val)
+      end
+      # For this embedded_doc, as we checked all restrictions,
+      # did we add only trues? If so, the embedded_doc is a match.
+      matches.all?(true)
+    end
+
+    # Ensure that the model class, e.g. Clusterable::XY, has the field Z
+    def validate_model_field(model_class, field)
```
aelkiss, line 153: Not covered in tests, doesn't appear to be used

lib/reports/dynamic.rb
```diff
@@ -0,0 +1,266 @@
...
+      if /\./.match?(str)
+        validate_clusterable(str.split(".").first)
+      else
+        @base
+      end
+    end
+
+    # Given a string, figure out field. ("x.y"->"y", "y"->"y")
+    def get_field(str)
+      if /\./.match?(str)
```
aelkiss, line 170: Under what conditions would this be used? This branch isn't covered by the tests

lib/reports/dynamic.rb
```diff
@@ -0,0 +1,266 @@
...
+
+        on, key = r.flatten.first.split(".")
+        val = r.flatten.last
+        restriction_objects << Reports::Restriction.new(on: on, key: key, val: val)
+      end
+      restriction_objects
+    end
+
+    # Puts it all together.
+    def run
```
aelkiss, line 89: Doesn't look like this is currently covered, and I would think we would want an integration test that puts this all together

aelkiss (commented):
Thanks for the detailed overview, this all makes sense & the design looks good. Most of the comments are based on looking at the code coverage from https://coveralls.io/builds/63176886/source?filename=lib%2Freports%2Fdynamic.rb - I think there is a fair amount here that either needs to be tested or should be removed.

lib/reports/dynamic.rb
mwarin: Correct. Removed.

lib/reports/dynamic.rb
mwarin: Correct. The functionality is in `QueryElement.validate!`, this was vestigial and is now removed.

lib/reports/dynamic.rb
mwarin: It *is* used. Once. Made it private.

See `doc_field_val = embedded_doc.send(get_field(restriction.key))`
... which sort of evaluates to:
`doc_field_val = commitment.organization`
... or whatever `clusterable.field` is currently being sought.

lib/reports/dynamic.rb
mwarin: Agree, that remains.

mwarin: Right. I have a test in the test suite that says it tests `run`, but then it actually only tests `records` which is a hefty part of `run`, but isn't === `run`.

mwarin: Covered now.

PR #286, opened 2023-09-08
DEV-573: Fix broken phase lookup for Shared Print reporting purposes

mwarin:
Problem (https://hathitrust.atlassian.net/browse/DEV-573): 

We are currently using `committed_date` to determine which "phase" a commitment belongs to. Meaning, when we want to e.g. look up commitments in phase 3, we have to look up "what's the date associated with phase 3" and query on that committed_date. This, in combination with sloppy date formatting and servers in different timezones, has led to issues with timezone confusion.

Proposed solution: 

* We add an integer field called `phase` to `lib/clusterable/commitment.rb`. 
* Keep `committed_date`, but don't use it as a proxy for phase.
* Write and run a piece of single-use code that can add correct `phase` values to commitments, based on date.
* Then make any code that loads commitments require that the user specifies phase.
* Add a mongo index for `commitments.phase`.

Steps toward solution:

* Enabling `bin/inspect_ocn.rb` to look up multiple ocns, and output them in a format that `bash phctl.sh load cluster-file` can load, for testing purposes.
* Added test suite fixture file for commitments (copied from production using the fix to `bin/inspect_ocn.rb`).
* Added `phase` as a field to `Clusterable::Commitment` and a corresponding index.
* Simplified `SharedPrint::Finder` and updated `SharedPrint::Phases`.
* Added a script to get all distinct `committed_date` from production: `lib/shared_print/select_distinct_phases.rb`
* Added a script to set phase on commitments that have a specific committed_date: `lib/shared_print/phase_updater.rb`

---

Guide to files changed:

`bin/inspect_ocn.rb`:                                                                                                                                                                            
Enable lookup of multiple ocns.
Enable output of multiple clusters.
This makes it easier to copy clusters from production and loading them in dev/test (`phctl load cluster-file`)

`lib/clusterable/commitment.rb`:
Added phase field and validation.

`lib/shared_print/finder.rb`:
Enable lookup by phase.
Simplify some of the stuff that was using committed\_date.

`lib/shared_print/phase_3_validator.rb`
Sets `phase` and `committed_date` to appropriate values for phase 3.
Broke apart `run` into `load_all` & `load_one` so I could test loading more directly.

`lib/shared_print/phase_updater.rb`:
Script to set phase on commitments based on their committed\_date.

`lib/shared_print/select_distinct_phases.rb`:
In order to set phase based on committed\_date, we must first have a list of all the committed\_date

`lib/shared_print/phases.rb`:
Added phase constants.

[...]

PR #291, opened 2023-10-10
A script to identify invalid clusters in Print Holdings MongoDB

mwarin:
## Overview

What is it? A script to identify the clusters in the Print Holdings MongoDB that are invalid. Clusters are the main organizing unit in this database. I would like to run this regularly, perhaps together with the db backup.

Why is it? An invalid cluster doesn't let you save any information on it, until you make it valid. We currently don't have good insight into this, and that could become a problem. It may in fact already be a problem and we just don't know. I'm also going to be doing stuff to the schema, which is one of the ways you can end up with invalid clusters, so that's why now.

How big is it? The diff size may look big, but when you break it down it's:

* 1 file of code: `bin/cluster_validator.rb`
* 1 file of tests: `spec/cluster_validator_spec.rb`
* 2 fixture files: `spec/fixtures/single_cluster_invalid.json` & `spec/fixtures/single_cluster_valid.json`

The fixture files are identical except for a single field (`commitment.phase`). The valid file has a 1 (valid phase), the invalid 999 (invalid phase). Why is this? Well, it's hard to make existing clusters invalid by just manipulating them, because the existing validations stop you from setting invalid values. However, data that are loaded directly into the db bypass these validations. The spec file uses the cluster-file loader to get valid/invalid clusters loaded, so it can perform tests on invalid clusters. 

Instead of putting this in https://github.com/hathitrust/holdings-backend/pull/286 (which is already large) I broke this out into its own PR.

## What kind of feedback am I looking for?

Just a general look at the validator and it's spec. It's all new code, so there isn't any tangle.

## Testing

The tests run fine on github but if you want to run them yourself it is:

1) git clone this repo
2) git checkout this branch
3) Spin it up and enter its shell:
```
bash bin/setup/setup_dev.sh
docker-compose up -d pushgateway
docker-compose run --rm -e MONGOID_ENV=test dev bash
```
4) `$ bundle exec rspec spec/cluster_validator_spec.rb`
5) `$ exit`
6) `$ docker-compose down`

### DIFF
--- a/bin/cluster_validator.rb
+++ b/bin/cluster_validator.rb
@@ -1,0 +1,35 @@
+require "cluster"
+require "services"
+Services.mongo!
+
+# Goes through all clusters, checks if they are valid,
+# and prints the first ocn of any invalid cluster to a file.
+
+class ClusterValidator
+  attr_reader :output_path # file name
+  def initialize
+    # Make an output file in the right place
+    ymd = Time.now.strftime("%Y-%m-%d")
+    dir = Settings.local_report_path
+    FileUtils.mkdir_p(dir)
+    @output_path = "#{dir}/cluster_validator_#{ymd}.txt"
+  end
+
+  def run
+    puts "Writing to #{output_path}"
+    File.open(output_path, "w") do |outf|
+      # Go through each cluster and check if valid.
+      outf.puts "# These are ocns of invalid clusters:"
+      Cluster.each do |c|
+        unless c.valid?
+          outf.puts(c.ocns.first)
+        end
+      end
+      outf.puts "# Done"
+    end
+  end
+end
+
+if __FILE__ == $0
+  ClusterValidator.new.run
+end
--- a/spec/cluster_validator_spec.rb
+++ b/spec/cluster_validator_spec.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "loader/cluster_loader"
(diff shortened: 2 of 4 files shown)


bin/cluster_validator.rb
```diff
@@ -0,0 +1,51 @@
...
+  def run
+    puts "Writing to #{output_path}"
+    @outf = File.open(output_path, "w")
+    # Go through each cluster and check if valid.
+    comment "These are ocns of invalid clusters:"
+    Cluster.each do |c|
+      unless c.valid?
+        @buffer << c.ocns.first
+      end
+      if @buffer.size > @buffer_max
```
aelkiss, line 29: Curious what motivated the buffer (i.e. did you try it and find that outputting every time was really slowing things down?) -- which is to say, I would be inclined to remove the buffering unless you have specific evidence it provides a benefit

aelkiss (commented):
Makes sense; all looks pretty straightforward. Main question is if the buffering is necessary or not.

bin/cluster_validator.rb
mwarin: Yeah this is one of those defensive moves arising from the fact that I have no idea how this scales from 10 clusters in dev to the many many clusters in prod. I'm guessing it's not saving us a ton, so it can go away in the name of simplicity.

mwarin: Buffering removed.

PR #292, opened 2023-10-12
Benchmark and improve performance for `Reports::Dynamic`

mwarin:
Gather here all performace fixes to related to the reporting platform, which was introduced in:
https://github.com/hathitrust/holdings-backend/pull/289

After putting into production and running against the production MongoDB, I could tell that this was going to need some tuning. It is slow and takes up an awful lot of resources. I can think of 3 approaches:

1) Benchmark ruby code and see what can be easily improved there.
2) Add indexes to MongoDB where missing.
3) See what can be parallelized without too much fussing.

In production, I would like the following basic query to take less than 20m. 
It's currently at 3h+.

```ruby
dyna_rep = Reports::Dynamic.new(
  base: "holdings",
  decorations: ["holdings.ocn", "holdings.local_id"],
  restrictions: {"holdings.organization" => "ucmerced"},
)
dyna_rep.run
```

41575


aelkiss:
@mwarin Any reason to keep this around at this point or should we close the PR?

mwarin:
While once potentially relevant, this no longer is. Closing.

PR #293, opened 2023-10-17
Fix broken fix of PhaseUpdater in Shared Print

mwarin:
## What?

This is yet another PR for DEV-573.

* Added `module` block to `lib/shared_print/phase_updater.rb`
* `@date` variable is now parsed as a `DateTime` before being used in query
* Added `spec/shared_print/phase_updater_spec.rb` with one big test

## Why?

The previous PR for DEV-573 (https://github.com/hathitrust/holdings-backend/pull/286) missed a few things, and did not have the desired effect. The test I'm adding here would have showed me what was missing, but I thought I could get away with not writing it. This is a teaching moment.

## For the reviewer

I'm just looking for a quick eyeball.

Turn on "hide whitespace" in the "Files changed" view, or it's going to look like the entire file `lib/shared_print/phase_updater.rb` is a diff.

Like so:
image

### DIFF
--- a/lib/shared_print/phase_updater.rb
+++ b/lib/shared_print/phase_updater.rb
@@ -1,50 +1,52 @@
 # frozen_string_literal: true
 
+require "date"
 require "mongo_updater"
 
 # This is an outer wrapper for a MongoUpdater call.
 # Objective: based on commitments.committed_date, set commitments.phase.
 # Usage: bundle exec ruby get_by_date.rb <date_str> <phase>
 # E.g. : bundle exec ruby get_by_date.rb "2023-01-31 00:00:00 UTC" 3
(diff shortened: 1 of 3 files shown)


liseli:
Did you create the module to organize better your code? It is just a question to understand better the change. It looks good to me, so I will approve it.

carylwyatt (approved):
I'm _not good_ at OOP, but this is small enough that I can see what's going on. Looks good, I think?

carylwyatt:
@mwarin I did wonder about that, but thought maybe the tests were in a different repo or something (which doesn't make sense now that i'm thinking that through) 😅 I'll ask next time!

moseshll (approved):
Looks good.

PR #294, opened 2023-10-19
DEV-562, report count of commitments by phase & org

mwarin:
## What? 

Originating ticket: [DEV-562](https://hathitrust.atlassian.net/browse/DEV-562)

This PR is primarily adding a new report class (`Reports::SharedPrintPhaseCount`), that gives a count of how many `commitment`s per `organization` given a `phase`.

So, if umich has 100 phase 3 commitments and smu has 50, and we make a phase 3 report using this new class, we expect an output file with:

```
organization  |phase  |commitment_count
smu           |3      |50
umich         |3      |100
```
(pipes and spacing added here for human readability)

Also taking this time to add a module for simplifying report output location (`Utils::ReportOutput`), which is starting to bug me.
If successful/useful we can (on a rainy day) start implementing in the other report scripts.

## How?

If you want to run the tests:

```bash
git clone https://github.com/hathitrust/holdings-backend.git
cd holdings-backend
git fetch
git checkout DEV-562
# Start bash sesh.
bash bin/setup/setup_dev.sh
docker-compose up -d pushgateway
# Start up a shell with a profile loaded
docker-compose run --rm -e MONGOID_ENV=test dev bash

# you are now in a holdings-backend shell
# run the tests
bundle exec rspec spec/reports/shared_print_phase_count_spec.rb spec/utils/report_output_spec.rb

# run a report (which will be empty unless you load some data):
bash phctl.sh report shared-print-phase-count --inline

# When bash sesh is over: do cleanup.
docker-compose down; yes

```
## Why?

The reporting script is simply needed.

The util script is not strictly needed, but it allows any given script that produces output files to not have to figure out where to put those files, how to name them, or know/care much about the directory structure. This also standardizes timestamps in filenames, and automatically adds a short unique-enough-ish string to avoid filename collisions.

## Files changed

4 all new files:
* `lib/reports/shared_print_phase_count.rb` : report script, the main payload of this PR
* `spec/reports/shared_print_phase_count_spec.rb` test for report script
* `lib/utils/report_output.rb` util class to potentially simplify report outputs
* `spec/utils/report_output_spec.rb` test for util class

2 minimally edited files to make report script runnable from `phctl.sh`:
* `lib/phctl.rb`
* `lib/sidekiq_jobs.rb`

## Reviewing/Testing

I'm mostly looking for a sanity check of the tests and the util class.

From @aelkiss I am most interested in what you think about `ReportOutput` as a solution for simplifying reports, since you did the trick for the temporary spec-ouputs.


lib/utils/report_output.rb
```diff
@@ -0,0 +1,63 @@
...
+      if mkdir
+        FileUtils.mkdir_p(dir_path)
+      end
+      @dir
+    end
+
+    def id
+      if @id.nil?
+        timestamp = Time.now.strftime("%Y%m%d_%H%M%S")
+        # Adding a random string in case 2 identical jobs that write a file
```
aelkiss, line 50: How likely is this to happen in practice? Will the extra hex digits cause any extra work or confusion when sending the reports to others?

lib/utils/report_output.rb
```diff
@@ -0,0 +1,63 @@
...
+      if @file.nil?
+        @file = File.join(dir, id) + ext
+      end
+      @file
+    end
+
+    def dir(mkdir: true)
+      # In testing it is useful to get the path without mkdir_p,
+      # hence the mkdir param
+      dir_path = File.join(Settings.local_report_path || "/tmp", report_name)
```
aelkiss, line 37: Could we do `@dir ||= File.join(...)` here?

Worth considering that right now, if you call this once without `mkdir` set, but then call it again later after changing `Settings.local_report_path` and with `mkdir` set, it will return the existing `@dir` but still make `dir_path`. Not sure that's intended.

lib/utils/report_output.rb
```diff
@@ -0,0 +1,63 @@
...
+module Utils
+  class ReportOutput
+    attr_reader :report_name, :ext
+    def initialize(report_name, ext = ".tsv")
+      @report_name = report_name # e.g. "cost_report" or "cost_report_umich"
+      @ext = ext
+    end
+
+    def file
+      if @file.nil?
```
aelkiss, line 28:
```ruby
def file
  @file ||= File.join(dir, id) + extg
end
```

lib/phctl.rb
```diff
@@ -179,6 +179,13 @@ def shared_print_newly_ingested
       run_common_job(Reports::SharedPrintNewlyIngested, options)
     end
 
+    desc "shared-print-phase-count (--phase=x)",
+      "Get tally of commitments per organization in the given phase"
+    option :phase, type: :numeric, default: nil
+    def shared_print_phase_count
+      run_common_job(Reports::SharedPrintPhaseCount, options)
```
aelkiss, line 186: Should add this to `phctl_spec.rb` & consider adding something to `integration_spec.rb`

lib/reports/shared_print_phase_count.rb
```diff
@@ -0,0 +1,50 @@
...
+      @output = Utils::ReportOutput.new("sp_phase#{@phase}_count")
+    end
+
+    def run
+      # Get an output handle.
+      handle = output.handle("w")
+      puts "Started writing to #{output.file}"
+      handle.puts(header)
+      # Get relevant commitments, tally organizations.
+      organization_tally = {}
```
aelkiss, line 24: could maybe do

```ruby
organization_tally = commitments.group_by(:organization).map_values { |v| v.count }
```

lib/utils/report_output.rb
aelkiss: Actually in looking at the test I would recommend removing the `mkdir` parameter and doing

```ruby
def dir()
  @dir ||= File.join(Settings.local_report_path || "/tmp", report_name).tap do |d|
    FileUtils.mkdir_p(dir_path)
  end
end
```

See separate comment below about the test

lib/utils/report_output.rb
mwarin: I have for sure already been bitten by this, where I have 2 competing (and possibly pathological) sidekiq jobs that are writing to the same file. Is it a bit of belt and suspenders? Perhaps. I'm keeping it until such a time when we start regretting it, easy enough to undo.

lib/utils/report_output.rb
mwarin: Yeah. Broke out dir into 2 methods and got rid of mkdir param. Close to what you suggested but not quite.

lib/phctl.rb
mwarin: Yep and yep.

spec/utils/report_output_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  it "has working attr_readers" do
+    expect(report_output.report_name).to eq report_name
+    expect(report_output.ext).to eq ext
+  end
+  it "has a default ext: .tsv" do
+    expect(described_class.new(report_name).ext).to eq default_ext
+  end
+  it "makes dir if missing" do
+    # Just get the path with mkdir: false
+    dir_path = report_output.dir(mkdir: false)
```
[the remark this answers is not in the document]
mwarin: Yes, sorta.

lib/utils/report_output.rb
mwarin: Yep.

mwarin:
@aelkiss I think I addressed all of your feedback. Thanks for catching the missing tests!

PR #295, opened 2023-10-25
bugfix for DEV-562: make SharedPrint::Finder phase queries more specific

mwarin:
Further fix for: https://hathitrust.atlassian.net/browse/DEV-562

`SharedPrint::Finder.new(phase: [x]).commitments` should only yield commitments with `phase: x`, but the released code was missing a condition and was erroneously yielding _all commitments in a cluster_ if any one of the commitments in the cluster had `phase: x`.

This PR adds that missing condition and adds the tests to prove it's doing the right thing.

### DIFF
--- a/spec/shared_print/finder_spec.rb
+++ b/spec/shared_print/finder_spec.rb
@@ -23,6 +23,18 @@ RSpec.describe SharedPrint::Finder do
   let(:pd2) { SharedPrint::Phases::PHASE_2_DATE }
   let(:pd3) { SharedPrint::Phases::PHASE_3_DATE }
 
+  def phase_in_phase_out(input_phases)
+    if input_phases.empty?
+      # Edge case:
+      # SharedPrint::Finder.new(phase: []).commitments
+      # ... actually returns commitments with all phases.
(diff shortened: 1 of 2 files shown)


spec/shared_print/finder_spec.rb
```diff
@@ -59,6 +71,25 @@
...
+      # you get only commitments with phase 3
+      cluster_tap_save(
+        build(:commitment, ocn: ocn1, phase: 1),
+        build(:commitment, ocn: ocn1, phase: 2),
+        build(:commitment, ocn: ocn1, phase: 3)
+      )
+      # Look up commitments using a phase array.
+      # Expect the phases of all commitments in the output
+      # to match the phases of the input.
+      phase_in_phase_out([1])
```
moseshll, line 85: I think I would be tempted to make an uber-array `[[1], [2], [3], ...]` and then do an `.each` on it to run the individual subtests. Seems a bit more DRY-ish.

moseshll (approved):
Alas, I was not able to run this locally because Mongo was acting up. Just the one suggestion.

PR #299, opened 2024-05-03
DEV-1124: holdings are equal if fields are equal or both empty

aelkiss:
This should address the underlying issue we were seeing with duplicate holdings not being cleaned up.

It probably would not hurt to add a test case that exercises the specific issue we were seeing -- create a cluster with a holding with some nil fields, update it with a holding with corresponding empty-string fields; ensure that they match and it doesn't end up with duplicate holdings. I will work on that next week, but wanted to get something up for a look now.

### DIFF
--- a/spec/clusterable/holding_spec.rb
+++ b/spec/clusterable/holding_spec.rb
@@ -18,7 +18,7 @@ RSpec.describe Clusterable::Holding do
     expect(c.holdings.first._parent).to be(c)
   end
 
-  it "normalizees enum_chron" do
+  it "normalizes enum_chron" do
     holding = build(:holding, enum_chron: "v.1 Jul 1999")
     expect(holding.n_enum).to eq("1")
     expect(holding.n_chron).to eq("Jul 1999")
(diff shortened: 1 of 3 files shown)


spec/clusterable/holding_spec.rb
```diff
@@ -36,17 +36,42 @@
...
+      h2.n_chron = ""
+      h2.condition = ""
+      expect(h).to eq(h2)
     end
 
     (described_class.fields.keys - ["date_received", "uuid", "_id"]).each do |attr|
       it "== is false if #{attr} doesn't match" do
-        h2[attr] = nil
-        expect(h == h2).to be(false)
+        # ensure attribute in h2 is different from h but
```
aelkiss, line 61: Before, we were always setting `h2[attr]` to nil, so the behavior change made this test fail -- now it sets it to an appropriate non-nil but non-equal type.

mwarin (approved):
Gah, I meant to actually approve not just say "APPROVE".
APPROVE.

aelkiss:
@mwarin I added a test to cover the actual update situation, which revealed that for performance we were using `update_key` (a modified hash value) rather than equality checking in the holdings update. Please take a look and make sure this all still makes sense!

mwarin (approved):
Ran tests (green), looked at code (makes sense), saw new test (good).
APPROVE.

PR #300, opened 2024-05-21
Update dependencies

aelkiss:
* update mongoid to latest mongoid 8 - this required a small fix to ensure that when we want to re-validate a cluster from the database we load the entire thing. Did not update to mongoid 9 - there are significant changes that break a lot of tests, and it's still at 9.0.0 and very new
* update to sidekiq 7
* update to ruby 3.3
* fix test for batches - starts a sidekiq worker in a thread in "embedded" mode and tests that the callback actually runs


aelkiss:
@mwarin Ready for review.

aelkiss:
Still to-do (but not on this branch):
* remove `wait-for`
* remove `Dockerfile.prod`

Gemfile
```diff
@@ -4,27 +4,24 @@ source "https://rubygems.org"
...
 gem "canister"
 gem "dotenv"
 gem "ettin"
 gem "faraday"
 gem "push_metrics", git: "https://github.com/hathitrust/push_metrics.git", tag: "v0.9.1"
 gem "mongo"
-gem "mongoid"
+gem "mongoid", "~> 8.1"
 gem "mysql2"
+gem "pry"
```
aelkiss, line 16: we require pry for `phctl pry` so I figured it should not be in the "development" group only

Gemfile
```diff
@@ -4,27 +4,24 @@ source "https://rubygems.org"
...
 gem "puma"
-gem "sidekiq", "~> 6.0"
-gem "sidekiq-batch"
-
-group :development, :test do
-  gem "pry"
-  gem "pry-byebug"
-end
+gem "sidekiq"
+gem "sidekiq-batch", git: "https://github.com/breamware/sidekiq-batch"
```
aelkiss, line 24: there is no official release on rubygems that works with sidekiq 7. I don't love that (and I want to excise sidekiq eventually) but this seems like a reasonable option for now. `Gemfile.lock` pins this to a specific commit.

Gemfile
```diff
@@ -4,27 +4,24 @@ source "https://rubygems.org"
...
 
+gem "base64"
 gem "canister"
 gem "dotenv"
 gem "ettin"
 gem "faraday"
 gem "push_metrics", git: "https://github.com/hathitrust/push_metrics.git", tag: "v0.9.1"
 gem "mongo"
-gem "mongoid"
+gem "mongoid", "~> 8.1"
```
aelkiss, line 14: mongoid 9.0.0 breaks a lot of things in the tests in somewhat inscrutable ways.

.github/workflows/tests.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         run: docker compose run --rm dev bundle exec standardrb
 
       - name: Run tests
-        run: docker compose run --rm -e MONGOID_ENV=test dev bin/setup/wait-for mariadb:3306 pushgateway:9091 -- bundle exec rspec
+        run: docker compose run --rm -e MONGOID_ENV=test dev bin/setup/wait-for mariadb:3306 pushgateway:9091 redis:6379 -- bundle exec rspec
```
aelkiss, line 22: the batching test relies on redis

aelkiss:
The uncovered line is related to the `ser/spm` branch in `ClusterOverlap`, but the behavior in that branch is the same as for `spm` so I don't think it's an emergency 🤷

lib/phctl.rb
```diff
@@ -20,7 +20,7 @@ def run_job(klass, *args, **kwargs)
 
       def run_common_job(klass, options, *args, **kwargs)
```
mwarin, line 21: You renamed `*args` and `**kwargs` in 2 other places in this PR, but not here. 
I cannot directly sense intention, but it _feels_ like you missed one.

mwarin (commented):
## General

These few items are less about the PR itself and more a list of things I noticed. I'd make issues out of them but I don't see "issues" in this repo.

* README still uses `docker-compose`, should be `docker compose` at this point.
* `setup_dev.sh` and `setup_test.sh` have mongo output that could be turned off (`disableTelemetry()` & `db.disableFreeMonitoring()`) ... just not sure where the smart place to do that is.
* Check https://github.com/hathitrust/holdings-backend/actions/runs/9181288834 for 3 warnings/"annotations"
* Looking at the github action output, I don't need to see 100s of lines of progress bars. Can we turn them off (just in the github action)?

## Actual changes

The functional changes (dockerfiles, gems, spec files) all make sense, no notes.

## Tests

Ran `bash bin/setup/setup_dev.sh`.

Ran tests using the invocation from README (`docker compose run --rm dev bundle exec rspec`) and got 3 errors.

Pruned docker and re-ran using the same command as the github action:

```
docker system prune -af
docker volume prune -af
/bin/setup/setup_test.sh
docker compose run --rm -e MONGOID_ENV=test dev bin/setup/wait-for mariadb:3306 pushgateway:9091 redis:6379 -- bundle exec rspec
```

Either way, I get the same three errors, no matter how I run this:

```
rspec ./spec/clusterable/holding_spec.rb:137 # Clusterable::Holding#inspect shows the uuid
rspec ./spec/shared_print/deprecation_record_spec.rb:172 # SharedPrint::DeprecationRecord#find_commitment shows full commitment data for duplicates, including uuid
rspec ./spec/shared_print/deprecator_spec.rb:56 # SharedPrint::Deprecator satisfies the acceptance criteria
```

I can tell that this test worked fine on github, but I'm still feeling uncomfortable because I was not able to replicate that, and as such I cannot approve.

lib/phctl.rb
aelkiss: Might be a standardrb issue; I'll take a look

aelkiss:
>    README still uses docker-compose, should be docker compose at this point.

Good catch, fixed.

    setup_dev.sh and setup_test.sh have mongo output that could be turned off (disableTelemetry() & db.disableFreeMonitoring()) ... just not sure where the smart place to do that is.

>    Check https://github.com/hathitrust/holdings-backend/actions/runs/9181288834 for 3 warnings/"annotations"

Addressed in #301 (which among other things updates the coveralls action); see https://github.com/hathitrust/holdings-backend/actions/runs/9194998083

> setup_dev.sh and setup_test.sh have mongo output that could be turned off (disableTelemetry() & db.disableFreeMonitoring()) ... just not sure where the smart place to do that is.

Adding `--quiet` to the `mongosh` command in `rs_initiate.sh` seems to suppress the messages without needing to run additional commands.

> Looking at the github action output, I don't need to see 100s of lines of progress bars. Can we turn them off (just in the github action)?

Assuming this is primarily about the docker compose build, I can quiet that relatively easily. Not sure if there are concerns about other outputs in the action as well.

lib/phctl.rb
aelkiss: Yes, I think standardrb kept it in this case because the anonymous use of `*` and `**` only appears to work if they are the individual last argument in a method call. I wasn't familiar with this use and don't particularly love it, but I'm not going to argue with standardrb...

.github/workflows/tests.yml
aelkiss: added in #301

mwarin (approved):
With Aaron's help I was able to observe passing tests, and I rejoice.
I see the feedback addressed, and I rejoice.
I approve.

PR #301, opened 2024-05-22
Docker compose updates

aelkiss:
Uses health checks & multi-stage builds in docker compose

* add base/dev/prod stage rather than Dockerfile vs Dockerfile.prod
* use docker compose health checks instead of wait-for

Would be helpful to review just to make sure the revised Dockerfile and docker-compose.yml still make sense


Ronster2018 (approved):
Over all, the docker related changes make sense to me.

PR #307, opened 2024-06-06
Sort using the default temp location

aelkiss:
Not sure what motivated this originally (maybe running out of space in the default temp location) but sorting in the application working directory won't work now because of the way permissions are set in the application -- in general best practice in containers for security reasons is to make the application working directory read-only by the running application.

### DIFF
--- a/lib/scrub/chunker.rb
+++ b/lib/scrub/chunker.rb
@@ -39,7 +39,6 @@ module Scrub
     def run
       # First we need to get all the data from @glob into one file.
       # `split --number=l/x` only works on files, not on STDIN.
-      # TODO: we might want another dir for sort -T .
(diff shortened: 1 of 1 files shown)


mwarin (commented):
Yes, `-T ./` "sort where you stand" was added back when I was guaranteed to run out `/tmp/` space. 

I'm not sure you want/need my explicit approval here, and I won't give it either, since this is a change that will work or not depending on the larger context, which you understand better than I do. Granted, if you _do_ want/need my approval, you have it.

aelkiss:
Whether or not we run out of space in `/tmp` is probably the same as whether we would in the application working directory. If this fails in the future I have some other things we could try.

PR #308, opened 2024-07-02
Update README.md

moseshll:
From tech staff group meeting, in need of additional comments


aelkiss:
This is no longer relevant, but we should definitely go back and review the documentation in the README for the mariadb setup.

PR #315, opened 2024-12-19
DEV-1423: Prepare for development against mariadb

aelkiss:
mariadb should be our branch for development work at this point.

At this point my goals are to have this be check-outable & runnable with no failing tests (but lots of skipped ones). I think you should be able to do the usual:

```bash
git clone https://github.com/hathitrust/holdings-backend
git checkout -b DEV-1423-remove-mongo
docker compose build
docker compose run test bundle install
docker compose run test
```


aelkiss:
@Ronster2018 I updated the github actions here. I am using a different image name `holdings-mariadb` since I don't want to accidentally stomp on the current mongodb version.

Ronster2018 (approved):

lib/keio_updater.rb
```diff
@@ -1,10 +1,9 @@
 require "cluster"
```
mwarin, line 1: This was more or less a single-use piece of code and can most likely be deleted.

lib/shared_print/phase_3_validator.rb
```diff
@@ -5,7 +5,6 @@
 require "services"
```
mwarin, line 5: The use of this class was rather time-bound. This likely does not need to be re-implemented for mariadb, but then again it is shared print so not anywhere near front of the line anyways.

lib/shared_print/phase_updater.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
```
mwarin, line 1: This class is likely never needed again.

mwarin (approved):
Well, it builds and the tests "work" (since they are all x'ed out). 

I scrolled through the files and made a couple of comments on files that we can likely drop.

Approve.

lib/shared_print/phase_3_validator.rb
aelkiss: There's nothing mongo-specific in this class, so I've kept it for now. I think it's worth getting a better sense of our plans for shared print first.

lib/keio_updater.rb
aelkiss: Done!

lib/shared_print/phase_updater.rb
aelkiss: Removed, since it's mongo-specific (i.e. relies on `MongoUpdater`)

PR #316, opened 2025-01-09
DEV-1456: instantiate ht item from hf table

aelkiss:
* add hathifile_database gem for test so we can create the schema
* test helper to map factorybot attrs to something we can put in hf
* test round-tripping

### DIFF
--- a/lib/clusterable/ht_item.rb
+++ b/lib/clusterable/ht_item.rb
@@ -4,56 +4,47 @@ require "enum_chron"
 require "services"
 
 module Clusterable
-  # An HT Item
-  # - ocns
-  # - item_id
-  # - ht_bib_key
-  # - rights
-  # - bib_fmt
-  # - enum_chron
-  # - n_enum
-  # - n_chron
-  # - collection_code
-  # - billing_entity
   class HtItem
     include EnumChron
-    # field :ocns, type: Array, default: []
-    # field :item_id, type: String
-    # field :ht_bib_key, type: Integer
-    # field :rights, type: String
-    # field :access, type: String
-    # field :bib_fmt, type: String
-    # field :enum_chron, type: String, default: ""
-    # field :n_enum, type: String, default: ""
-    # field :n_chron, type: String, default: ""
-    # field :n_enum_chron, type: String, default: ""
-    # field :collection_code, type: String
-    # field :billing_entity, type: String
 
-    # embedded_in :cluster
-    # validates :item_id, uniqueness: true
-    # validates_presence_of :item_id, :ht_bib_key, :rights, :bib_fmt, :access
+    ACCESSOR_ATTRS = [:ocns, :item_id, :ht_bib_key, :rights, :access, :bib_fmt, :n_enum, :n_chron, :n_enum_chron, :billing_entity, :enum_chron]
+    READER_ATTRS = [:collection_code]
 
-    # validates_each :ocns do |record, attr, value|
-    #   value.each do |ocn|
-    #     record.errors.add attr, "must be an integer" \
-    #       unless (ocn.to_i if /\A[+-]?\d+\Z/.match?(ocn.to_s))
-    #   end
-    # end
+    ALL_ATTRS = ACCESSOR_ATTRS + READER_ATTRS
 
-    def initialize(params = nil)
-      raise "not implemented"
-      super
-      set_billing_entity if collection_code
+    ACCESSOR_ATTRS.each { |attr| attr_accessor attr }
+    READER_ATTRS.each { |attr| attr_reader attr }
+
+    def self.find(item_id:)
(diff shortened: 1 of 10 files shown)


lib/clusterable/ht_item.rb
```diff
@@ -4,56 +4,52 @@
...
-    #       unless (ocn.to_i if /\A[+-]?\d+\Z/.match?(ocn.to_s))
-    #   end
-    # end
-
-    def initialize(params = nil)
-      raise "not implemented"
-      super
-      set_billing_entity if collection_code
+
+    def self.accessor_attrs
```
mwarin, line 10: I'd rather rename this `attr_accessors`, and have a similar pattern for `attr_readers`, for consistency.

```ruby
def self.attr_accessors
  [:ocns, :item_id, :ht_bib_key, :rights, :access, :bib_fmt, :n_enum, :n_chron, :n_enum_chron, :billing_entity, :enum_chron]
end

def self.attr_readers
  [:collection_code]
end

def self.all_attrs
  attr_accessors + attr_readers
end

attr_accessors.each { |attr| attr_accessor attr }
attr_readers.each { |attr| attr_reader attr }
```

lib/clusterable/ht_item.rb
```diff
@@ -4,56 +4,52 @@
...
+    def self.all_attrs
+      accessor_attrs + [:collection_code]
+    end
+
+    accessor_attrs.each { |attr| attr_accessor attr }
+    attr_reader :collection_code
+
+    def self.find(item_id:)
+      row = Services.hathifiles_table.where(htid: item_id).first
+
```
mwarin, line 23: Do we want to raise if number of returned rows != 1? E.g.

```ruby
results = Services.hathifiles_table.where(htid: item_id)
raise KeyError, "Too many hits (#{results.size}) for item_id:'#{item_id}'" if results.size > 1
row = results.first
raise KeyError, "No htitem with key '#{item_id}'" unless row
```

lib/clusterable/ht_item.rb
```diff
@@ -4,56 +4,52 @@
...
 
     def collection_code=(collection_code)
-      super
+      @collection_code = collection_code
       set_billing_entity
     end
 
     def to_hash
-      attributes.with_indifferent_access.except(:_id)
+      self.class.all_attrs.map { |a| [a, send(a)] }.to_h
```
mwarin, line 52: neat

spec/clusterable/ht_item_spec.rb
```diff
@@ -3,77 +3,98 @@
...
 
-  before(:each) do
-    described_class.collection.find.delete_many
-    Cluster.collection.find.delete_many
+  def insert_htitem(htitem)
+    new_htitem_attrs = {
+      htid: htitem.item_id,
+      bib_num: htitem.ht_bib_key,
+      rights_code: htitem.rights,
+      access: htitem.access == "allow",
```
mwarin, line 30: we might want some clever & reusable way of mapping for access values (`allow/deny => 1/0`), or do the effort of making the switch to 1/0 to bring us closer to hf (if hf is as close to a gold standard as we'll ever have)

mwarin (approved):
I had some suggestions re e.g. `self.accessor_attrs`, but no showstoppers.

lib/clusterable/ht_item.rb
aelkiss: Done (and changed to use constants)

spec/clusterable/ht_item_spec.rb
aelkiss: Well -- ultimately I think we may want to not even use this and rather defer to the rights directly (recall we had discussed changing https://github.com/hathitrust/holdings-backend/blob/main/lib/reports/cost_report.rb#L151 to use the rights directly rather than "allow" / "deny"). If we do keep it in the long run, it would probably make more sense to have it on the model as a boolean rather than the string, and there could be some kind of presenter function for anything that needs to format/export it in a report. I think I'm inclined to keep this as-is for now and keep future opportunities in mind (https://github.com/hathitrust/holdings-backend/issues/317)

lib/clusterable/ht_item.rb
aelkiss: I think we can think of `Services.hathifiles_table` as a sort of API where we are guaranteed to receive 1 or 0 records, since `htid` is a primary key -- if that's not met, that's a problem with the API, not with our code, and I think checking it feels a bit unnecessarily defensive to me.

However, it does look like Sequel has a `single_record!` call that will raise an exception if there is no matching row  -- but it doesn't check if there are other rows. I think it does make the **intention** a little bit more clear than `first`. The documentation (http://sequel.jeremyevans.net/rdoc/classes/Sequel/Dataset.html#method-i-single_record-21) is a bit noncommittal here, but it probably doesn't hurt to try it.

PR #318, opened 2025-01-10
DEV-1457: retrieve ocns in cluster given id

aelkiss:
* add cluster_ocn table
* enable tests in ocn for oclc #
* test: insert data for cluster ocns & ensure Cluster can retrieve

### DIFF
--- a/lib/cluster.rb
+++ b/lib/cluster.rb
@@ -14,55 +14,36 @@ require "cluster_error"
 # - htitems
 # - commitments
 class Cluster
-  # store_in collection: "clusters"
+  attr_reader :ocns
 
-  # # Cluster level stuff:
-  # field :ocns
-  # field :last_modified, type: DateTime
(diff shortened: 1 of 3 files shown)


lib/cluster.rb
```diff
@@ -14,55 +14,36 @@
...
-  # def ht_item(item_id)
-  #   ht_items.to_a.find { |h| h.item_id == item_id }
-  # end
+  attr_reader :ocns
+
+  def self.db
+    Services.holdings_db
+  end
+
+  def initialize(ocns: [])
```
mwarin, line 23: Will we ever want to initialize using a `cluster_id`? 
No, wait... that's what `find` does.
Nevermind.

mwarin (approved):
approved, but see comment on `sql/005_cluster_ocns.sql`

lib/cluster.rb
aelkiss: The cluster does need to know its ID, and that's coming in the next subtask.

PR #319, opened 2025-01-10
DEV-1458: instantiate (but not persist) holdings & fetch associated cluster

aelkiss:
This:

* adds the ability to instantiate `Holding`, basically the same as what we did with `HtItem`. While there may eventually be a superclass to extract here (and for other Clusterables,) I think I want to wait until at least we're handling the OCLC resolution table and the classes support validation & saving.
* adds a support context for populating the table mapping OCNs to cluster IDs
* adds the ability to find a cluster by OCN.

See also comments on individual commits.

### DIFF
--- a/lib/cluster.rb
+++ b/lib/cluster.rb
@@ -14,19 +14,26 @@ require "cluster_error"
 # - htitems
 # - commitments
 class Cluster
-  attr_reader :ocns
+  attr_reader :ocns, :id
 
   def self.db
     Services.holdings_db
   end
 
-  def initialize(ocns: [])
+  def initialize(id: nil, ocns: [])
+    @id = id
     @ocns = ocns
   end
 
   def self.find(id:)
     ocns = db[:cluster_ocns].select(:ocn).where(cluster_id: id).map(:ocn)
-    new(ocns: ocns)
+    new(id: id, ocns: ocns)
+  end
+
+  def self.for_ocns(ocns)
+    db[:cluster_ocns].select(:cluster_id).distinct.where(ocn: ocns).map do |row|
+      find(id: row[:cluster_id])
+    end
   end
 
   def ocn_resolutions
--- a/lib/clusterable/holding.rb
+++ b/lib/clusterable/holding.rb
@@ -56,7 +56,7 @@ module Clusterable
     # field :issn, type: String
 
     def cluster
-      Cluster.for_ocns([ocn])
+      Cluster.for_ocns([ocn]).first
     end
 
     # validates_presence_of :ocn, :organization, :mono_multi_serial, :date_received
--- a/lib/clusterable/holding.rb
+++ b/lib/clusterable/holding.rb
@@ -8,7 +8,36 @@ module Clusterable
   # A holding
   class Holding
     include EnumChron
+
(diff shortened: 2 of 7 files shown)


lib/clusterable/holding.rb
```diff
@@ -8,7 +8,35 @@ module Clusterable
...
+        :uuid,
+        :issn
+      ].freeze
+
+    EQUALITY_EXCLUDED_ATTRS = [:uuid, :date_received].freeze
+    EQUALITY_ATTRS = (ACCESSOR_ATTRS - EQUALITY_EXCLUDED_ATTRS).freeze
+    ALL_ATTRS = ACCESSOR_ATTRS + [:organization]
+
+    ACCESSOR_ATTRS.each { |attr| attr_accessor attr }
+    attr_reader :organization
```
aelkiss, line 36: Need to add `READER_ATTRS` here.

lib/cluster.rb
```diff
@@ -14,19 +14,26 @@
...
     @ocns = ocns
   end
 
   def self.find(id:)
     ocns = db[:cluster_ocns].select(:ocn).where(cluster_id: id).map(:ocn)
-    new(ocns: ocns)
+    new(id: id, ocns: ocns)
+  end
+
+  def self.for_ocns(ocns)
```
mwarin, line 33: I found this a little tricky to parse, esp without clarifying comments.
Maybe split it apart into `<getting the rows>` and `<making cluster objects based on found rows>`

```ruby
rows = db[:cluster_ocns].select(:cluster_id).distinct.where(ocn: ocns)
rows.map do |row|
  find(id: row[:cluster_id])
end
```

lib/clusterable/holding.rb
```diff
@@ -8,7 +8,35 @@ module Clusterable
...
+        :mono_multi_serial,
+        :date_received,
+        :country_code,
+        :weight,
+        :uuid,
+        :issn
+      ].freeze
+
+    EQUALITY_EXCLUDED_ATTRS = [:uuid, :date_received].freeze
+    EQUALITY_ATTRS = (ACCESSOR_ATTRS - EQUALITY_EXCLUDED_ATTRS).freeze
```
mwarin: shouldn't `organization` also be in `EQUALITY_ATTRS`?

mwarin (approved):

lib/clusterable/holding.rb
mwarin: you fixed this mid-review, so nevermind

lib/cluster.rb
aelkiss: Let's discuss -- `db[:cluster_ocns].select(:cluster_id).distinct.where(ocn: ocns)` is essentially a glorified SQL statement, and it doesn't execute until you call `map` on it. I also changed these to enumerators here: https://github.com/hathitrust/holdings-backend/commit/328f970d9d876d3ba797b77d8d0f05178e5df746 

I think there will be a fair amount of this kind of stuff, so it's worth finding a pattern we're happy with.

lib/clusterable/holding.rb
aelkiss: It's fixed now, in any case.

PR #321, opened 2025-01-13
DEV-1459: retrieve htitems in cluster

aelkiss:
* Retrieve distinct ht items with a specific set of OCNs
* Support context for adding ht items
* Change functions that return mapped rows to enumerators (to avoid unnecessarily pulling all rows into memory)


mwarin (approved):
looks good, no specific comments

PR #322, opened 2025-01-13
[DEV-1460] started work on holdings, added holdings table, holdings i…

mwarin:
Started work on holdings, added holdings table, holdings in overlap_spec.


aelkiss:
Done in parallel with @mwarin

PR #324, opened 2025-01-15
Dev 1454: ht_item_overlap for spms

mwarin:
Context: https://hathitrust.atlassian.net/browse/DEV-1454

Small changes to:
* `lib/calculate_format.rb`
* `lib/cluster.rb`
* `lib/overlap/ht_item_overlap.rb`
... in order to get `lib/overlap/ht_item_overlap.rb` to run.

Remaining changes to `spec/overlap/ht_item_overlap_spec.rb` consisting either of:
* Re-organizing tests to make it cleaner which ones are enabled/disabled
* Modifying enabled tests so they work
* Disabling tests related to mpms, copying relevant tests and making them about spms
* Adding a few extra tests for good measure

Test:
```
bundle exec rspec spec/overlap/ht_item_overlap_spec.rb
bundle exec rspec
```

### DIFF
--- a/lib/calculate_format.rb
+++ b/lib/calculate_format.rb
@@ -12,7 +12,7 @@ class CalculateFormat
   #
   # @param ht_item, the HT item to calculate the format on.
   def item_format(ht_item)
-    if ht_item.bib_fmt == "SE" || @cluster.large?
+    if ht_item.bib_fmt == "SE"
       "ser"
     elsif cluster_has_item_with_enum_and_same_ht_bib_key? ht_item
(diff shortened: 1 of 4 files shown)


aelkiss (approved):
This all looks good to me. Minimal changes in the code itself; new tests & reorganizations of the tests make sense. I think we should be able to demo this with the same data we have in the dev database already (since we have multiple holdings loaded for that OCN).

Let's go ahead and squash before merging, please, since this is all for DEV-1451.

PR #325, opened 2025-01-16
DEV-1451 - load cluster stuff

aelkiss:
DEV-1451: Load OCLC resolutions
* Create and save clusters
* Find cluster from clusterable
* Create and save OCLC resolution entries (add_oclc_resolutions)
* Update cluster with new OCNs (merge not supported yet)
* Store OCNs as Set in cluster object
* Implement Cluster.each
* Implement Cluster#add_ocn_resolutions
* Enable some ht item loading tests (those about new ht items only)


mwarin:
#324 has been merged

lib/clusterable/ht_item.rb
```diff
@@ -26,16 +26,16 @@ class HtItem
     ACCESSOR_ATTRS.each { |attr| attr_accessor attr }
     READER_ATTRS.each { |attr| attr_reader attr }
 
-    def self.db
+    def self.table
```
mwarin, line 29: i like this change, db -> table. more accurately descriptive, less room for confusion

lib/clusterable/ht_item.rb
```diff
@@ -44,7 +44,20 @@ def self.with_ocns(ocns)
...
     def self.find(item_id:)
-      from_row(db.where(htid: item_id).first!)
+      from_row(table.where(htid: item_id).first!)
+    end
+
+    # Returns the htitem if it is in the listed cluster, otherwise nil.
+    def self.find_in_cluster(item_id:, ocns:)
+      row = table.select { hf.* }
+        .natural_join(:hf_oclc)
+        .where(value: ocns.map(&:to_s), htid: item_id)
```
mwarin, line 54: so this is a `self.find` but with the added constraint that items match the given ocns?
could use a clarifying comment

lib/clusterable/ocn_resolution.rb
```diff
@@ -18,19 +19,50 @@ class OCNResolution
...
+    #
+    def self.table
+      Services[:holdings_db][:oclc_concordance]
+    end
+
+    def self.with_ocns(ocns)
+      return to_enum(__method__, ocns) unless block_given?
+      dataset = table.where(oclc: ocns).or(canonical: ocns)
+
+      dataset.each do |row|
```
mwarin, line 31: here's a chance to rename either the accessors `:deprecated, :resolved` or the table columns `oclc, canonical` so that they are the same, to bring harmony and reduce potential for confusion

lib/clustering/cluster_getter.rb
```diff
@@ -15,10 +15,7 @@ def initialize(ocns)
     end
 
     def get
-      raise "not implemented"
-      Retryable.new.run do
-        try_strategies.tap { |c| yield c if block_given? }
-      end
+      try_strategies.tap { |c| yield c if block_given? }
```
mwarin, line 18: this whole chain of `get -> try_strategies -> find_or_merge` is hard to follow, but i don't have a better solution on hand

spec/clustering/cluster_ht_item_spec.rb
```diff
@@ -2,18 +2,17 @@
 
 require "spec_helper"
 require "clustering/cluster_ht_item"
-RSpec.xdescribe Clustering::ClusterHtItem do
+RSpec.describe Clustering::ClusterHtItem do
   let(:item) { build(:ht_item) }
```
mwarin, line 6: i'd hard-code the ocns for the items where the tests are relying on them being different

spec/factories/ocn_resolution.rb
```diff
@@ -6,6 +6,5 @@
   factory :ocn_resolution, class: "Clusterable::OCNResolution" do
     deprecated { rand(1_000_000) }
```
mwarin, line 7: perhaps worth enforcing that `deprecated != resolved` just for that one-in-a-million time that is bound to happen more often than that out of spite

mwarin (commented):
* still got failing tests. do you have uncommitted changes or is it a merge issue?
* left some comments but no blockers
* approval pending once tests are green

lib/clustering/cluster_getter.rb
aelkiss: Agreed. Worth putting a pin in but maybe not actionable right now.

aelkiss:
I will plan to address some of the comments before merge but we can still try to take a look with @nfulkers today. The failing tests I think were just from a missing file.

lib/clusterable/ht_item.rb
aelkiss: I can also see if there's a common way to extract out the query. It might make more sense to have find take an optional `ocns` parameter.

lib/clusterable/ocn_resolution.rb
aelkiss: I don't love `oclc` and `canonical`; the old (pre-MongoDB) print holdings system called it `variant` and `canonical`, which I like better than `oclc` and `canonical`, but I'm not sure whether I prefer ` 

I think the only thing that relies on this table is:

https://github.com/hathitrust/hathitrust_catalog_indexer/blob/bb84f32c03c56108422363798df76ed996066750/lib/ht_traject/oclc_resolution.rb#L7

so that should be easy enough to change.

The table is also in db_image so we'd need to rename the columns there, but nothing other than catalog indexer is using it.



That's all limited enough impact that I think it's worth doing.

lib/clustering/cluster_getter.rb
aelkiss: This whole thing is **super confusing** and it works for unexpected reasons in the old (MongoDB) version...

lib/clustering/cluster_getter.rb
```diff
@@ -38,7 +35,7 @@ def find_or_merge
       # operation, but don't start a transaction until we know we actually need
       # to merge clusters.
 
-      @clusters = Cluster.for_ocns(@ocns)
+      @clusters = Cluster.for_ocns(@ocns).to_a
       @target = @clusters.shift
 
       return @target if @target.nil? || @clusters.empty?
```
aelkiss, line 41: In MongoDB, `@clusters` is a `Mongoid::Criteria`. I can't even figure out how it implements `shift`, but somehow it appears to. The behavior of `@clusters.shift` appears to be to return the first cluster, but it doesn't actually iterate through the clusters -- i.e. if you call `@clusters.shift` repeatedly, you always get the first one.

Thus, you'll _never_ fall through to the `@clusters.empty?` condition -- if the criteria matches any clusters, then `@target` will be set to the first matching cluster; if it doesn't, then `@target` will be empty.

So, that means if we find **any** clusters, we call `merge`. We skip trying to merge a cluster into itself (`next if source._id == @target._id`), but at the end we add the extra OCNs to the cluster `@target.add_to_set(ocns: @ocns)`. This appears to then save the cluster.

spec/factories/ocn_resolution.rb
aelkiss: Done. I am enforcing now in the factory that the OCNs are positive and that one is a positive amount greater than the other.

lib/clusterable/ocn_resolution.rb
aelkiss: I think we can do this as part of https://hathitrust.atlassian.net/browse/DEV-1448.

lib/clustering/cluster_getter.rb
aelkiss: I rewrote this part in a way I hope is easier to follow. I am also thinking about changing `ClusterGetter.new(@ocns).get` to `ClusterOCNs.new(@ocns).cluster` to better reflect that the purpose of this class is to enforce that these OCNs are together in a cluster.

lib/cluster.rb
```diff
@@ -15,26 +15,66 @@
 # - commitments
 class Cluster
   attr_reader :id
-  attr_accessor :ocns
 
   def self.db
     Services.holdings_db
   end
 
+  def db
```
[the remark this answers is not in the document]
aelkiss: I like it. Done.

lib/clusterable/ht_item.rb
aelkiss: I applied this elsewhere too.

lib/clusterable/ht_item.rb
aelkiss: I tried to improve this; might have gone a bit overboard, let me know if it's too hard to follow now..

spec/clustering/cluster_ht_item_spec.rb
aelkiss: Cleaned up and clarified OCN usage here.

lib/clusterable/ht_item.rb
```diff
@@ -66,6 +85,11 @@ def initialize(params = {})
       end
     end
 
+    def cluster
+      # TODO: raise if there's more than one (i.e. needs merge?)
```
aelkiss, line 89: Should address this TODO before merge

lib/clustering/cluster_getter.rb
```diff
@@ -3,50 +3,59 @@
...
 module Clustering
-  # This class provides a Cluster that contains the OCNs you want. There are
-  # effectively three strategies:
+  # This class provides a Cluster that contains all the given OCNs. Thus,
+  # a call to ClusterGetter.get(@ocns) effectively says: these OCNs belong
+  # together in a cluster.
   #
-  # - Fetch an existing cluster that has the OCNs you want
-  # - Merge multiple clusters together to get a single one with the OCNs you want
+  # TODO: consider renaming to ClusterOCNs.cluster(@ocns)?
```
aelkiss, line 10: Would like to address this TODO before merge

lib/clustering/cluster_ocn_resolution.rb
```diff
@@ -16,9 +16,10 @@ def initialize(*resolutions)
       end
     end
 
+    # TODO make sure this whole thing is in the scope of a transaction incl ClusterGetter.
```
aelkiss, line 19: **must** address this todo before merge

lib/clustering/cluster_getter.rb
aelkiss: @mwarin When you're back I'd like to discuss a couple options regarding naming here:

* `ClusterGetter.new(@ocns).get` (i.e. the same as now)
* `ClusterOCNs.new(@ocns).cluster` (renaming this class)
* `Cluster.cluster_ocns!(@ocns)` (inline into `Cluster`)

Likewise I'd like to discuss whether the names here make sense, especially `find_or_create`.

Another option would be to leave it as it is for now and to discuss/reconsider after we re-implement `#merge`.

lib/clustering/cluster_getter.rb
aelkiss: See comment above: https://github.com/hathitrust/holdings-backend/pull/325#discussion_r1924181218 -- will discuss and see what we want to do.

lib/clustering/cluster_ocn_resolution.rb
aelkiss: As far as adding new unattested OCNs to different clusters in different transactions, the behavior appears to be that the one that runs `_second_` will block; after the first one commits, it will receive a duplicate key error. (If the first transaction rolls back, then the second will just succeed.)

aelkiss: If we add the `ON DUPLICATE KEY UPDATE` clause, however, then the transaction that runs second blocks until the first one commits, but it does go through. This is probably _not_ what we want, and we want to be very sure that we know the previous status of every OCN before we do anything to it.

aelkiss:
@moseshll Added you to take a look for discussion tomorrow, especially regarding `ClusterGetter` and concurrency issues.

aelkiss:
@mwarin Tagged for re-review. I've added quite a bit around _ensuring any of it actually works_

aelkiss:
Tests are passing for me locally but failing in github. Maybe it's pulling a different version of MariaDB? I can experiment and see and pin if needed in the docker-compose to whatever version we have on mysql-htdev currently

aelkiss:
After doing `docker compose pull` locally I get the same error, so I can try pinning to the same version as on mysql-htdev.

lib/clusterable/ht_item.rb
```diff
@@ -26,38 +26,57 @@ class HtItem
     ACCESSOR_ATTRS.each { |attr| attr_accessor attr }
     READER_ATTRS.each { |attr| attr_reader attr }
 
-    def self.db
-      Services.hathifiles_table
-    end
+    class << self
```
mwarin, line 29: I think I prefer having each class method defined as `def self.foo ...`  as opposed to being defined inside a `class << self` block. Just for readability.

spec/clusterable/ht_item_spec.rb
```diff
@@ -60,27 +60,56 @@
...
+      h = build(:ht_item, ocns: [1, 2])
+
+      expect(h.cluster.ocns).to contain_exactly(1, 2)
+    end
+
+    it "returns a cluster with some of its ocns (if there is only one)" do
+      create(:cluster, ocns: [1])
+      h = build(:ht_item, ocns: [1, 2])
+
+      expect(h.cluster.ocns).to contain_exactly(1)
```
mwarin, line 87: I must be confused. I expected `expect(h.cluster.ocns).to contain_exactly(1, 2)` here too.

spec/clustering/cluster_getter_spec.rb
```diff
@@ -2,20 +2,270 @@
...
+    let(:syncer) { Synchronizer.new }
+
+    # We use threading, the Synchronizer, and the InstrumentedClusterGetter to control
+    # the sequence of when various things happen across two different transactions
+    # to provoke different errors.
+    #
+    # Sequel is thread-safe by default, so it will use its connection pool to allocate
+    # a new connection to each thread.
+
+    # case 1:
```
mwarin, line 87: these comments laying out the synchronized tests are super helpful

mwarin (approved):
It makes more sense, and probably will make even more sense if/when `Cluster` incorporates `ClusterGetter`.

Approve.

spec/clusterable/ht_item_spec.rb
aelkiss: At this point the ht item has been initialized but hasn't changed anything in the database. The point here is that `h.cluster` on its own doesn't change anything.

lib/clustering/cluster_getter.rb
aelkiss: We will move to `Cluster.cluster_ocns!` when implementing `merge`.

lib/clustering/cluster_getter.rb
aelkiss: We will move to `Cluster.cluster_ocns!` when implementing `merge`.

aelkiss:
I will squash this a bit and then merge.

aelkiss:
@moseshll Feel free to add comments; I can address while working on `merge`.

lib/clusterable/ht_item.rb
mwarin: ah gotcha

PR #326, opened 2025-01-24
DEV-1485: Merge clusters

aelkiss:
Added merge functionality, and moved `ClusterGetter` to `Cluster.cluster_ocns!`. 

The concurrency tests now have an `InstrumentedCluster` class and a separate `ClusterInstrumentation` class for storing the status in each thread. We:

* instantiate ClusterInstrumentation for each thread
* set this separately in thread-local storage for each thread.
* access the ClusterInstrumentation set in thread local storage in InstrumentedCluster

I am reasonably happy with this approach, I think, and welcome feedback here.


aelkiss:
I've added an approach where we:
* instantiate `ClusterInstrumentation` for each thread
* set this separately in thread-local storage
* access the `ClusterInstrumentation` set in thread local storage in `InstrumentedCluster`

I think this is quite a bit clearer than putting it all in `InstrumentedCluster`.

spec/cluster_spec.rb
```diff
@@ -347,4 +429,285 @@
       end
     end
   end
+
```
aelkiss, line 432: This is all moved from `cluster_getter_spec.rb`

lib/cluster.rb
```diff
@@ -28,6 +28,53 @@ def self.table
     db[:cluster_ocns]
   end
 
+  def self.cluster_ocns!(ocns)
```
aelkiss, line 31: These changes are the meat of what used to be `ClusterGetter`

lib/cluster.rb
```diff
@@ -127,7 +174,20 @@ def add_ocns(ocns)
       table.import([:cluster_id, :ocn], data)
       db.after_commit { @ocns.merge(new_ocns) }
     end
-    @ocns.merge(ocns)
+  end
+
+  # Moves OCLC numbers from other clusters into this one.
```
aelkiss, line 179: This effectively merges clusters

lib/cluster.rb
```diff
@@ -28,6 +28,53 @@ def self.table
...
+        end
+      end
+    end
+
+    # Sets the OCNs in target_cluster to all the OCNs present in all found
+    # clusters (i.e. merges clusters) and adds any additional OCNs from @ocns
+    # not found in any cluster.
+    def update_cluster_ocns(ocns, target_cluster, clusters)
+      # OCNs (from the ones we want) that are in any cluster
+      attested_ocns = clusters.collect(&:ocns).reduce(Set.new, &:merge)
```
aelkiss, line 69: Would `existing_ocns` be clearer @mwarin? I think this was a point of confusion when we looked at `ClusterGetter` before.

lib/clustering/cluster_ht_item.rb
```diff
@@ -22,79 +20,16 @@ def initialize(*ht_items)
...
+    def cluster
+      return if ocns.empty?
 
-      getter.get do |cluster|
-        update_or_add_ht_items(cluster)
-      end
-    end
-
-    # Removes an HTItem
-    def delete
```
aelkiss, line 36: Removed this all for now since we aren't persisting anything about ht items other than OCNs in the same cluster (as the comment in the new stuff indicates)

spec/cluster_spec.rb
```diff
@@ -3,6 +3,88 @@
 require "spec_helper"
 require "cluster"
 
+class Synchronizer
```
aelkiss, line 6: All the changes here reflect stuff that used to be in `ClusterGetter`

spec/cluster_spec.rb
```diff
@@ -347,4 +429,284 @@
...
+    # an HT item update that causes a recluster, and simultaneously an OCN
+    # resolution update the restores the 'glue' - we would want ultimately the
+    # cluster to be merged.
+
+    # Likewise, if one transaction adds something that would cause a merge, and
+    # another transaction does a recluster, we would want the merged thing to
+    # still be there in the end and the cluster to stay together.
+  end
+
+  context "when merging two clusters" do
```
aelkiss, line 643: Note the tests around merge are all enabled now

spec/cluster_spec.rb
```diff
@@ -347,4 +429,284 @@
...
+
+    # case 3:
+    # cluster [1] exists
+    # thread1: [1,2]; thread2: [2]
+    # thread1 find -- gets [1]
+    # thread2 find -- gets nothing
+    # thread2 create - makes [2]
+    # thread1 add_ocns - gets duplicate key error, should retry & merge
+    # thread2 has an outdated cluster id & list of OCNs... but that's probably OK (see below)
+    it "when one transaction tries to create a cluster with two OCNs and another tries to make a cluster with one of those OCNs, retries and merges" do
```
aelkiss, line 535: Note the expectations around this test that results in merging are now enabled

aelkiss (commented):
See comments inline explaining the changes

lib/cluster.rb
```diff
@@ -28,6 +28,53 @@ def self.table
...
+    end
+
+    # separately defined as a method so we can instrument it
+    # in concurrency testing
+    def array_for_ocns(ocns)
+      for_ocns(ocns).to_a
+    end
+
+    def find_or_create(ocns)
+      return unless ocns.any?
```
moseshll, line 51: I believe `return if ocns.none?` is ever so slightly clearer.

moseshll (approved):
I see nothing here that further troubles me. APPROVE

lib/cluster.rb
mwarin: i think this is fine, given the inline comments

spec/cluster_spec.rb
mwarin: My main concern here is that I don't understand `InstrumentedCluster` well enough to tell if this is testing conditions similar to what we expect to see in production. I can be OK with this for now, as long as I get to spend some time with this later.

mwarin (approved):
This seems fine. Happy to see `ClusterGetter` go!

I need to spend some time absorbing the stuff you added to `cluster_spec`, but that's no reason to hold back an approval.

PR #327, opened 2025-01-27
DEV-1490: enable loading holdings

aelkiss:
To be merged after #326

This adds saving holdings at a basic level but doesn't enable updating or deleting holdings.

This also enables some integration specs, including for loading holdings & creating the clustering for htitems (which will help with DEV-1486 / DEV-1487)

### DIFF
--- a/lib/clusterable/holding.rb
+++ b/lib/clusterable/holding.rb
@@ -154,6 +154,12 @@ module Clusterable
       Services.ht_organizations[organization].weight
     end
 
+    def save
+      self.class.table.insert(to_hash)
+    end
+
+    alias_method :save!, :save
+
     private
 
     def blank?(value)
--- a/lib/clusterable/holding.rb
+++ b/lib/clusterable/holding.rb
@@ -22,12 +22,11 @@ module Clusterable
         :condition,
         :gov_doc_flag,
         :mono_multi_serial,
-        :date_received,
         :uuid,
         :issn
       ]
 
-    READER_ATTRS = []
+    READER_ATTRS = [:date_received]
     ALL_ATTRS = ACCESSOR_ATTRS + READER_ATTRS
 
     EQUALITY_EXCLUDED_ATTRS = [:uuid, :date_received]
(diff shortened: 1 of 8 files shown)


aelkiss:
To also consider before merging:

* remove insert_holding and just use create/save directly
* use a common support context for truncating all relevant tables before each test

sql/006_holdings.sql
```diff
@@ -11,7 +11,7 @@ CREATE TABLE `holdings` (
   `n_enum_chron` varchar(255) NULL,
   `status`       enum('CH', 'LM', 'WD') NULL,
   `condition`    enum('BRT', '') NULL,
-  `gov_doc_flag` tinyint(1) DEFAULT '0' NOT NULL,
+  `gov_doc_flag` tinyint(1),
```
aelkiss, line 14: The mongo implementation allows the gov doc flag to be nil, and it doesn't seem worth quibbling over the behavior here -- especially given that we aren't using this flag for anything AFAIK.

moseshll (approved):
No code smells detected.

lib/clusterable/holding.rb
```diff
@@ -22,12 +22,11 @@ class Holding
         :condition,
         :gov_doc_flag,
         :mono_multi_serial,
-        :date_received,
         :uuid,
         :issn
       ]
 
-    READER_ATTRS = []
+    READER_ATTRS = [:date_received]
```
mwarin, line 29: Maybe worth mentioning that `date_received` has its own writer-method defined later. Here it looks like it's just a reader attribute.

sql/006_holdings.sql
mwarin: Yes, you are correct, we need to allow it to be unspecified.
Worth repeating what I said in our Zoom earlier today:
Most holdings don't report govdocness, and it would be incorrect to apply default=0, since a nil value is really more of a "n/a" or a "don't know/care".

mwarin (approved):
All good.

lib/clusterable/holding.rb
aelkiss: I see this as essentially the same as :

```ruby
attr_reader :date_received

def date_received=(date)
   ...
end
```

so I think it's probably fine -- but also fine to add a comment if we find ourselves confused later.

Note `HtItem` also has one of these for collection_code.

PR #328, opened 2025-01-29
DEV-1453: serials overlap; reorganize helpers

aelkiss:
Mostly this reorganizes the test helpers.

The serial overlap specs work without changing the underlying code, after removing the mongoidisms in the specs.


lib/cluster.rb
```diff
@@ -106,7 +106,7 @@ def self.for_ocns(ocns)
     dataset = table
       .select(:cluster_id)
       .distinct
-      .where(ocn: ocns)
+      .where(ocn: ocns.to_a)
```
aelkiss, line 109: Sequel doesn't like it when given a `Set`; might as well coerce to an array rather than making the caller do it.

moseshll (approved):
This seems straightforward. APPROVE (pending appeasement of standardrb)

PR #329, opened 2025-01-29
DEV-1488 ocnless clusters

moseshll:
DEV-1488-related work on OCN-less clusters
- New `OCNLessCluster` class that acts rather like a bib record.
- First crack at FrequencyTable class with freq_spec.rb tests

### DIFF
--- a/lib/calculate_format.rb
+++ b/lib/calculate_format.rb
@@ -41,12 +41,8 @@ class CalculateFormat
   end
 
   def record_has_any_enum?(ht_item)
-    # TODO: FIX with n_enum
-    # Query if catalog record has any n_enum
-    return false
-
     @cluster.ht_items.any? do |ht|
-      ht.ht_bib_key == ht_item.ht_bib_key && !ht.n_enum&.empty?
+      ht.ht_bib_key == ht_item.ht_bib_key && !(ht.n_enum || "").empty?
     end
   end
 end
--- a/lib/calculate_format.rb
+++ b/lib/calculate_format.rb
@@ -42,6 +42,11 @@ class CalculateFormat
 
   def record_has_any_enum?(ht_item)
     @cluster.ht_items.any? do |ht|
+      # TODO: reconsider what happens if n_enum is nil
+      # irb(main):002> !""&.empty?
+      # => false
+      # irb(main):003> !nil&.empty?
+      # => true
       ht.ht_bib_key == ht_item.ht_bib_key && !ht.n_enum&.empty?
     end
   end
--- a/lib/calculate_format.rb
+++ b/lib/calculate_format.rb
@@ -14,7 +14,7 @@ class CalculateFormat
   def item_format(ht_item)
     if ht_item.bib_fmt == "SE"
       "ser"
-    elsif cluster_has_item_with_enum_and_same_ht_bib_key? ht_item
+    elsif record_has_any_enum? ht_item
       "mpm"
     else
       "spm"
@@ -40,7 +40,11 @@ class CalculateFormat
     @cluster.ht_items.any? { |ht| item_format(ht) == format }
   end
 
-  def cluster_has_item_with_enum_and_same_ht_bib_key?(ht_item)
+  def record_has_any_enum?(ht_item)
+    # TODO: FIX with n_enum
+    # Query if catalog record has any n_enum
+    return false
+
     @cluster.ht_items.any? do |ht|
       ht.ht_bib_key == ht_item.ht_bib_key && !ht.n_enum&.empty?
     end
--- a/lib/calculate_format.rb
+++ b/lib/calculate_format.rb
@@ -42,7 +42,7 @@ class CalculateFormat
 
   def record_has_any_enum?(ht_item)
     @cluster.ht_items.any? do |ht|
-      ht.ht_bib_key == ht_item.ht_bib_key && !(ht.n_enum || "").empty?
+      ht.ht_bib_key == ht_item.ht_bib_key && !ht.n_enum&.empty?
     end
   end
 end
--- a/lib/clusterable/ht_item.rb
+++ b/lib/clusterable/ht_item.rb
@@ -9,6 +9,8 @@ module Clusterable
   class HtItem < Clusterable::Base
     include EnumChron
 
+    IC_RIGHTS_CODES = %w(ic op und).freeze
+
     attr_accessor :ocns, :item_id, :ht_bib_key, :rights, :access, :bib_fmt,
       :n_enum, :n_chron, :n_enum_chron, :billing_entity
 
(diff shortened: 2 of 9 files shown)


lib/calculate_format.rb
```diff
@@ -40,7 +40,11 @@ def any_of_format?(format)
     @cluster.ht_items.any? { |ht| item_format(ht) == format }
   end
 
-  def cluster_has_item_with_enum_and_same_ht_bib_key?(ht_item)
+  def record_has_any_enum?(ht_item)
+    # TODO: FIX with n_enum
```
aelkiss, line 44: Can test this after https://github.com/hathitrust/holdings-backend/pull/330 is merged. I think the existing code there after the `return false` should work (although it might not be the most efficient)

lib/clusterable/ht_item.rb
```diff
@@ -39,6 +40,22 @@ def with_ocns(ocns)
         end
       end
 
+      def ic_volumes
+        return to_enum(__method__) unless block_given?
+
+        table.where(Sequel.or(access: "deny", rights_code: "icus")).each do |row|
```
aelkiss, line 46: I think we had said something about enumerating the specific codes we considered for holdings purposes to be ic vs. pd -- just noting this is where that happens now @nfulkers @mwarin

lib/ocnless_cluster.rb
```diff
@@ -0,0 +1,83 @@
...
+# - htitems
+# - commitments
+class OCNLessCluster
+  attr_reader :id
+
+  def initialize(bib_key:)
+    @bib_key = bib_key
+  end
+
+  def ocns
```
aelkiss, line 23: Should just always return an empty array?

spec/ocnless_cluster_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+  end
+
+  describe "#format" do
+    it "has a format" do
+      formats = ["spm", "mpm", "ser"]
+      expect(formats).to include(c1.format)
+    end
+  end
+
+  # Not clear yet how to set up the scaffolding to test the methods we want to keep
```
aelkiss, line 57: I'm not sure here either and we haven't re-enabled these specs for Cluster yet I don't think. I think it would be fine to note this that there may be future work here for the "production overlap table"

spec/reports/freq_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+require "reports/cost_report"
+require "clustering/cluster_holding"
+require "clustering/cluster_ht_item"
+
+RSpec.describe Reports::CostReport do
```
aelkiss, line 8: The PR description says something about `FrequencyTable` but I don't yet see it here -- not yet committed?

lib/ocnless_cluster.rb
moseshll: `Cluster.new.ocns.class` returns Set. Behavior for this class chosen to parallel that of `Cluster` (in this case at least LOL)

spec/reports/freq_spec.rb
moseshll: Do we want FrequencyTable in its own class for the purposes of this branch? My description is intended to point out the repurposable specs that were added.

lib/calculate_format.rb
moseshll: This seems to work and re-enables failing tests; however, (quoting the existing code below) the use of safe dereference on `ht.n_enum` leads to `nil` behaving opposite empty string. Is it expected that there will ever be a nil `n_enum` (and we would not want to bail out with an error)? My knee-jerk is to do something more like `!(ht.n_enum || "").empty?`
```
@cluster.ht_items.any? do |ht|
  ht.ht_bib_key == ht_item.ht_bib_key && !ht.n_enum&.empty?
end
```

lib/ocnless_cluster.rb
moseshll: Done.

lib/calculate_format.rb
aelkiss: n_enum could be nil, and we should verify that it has the same behavior as "".

spec/reports/freq_spec.rb
aelkiss: I don't think we need to do it on this branch, no.

lib/clusterable/ht_item.rb
aelkiss: @moseshll I think what we wrote here is incorrect or at least misleading. Material with rights_code icus is already "deny", and I think in fact we want to find material with access deny but where rights code _isn't_ `icus`.

(If we were to keep `access`, the hathifiles table also uses a tinyint / boolean, so we need to put access: false rather than access: "deny").

I think we should enumerate the specific rights attributes that should be counted as in-copyright for the purposes of the cost report:

* definitely ic, op, und
* not sure about nobody and pd-pvt -- this may be worth checking with Mike on desired behavior -- i.e. maybe nobody should pay for the nobody items, and maybe the pd-pvt items should be counted as public domain

The remaining codes should be pd:

* any `cc-` code
* ic-world
* pd
* pdus
* und-world

As of 2025-01-31 those are the only attested rights codes in the hathifiles. There are a couple others (orph, orphcand) that aren't in use at all, and others that cause things not to show up in hathifiles (supp)



Based on this comment:

https://github.com/hathitrust/holdings-backend/blob/ffa943dd8f9579f086b46887626629091d5699ef/lib/reports/cost_report.rb#L6-L8

and the mongodb query in num_pd_items:
https://github.com/hathitrust/holdings-backend/blob/ffa943dd8f9579f086b46887626629091d5699ef/lib/reports/cost_report.rb#L84-L111

I think we can just query for items where access is deny --

lib/calculate_format.rb
moseshll: I've re-enabled the `CalculateFormat` tests and added a test for the change. Reverting it fails the test.

lib/clusterable/ht_item.rb
moseshll: Reformulated based on list of IC attributes, punted on `nobody` and `pd-pvt`.

lib/calculate_format.rb
moseshll: Removed the test as it was flawed and was not doing what I thought it was. Reverted to original form of `record_has_any_enum?` with a TODO comment.

lib/clusterable/ht_item.rb
aelkiss: By default then `nobody` and `pd-pvt` are allocated to the 'shared' portion of the fees, whereas the current behavior is to have them be individual based on holdings. I want to keep that behavior, but this is fine for now -- we will address it separately in https://github.com/hathitrust/holdings-backend/pull/335

lib/calculate_format.rb
```diff
@@ -40,8 +40,13 @@ def any_of_format?(format)
     @cluster.ht_items.any? { |ht| item_format(ht) == format }
   end
 
-  def cluster_has_item_with_enum_and_same_ht_bib_key?(ht_item)
+  def record_has_any_enum?(ht_item)
     @cluster.ht_items.any? do |ht|
+      # TODO: reconsider what happens if n_enum is nil
```
aelkiss, line 45: I added an issue https://hathitrust.atlassian.net/browse/DEV-1531 to address this more broadly

lib/reports/cost_report.rb
```diff
@@ -73,6 +73,8 @@ def active_members
     end
 
     def num_volumes
+      return Clusterable::HtItem.all_volumes.count
```
aelkiss, line 76: This is (and `num_pd_volumes` below) are very inefficient because it will marshal all rows as objects, then count them as an array. We can address this on https://github.com/hathitrust/holdings-backend/pull/335 though

aelkiss (approved):
I think this is all fine for now in terms of the OCNless cluster part. Some of the things that need changes we'll address as part of https://github.com/hathitrust/holdings-backend/pull/335. Feel free to squash any of this as desired & then merge.

PR #330, opened 2025-01-29
DEV-1449: enumchron handling & mpm overlap

aelkiss:
This will need to be squashed down, but the basic points are:

* un-mongoify the multipart overlap specs
* get the `EnumChron` module working -- in the process, we ended up pulling out a base class for `HtItem` and `Holding`.

We also:

* reorganized the support contexts
* moved aside shared print & other unused specs -- so we can start getting a handle on how close we are to enabling all the tests we need for now without seeing a bunch of skipped tests we aren't planning to enable in the near term


lib/enum_chron.rb
```diff
@@ -1,26 +1,42 @@
 # frozen_string_literal: true
 
 require "enum_chron_parser"
-require "services"
 
 # Handles automatic normalization of enum_chron fields for Holdings and HTItems
 module EnumChron
-  def initialize(params = nil)
```
aelkiss, line 8: I don't know how the initialize here ever worked. Rather, we default the value of the attribute to blank, and that seems to work well enough.

mwarin (approved):
I like how much neater the clusterable classes are now. 
Do we want any tests directly for `Clusterable::Base`? I don't think so?
APPROVE.

moseshll (approved):
This will re-enable a spec in my PR now that enumchron functionality is resurrected. Looks good.

PR #331, opened 2025-01-31
DEV-1489 Overlap Count

moseshll:
- Add `FrequencyTable` class and specs.
- Wire new class into `CostReport`.
- Integrate tests from `spec/reports/freq_spec.rb` into `CostReport` spec.


lib/frequency_table.rb
```diff
@@ -0,0 +1,44 @@
...
+    @table = Hash.new do |hash, member|
+      hash[member] = Hash.new { |fmt_hash, fmt| fmt_hash[fmt] = Hash.new(0) }
+    end
+  end
+
+  def organizations
+    @table.keys.sort
+  end
+
+  def [](organization)
```
moseshll, line 19: I originally had a `def [](org, fmt)` type of accessor but the `CostReport` tests seem to favor getting all the stats for one org.

lib/frequency_table.rb
```diff
@@ -0,0 +1,44 @@
...
+    @table[organization.to_sym]
+  end
+
+  # For testing. I don't know if this is useful in the long run.
+  def to_h
```
moseshll, line 24: This was added just so I could easily integrate this class into `CostReport` specs which were expecting a Hash.

aelkiss (approved):
All this makes sense to me. We can further iterate on the interface to `FrequencyTable` if needed. I think this will be easy to use further in service of parallelizing `CostReport` by building multiple `FrequencyTable`s and then summing them.

PR #332, opened 2025-02-03
Dev 1518

mwarin:
https://hathitrust.atlassian.net/browse/DEV-1518

To test manually:

```bash
docker compose run --rm dev /bin/bash

mkdir -p /concordance/diffs/
touch /concordance/diffs/comm_diff_2025-01-01.txt.deletes
echo -e '56\t78' > /concordance/diffs/comm_diff_2025-01-01.txt.adds
export CONCORDANCE_PATH=/concordance
bash phctl.sh load concordance 2025-01-01 --inline
```

### DIFF
--- a/spec/integration_spec.rb
+++ b/spec/integration_spec.rb
@@ -40,7 +40,7 @@ RSpec.describe "phctl integration" do
         .to change { Cluster.count }.by(5)
     end
 
-    xit "Concordance loads concordance diffs" do
+    it "Concordance loads concordance diffs" do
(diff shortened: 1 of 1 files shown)


aelkiss:
We don't need to implement recluster yet, so I wouldn't worry about deletion at this point. If the existing integration test requires deletion/reclustering then I would say we should add a new integration test that only involves load/add/merge.

moseshll (approved):
Ran tests and Aaron's manual test with the expected results. APPROVE

PR #334, opened 2025-02-05
DEV-1449: schema updates for holdings/hathifile loading

aelkiss:
* ocn uses bigint -- while the maximum real OCN fits comfortably in 32 bits, it's possible we'll have other non-OCN numbers come through, and validation practices differ across different things (hathifiles, holdings, etc)

* index ocn and organization for holdings


moseshll (approved):
No red flags, APPROVE

PR #333, opened 2025-02-05
DEV-1525: separate dev & test databases

aelkiss:
* reorganize the schema into two files - one for the table definitions & one for the collections data

* bin/reset_database loads schema -- this is now required to run for dev (manually, noted in the README) and for test (run automatically when starting specs)

* docker compose just uses the mariadb root user for simplicity & so tests can recreate the database

* fixes generate_fake_data.rb -- avoids a couple mongoisms, loads the fake hathifiles data, and uses "docker compose" rather than "docker-compose"

PR #335, opened 2025-02-05
DEV-1520 - get cost report working

aelkiss:
This incorporates work from #329 and #333, and we will need to rebase after that's merged (and likely there will be more changes needed after #331 is merged as well). The relevant work here is my commits at the end. I'll leave some comments inline about specific things I'd like to think about.


spec/reports/cost_report_spec.rb
```diff
@@ -6,33 +6,37 @@
...
 require "data_sources/ht_organizations"
 
-RSpec.xdescribe Reports::CostReport do
+
+RSpec.describe Reports::CostReport do
   include_context "with tables for holdings"
 
-  let(:alo) { "allow" }
-  let(:dni) { "deny" }
+  let(:allow) { "allow" }
```
aelkiss, line 12: @mwarin I don't recall what specifically motivated pulling these out. If we still feel there's value in having them be `let`s rather than having them inlined as strings, maybe this is a reasonable compromise?

spec/reports/cost_report_spec.rb
```diff
@@ -131,25 +138,6 @@
     end
   end
 
-  describe "#matching_clusters" do
```
aelkiss, line 134: We aren't implementing this at all now, so tests are gone.

spec/spec_helper.rb
```diff
@@ -83,6 +80,11 @@
   end
 
   config.before(:each) do
+    # mock HT member data to use in tests. Tests may change this data, so
```
aelkiss: I don't know why this worked before. What I observed was that data added with `add_temp` leaked between tests and caused issues. This seemed like a reasonable fix.

spec/spec_helper.rb
```diff
@@ -118,24 +120,6 @@
   end
 end
 
-# Clusters and saves each element in an array of clusterables
-# DRY for the many times the tests need to do something like:
-# Clustering::ClusterXYZ.new(xyz).cluster.tap(&:save)
-def cluster_tap_save(*clusterables)
```
aelkiss, line 117: This isn't going to work as-is; eventually, we might find that we want a different approach to DRYing out setting up fixture data.

spec/reports/cost_report_spec.rb
```diff
@@ -91,28 +85,43 @@ def add_cluster_htitem(*htitems)
 
   describe "#cost_per_volume" do
     it "calculates cost per volume" do
-      expect(cr.cost_per_volume).to eq(cr.target_cost / cr.num_volumes)
+      # 3 volumes, target cost $10
```
aelkiss, line 87: I found it hard to understand what the state of the database was given the various before blocks and additional stuff being added. I figured being explicit in each test about what's there was the best option.

spec/reports/cost_report_spec.rb
```diff
@@ -91,28 +85,43 @@ def add_cluster_htitem(*htitems)
...
+      add_cluster_htitem(spm, mpm, ht_allow)
+      expect(cr.cost_per_volume).to be_within(0.01).of(3.33)
```
aelkiss, line 89: I also wanted to be explicit here about the expected result, rather than relying on other methods to compute it for us. That makes the tests somewhat more brittle, but that brittleness is kind of endemic in this class anyway given the fixture data..

spec/integration_spec.rb
```diff
@@ -101,9 +101,76 @@ def cleanup(output)
 
   # SharedPrintOps - integration tests are in the respective SharedPrint class
 
-  xdescribe "Report" do
+  describe "Report" do
     before(:each) do
-      phctl("load", "cluster_file", fixture("cluster_2503661.json"))
+      # Data from spec/fixtures/cluster_2503661.json, without commitments
```
aelkiss, line 106: An alternative could be loading a SQL dump, or at least a separate shared context with the insert statements.

aelkiss:
Rebased; ready for review. See comments inline.

lib/clusterable/ht_item.rb
```diff
@@ -37,20 +37,12 @@ def ic_volumes
         end
       end
 
-      def pd_volumes
```
aelkiss, line 40: We don't currently need to iterate over pd or all volumes, but we can add this back if we need it.

spec/spec_helper.rb
aelkiss: It's probably at the place where we could -- i.e. there are enough examples to figure out what exactly we don't want to repeat, and it might save some thinking in the specs if there were a generalized "persist" helper function that could do whatever it needed to do.

moseshll (approved):
I agree with the changes made to parts I have had my hands on. The rest I find comprehensible. No issues in the tests. APPROVE

mwarin (approved):
This all looks good to me.

aelkiss:
@mwarin thanks for the feedback. I will try to address those items.

spec/integration_spec.rb
aelkiss: Moved to a support context.

aelkiss:
@mwarin I've addressed feedback and will go ahead and merge once specs & standardrb are happy.

PR #336, opened 2025-02-07
DEV-1531: verify blank n_enum isn't an mpm

aelkiss:
Ensure that regardless of whether we get a blank or nil n_enum, CalculateFormat says the item is not a multi-part monograph.

### DIFF
--- a/spec/calculate_format_spec.rb
+++ b/spec/calculate_format_spec.rb
@@ -34,6 +34,24 @@ RSpec.describe CalculateFormat do
       ).to eq("spm")
     end
 
+    it "is NOT an MPM with null n_enum" do
+      htitem = double(:htitem, ht_bib_key: "12345", bib_fmt: "BK", n_enum: nil)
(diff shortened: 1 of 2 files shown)


spec/calculate_format_spec.rb
```diff
@@ -34,6 +34,24 @@
       ).to eq("spm")
     end
 
+    it "is NOT an MPM with null n_enum" do
+      htitem = double(:htitem, ht_bib_key: '12345', bib_fmt: 'BK', n_enum: nil)
```
aelkiss, line 38: Rather than trying to get the htitem/cluster to do what we wanted, basically we say -- what if htitem behaves like this? Then `CalculateFormat` should behave the way we want.

mwarin (approved):
Well, isn't this is the perfectly sized and scoped PR?
One line logic change and 2 clean tests. Neat!

The logic change makes sense and is cleaner than the original.

Approve.

PR #338, opened 2025-02-07
DEV-1528/DEV-1529: Optimize holdings load

aelkiss:
As we discussed; see notes on the individual commits.


aelkiss:
On my workstation this is about 3x as fast (1000 records per second) than before. I do probably want to add the key on uuid and ignore on duplicate, which gives us some measure of idempotence.

moseshll (approved):

PR #339, opened 2025-02-11
Fix some small issues with phctl report costreport

aelkiss:
* target cost setting was not honored on command line
* creating CommonJob had activesupportism


mwarin (changes requested):
This was my test plan:

```
$ docker compose build
$ docker compose run --rm test
$ docker compose run --rm test bash
$$ bundle install
$$ bash phctl.sh report costreport --target_cost XYZ --inline
$$ grep 'Target cost' /cost_reports/2025/costreport_20250212.tsv
```

Couldn't get phctl to work initially. I had to change: 

```diff
-    MARIADB_HOLDINGS_RW_DATABASE: holdings_dev
+    MARIADB_HOLDINGS_RW_DATABASE: holdings_test
...
-    MARIADB_HATHIFILES_RW_DATABASE: holdings_dev
+    MARIADB_HATHIFILES_RW_DATABASE: holdings_test
```

Not sure what the actual solution is, but this was a workaround that worked for me. With that, I could run the report: 

```
$ bash phctl.sh report costreport --target_cost 10101 --inline
$ grep 'Target cost' /cost_reports/2025/costreport_20250212.tsv
Target cost: 10101.0
```

... and see that target cost was indeed carried over from phctl arg to report.

I also noticed that the weights are being written in some kind of exponential notation?

```
Total weight: 0.18939e3
...
member_id       spm     mpm     ser     pd      weight  extra   total
allegheny       0.0     0.0     0.0     0.0     0.67e0  0.0     0.0
amherst 0.0     0.0     0.0     0.0     0.67e0  0.0     0.0
anu     0.0     0.0     0.0     0.0     0.1e1   0.0     0.0
...
```
Probably not introduced in this PR, but since we're in here.
I would also request something like:

```diff
     def run(output_filename = report_file)
       logger.info "Starting #{Pathname.new(__FILE__).basename}. Batches of #{ppnum maxlines}"
-
+      logger.info "Writing to #{output_filename}"
```
... to make it easier to see what file was or should have been written.

So, kinda approve (the things you set out to change), but also 3 changes requested:

# holdings_dev vs holdings_test in docker-compose.yml
# Weights being expressed as exponentials for no good reason
# Log output_filename

aelkiss:
I think for the dev one we need to run it in the dev container rather than the test container:

```ruby
# if you haven't yet set up the database schema
docker compose run --rm dev bin/reset_database.sh
docker compose run --rm dev ./phctl.sh phctl.sh report costreport --target_cost XYZ --inline
```

aelkiss:
The exponential notation seems to be an issue of MariaDB returning the weight as a `BigDecimal` rather than a `Float`. Forcing `to_f` before outputting seems like the easiest option.

aelkiss:
@mwarin Ready for re-review.

aelkiss:
I've deployed this branch in kubernetes (holdings-mariadb-testing) and will try running the cost report. We'll see how it goes.

mwarin (approved):
Looks fine! Fixes all my concerns and then some. 
Approve.

aelkiss:
I will fix the merge conflict and then merge this.

PR #340, opened 2025-02-13
Dev 1516 rename oclc concordance fields

moseshll:
- The names we dislike least: `variant` and `canonical`
   - `oclc_concordance.oclc` becomes `oclc_concordance.variant`
   - `oclc_concordance.canonical` stays the same
   - `OCNResolution` exposes `variant` in place of `deprecated` and `canonical` in place of `resolved`
- Everything else is a big old find and replace, so lots of very predictable name changes.


mwarin (approved):
I see no issue with what's in here, and the tests are green, so APPROVE. 

I'm only wondering if `lib/concordance_validation/` was intentionally left out? It's using terms like `resolved` vs `raw` and `resolved_ocn` vs `dead_ocn`. It's pretty limited, but since we're mass-replacing...

PR #342, opened 2025-02-13
DEV-1530 - improve cost report performance

aelkiss:
* memoize HtItem#cluster, Cluster#ht_items, Cluster#holdings,  Cluster#ocn_resolutions
* Cluster#reload - for invalidating memoized things (doesn't actually  reload, just sets to nil & "lazy-reloads")
* experiment with parallelizing frequency table generation

This is still very much WIP but may be worth taking a look at.


bin/generate_freq_table.rb
```diff
@@ -0,0 +1,41 @@
...
+marker = Milemarker.new(batch_size: 5000)
+
+#Services.holdings_db.loggers << Logger.new($stdout);
+log.info("#{$$}: starting freq table")
+filename = ARGV[0]
+fh = File.open(filename,"r")
+
+ARGF.each_line do |line|
+  fields = line.split("\t",-1)
+  # ignore things missing collection code?
```
mwarin, line 17: Move this to part of the query in bin/cost_report_chunk.sh ?

`WHERE ... AND hf.collection_code IS NOT NULL`

mwarin (commented):
I don't have a lot to say yet, but I think it makes sense.
I think a lot of the diff is/was from another cost report PR?

bin/parallel_cost_report/generate_freq_table.rb
```diff
@@ -0,0 +1,40 @@
...
+    collection_code: fields[6],
+    ocns: fields[7]
+  )
+  # ensure ocns are on the same cluster
+  Cluster.cluster_ocns!(h.ocns)
+
+  f.add_ht_item(h)
+
+  marker.incr
+  marker.on_batch { |m| log.info "#{$$}: #{m.batch_line}" }
```
moseshll, line 34: Consider a private wrapper for the logger call because of the repeated `$$` (which I had to look up).

moseshll (approved):
I see nothing surprising here. APPROVE

aelkiss:
Tests currently claim:

"Stopped processing SimpleCov as a previous error not related to SimpleCov has been detected'

but I don't see any other errors in the test output, so 🤷 . We can see if it continues.

aelkiss:
Superceded by #351

PR #343, opened 2025-02-14
DEV-1533 FrequencyTable: implement `+` method

moseshll:
- We base `+` on a new public method `append!` which does the heavy lifting.
- We discussed a number of possible optimizations, including copying parts of the addend's data structure
  - Unfortunately this approach runs the risk of clone/dup copying by reference and resulting in two FTs sharing data.
  - Added an `each` method so large tables need not be cloned with `to_h`
- If it turns out we don't need to be so rigorous about protecting the internals of this class, we can just make an accessor to `@table`
  - Our implementation of `to_h` (`.clone.freeze`) is flawed in that the return is not a deep copy anyway.

### DIFF
--- a/lib/frequency_table.rb
+++ b/lib/frequency_table.rb
@@ -25,6 +25,49 @@ class FrequencyTable
     @table.clone.freeze
   end
 
+  # Lower-memory alternative to `to_h` for the purposes of `+` and `append!`
+  def each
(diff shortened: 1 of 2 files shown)


mwarin (approved):
Clean code, good comments & excellent tests.

Heartily approved!

PR #344, opened 2025-02-17
DEV-1534 FrequencyTable: implement 'deserialize' method

moseshll:
- Opted to use JSON as the serialization format rather than the pre-existing custom jobby.
  - As a result, all keys are symbolized including the bucket name
    - E.g., `{1 => 1}` at a leaf node becomes `{:"1" => 1}`
    - One-line change in `lib/frequency_table.rb` with the needed conversion to symbol.
- Further work to try to hide the details of the underlying Hash implementation.
  - Remove `to_h` method and replace with `fetch` taking keywords.
  - Remove default_proc in the underlying hash. Maintaining it while implementing `+` and deserialization became painful and confusing.
- Related to the above, further work to make sure `FrequencyTable` is operating on a copy of data from outside and not subject to inadvertent data sharing.
  - That said, the data returned by `#fetch` is not a copy. Futz with it at your peril.
- To make the tests more comprehensible, can pass a Hash to the `FrequencyTable` initializer rather than populating with `add_ht_item`
  - There's a `CostReport` spec that used to futz with the internal `FrequencyTable` data, updated to now use JSON, could be further updated to take a Hash.

Outstanding questions:
- Is the use of Symbols for the buckets, instead of integers, going to be a problem?
- Is the use of pure JSON for the serialization format going to be a problem? (i.e., do we have existing files we must maintain compatibility with?)

### DIFF
--- a/lib/frequency_table.rb
+++ b/lib/frequency_table.rb
@@ -33,8 +33,9 @@ class FrequencyTable
     table.keys.sort
   end
 
-  def fetch(organization:, format: nil, bucket: nil)
-    # If bucket is passed and format is not, that's an error.
+  def fetch(organization: nil, format: nil, bucket: nil)
+    return table if organization.nil?
+
     data = table[organization.to_sym] || {}
     if format
       data = data[format.to_sym] || {}
--- a/lib/frequency_table.rb
+++ b/lib/frequency_table.rb
@@ -6,33 +6,45 @@ require "calculate_format"
 require "overlap/ht_item_overlap"
 
 class FrequencyTable
-  def initialize
-    @table = Hash.new do |hash, member|
-      hash[member] = Hash.new { |fmt_hash, fmt| fmt_hash[fmt] = Hash.new(0) }
-    end
+  protected attr_reader :table
+
+  # TODO: the tests would be more readable if we could pass either JSON or a Hash.
+  # Maybe try a `case` on the class of the parameter and parse JSON if it's a String
+  # and deep copy if it's a Hash.
+  def initialize(json: nil)
+    @table = json ? JSON.parse(json, symbolize_names: true) : {}
+  end
+
+  def to_json
+    JSON.generate(table)
+  end
+
+  def ==(other)
+    self.class == other.class && table == other.table
   end
 
   def organizations
-    @table.keys.sort
+    table.keys.sort
   end
 
-  def [](organization)
-    @table[organization.to_sym]
+  def fetch(organization:, format: nil, bucket: nil)
+    # If bucket is passed and format is not, that's an error.
+    data = table[organization.to_sym] || {}
+    if format
+      data = data[format.to_sym] || {}
+      if bucket
+        data = data[bucket.to_s.to_sym] || 0
+      end
+    end
+    data
   end
 
-  # For testing. I don't know if this is useful in the long run.
-  def to_h
-    @table.clone.freeze
-  end
-
-  # Lower-memory alternative to `to_h` for the purposes of `+` and `append!`
   def each
-    @table.each do |key, value|
+    table.each do |key, value|
       yield key, value
     end
   end
 
-  # Create a public #append(ft) method that we can call on a clone?
   def +(other)
     new_obj = self.class.new
     new_obj.append! self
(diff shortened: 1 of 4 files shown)


lib/frequency_table.rb
```diff
@@ -6,33 +6,51 @@
...
 
 class FrequencyTable
-  def initialize
-    @table = Hash.new do |hash, member|
-      hash[member] = Hash.new { |fmt_hash, fmt| fmt_hash[fmt] = Hash.new(0) }
+  protected attr_reader :table
+
+  def initialize(data: nil)
+    @table = case data
+    when String
```
mwarin, line 13: neat

lib/frequency_table.rb
```diff
@@ -6,33 +6,51 @@
...
+    JSON.generate(table)
+  end
+
+  def ==(other)
+    self.class == other.class && table == other.table
   end
 
-  def [](organization)
-    @table[organization.to_sym]
+  def organizations
```
mwarin, line 32: Is `organizations` used outside of tests?
If it's only used in tests I'd remove it and replace with `ft.table.keys.sort` or similar where needed.

spec/frequency_table_spec.rb
```diff
@@ -7,109 +7,88 @@
...
-
-  before(:each) do
-    Cluster.each(&:delete)
-    Cluster.create(ocns: ht2.ocns)
-    insert_htitem ht1
-    insert_htitem ht2
-  end
+  let(:umich_data) { {umich: {spm: {"1": 1}}} }
+  let(:upenn_data) { {upenn: {spm: {"1": 1}}} }
+  let(:ft_data) { umich_data.merge upenn_data }
```
mwarin, line 12: Doesn't look like you're using `ft_data` anywhere else. If not, I'd delete line 12 and change the next line to:

```
let(:ft_with_data) { described_class.new(data: umich_data.merge upenn_data) }
```

spec/frequency_table_spec.rb
```diff
@@ -7,109 +7,88 @@
...
+      expect(round_tripped.fetch(organization: :upenn)).to eq(ft_with_data.fetch(organization: :upenn))
     end
+  end
 
-    it "returns an Array of statistics when available" do
-      ft.add_ht_item ht1
-      ft.add_ht_item ht2
-      expect(ft[:umich][:spm]).to eq({1 => 1})
-      expect(ft[:upenn][:spm]).to eq({1 => 1})
+  describe "#organizations" do
```
mwarin, line 42: Possibly drop `organizations` and these tests if `organizations` isn't used anywhere else, or move the method to the spec file as a test method.

mwarin (approved):
Only minor nitpicks. APPROVE

lib/frequency_table.rb
```diff
@@ -73,15 +88,23 @@ def add_ht_item(ht_item)
...
-    [].tap do |data|
-      @table.sort.each do |org, freq_data|
-        data << [org, JSON.generate(freq_data)].join("\t")
-      end
-    end.join "\n"
+  def increment(organization:, format:, bucket:)
+    org = organization.to_sym
+    fmt = format.to_sym
+    bucket = bucket.to_s.to_sym
+    table[org] = {} unless table.key?(org)
```
aelkiss, line 99: I think this is much clearer than relying on a deeply-nested layer of defaults.

lib/frequency_table.rb
```diff
@@ -6,33 +6,48 @@
...
-  def [](organization)
-    @table[organization.to_sym]
+  def ==(other)
+    self.class == other.class && table == other.table
   end
 
-  # For testing. I don't know if this is useful in the long run.
-  def to_h
-    @table.clone.freeze
+  def fetch(organization: nil, format: nil, bucket: nil)
```
aelkiss, line 32: This makes things much clearer in terms of understanding what the layers in the hash are

aelkiss (approved):
This all looks great to me. FWIW I don't think we need to worry _too_ much about performance of merging, so long as it's O(m*n) where m is the number of organizations and n is the number of frequency tables we're merging -- the actual generation of the frequency tables is going to be the vast majority of the run time

PR #347, opened 2025-02-20
DEV-1526: fixes to get the old style HT003.*.tsv holdings files to load

mwarin:
In order to "load all the data" we need to be able to load the legacy .tsv format.

There were a few small changes needed to the tsv-loader, presumably because the holdings table definitions have changed some since last we tried loading data this old.

Lines were rejected if they: lacked status, or lacked uuid, or had an outdated mono_multi_serial value. Luckily the correct or dummy value is easy to figure out. 

(I noticed that there are currently no uniqueness constraints on the `holdings` table. I think we need to figure that out, so that one cannot accidentally load the same holdings file over and over again and grow the holdings table each time. However, I don't think it is necessary to figure that out right now.)

### DIFF
--- a/lib/clusterable/holding.rb
+++ b/lib/clusterable/holding.rb
@@ -107,7 +107,7 @@ module Clusterable
     def self.batch_add(batch)
       columns = table.columns
       rows = batch.map { |h| columns.map { |c| h.public_send(c) } }
-      table.import(columns, rows)
+      table.insert_ignore.import(columns, rows)
     end
 
     def date_received=(date)
--- a/lib/clusterable/holding.rb
+++ b/lib/clusterable/holding.rb
@@ -50,6 +50,8 @@ module Clusterable
         fields[7] = "mpm"
       when "serial"
         fields[7] = "ser"
+      when ""
+        fields[7] = "mix"
       end
 
       # The old tsv files likely don't have a uuid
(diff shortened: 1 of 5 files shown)


lib/clusterable/holding.rb
```diff
@@ -36,19 +37,43 @@ def self.holding_to_record(holding_line)
...
-       date_received: DateTime.parse(fields[5]),
-       enum_chron: fields[6],
-       mono_multi_serial: fields[7],
-       issn: fields[8],
-       n_enum: fields[9],
-       n_chron: fields[10],
-       gov_doc_flag: !fields[11].to_i.zero?,
-       uuid: fields[12]}
+
+      if fields[3].empty?
```
moseshll, line 41: This is a long method. Since there is already some inline translation of field values going on in the Hash initializer (e.g., `DateTime.parse`) one could do class methods `self.translate_tsv_status`, `self.translate_tsv_type` (or `*_mono_multi_serial` or whatever) and `self.translate_tsv_uuid` (even though we expect it to always be `nil` I guess). That would make these procedures individually testable, or just make them private to reduce entrypoint bloat.

That would make the Hash initializer dense, but would keep semantically related items close to each other, e.g.,
```
...
mono_multi_serial: translate_tsv_mono_multi_serial(fields[7]),
...
```

I suspect this is largely a matter of taste. (I am Mr. Dense when left to my own devices)

PR #348, opened 2025-02-27
DEV-1570: gnuplot visualization for frequency table

aelkiss:
This should be adequate to generate visualizations as needed/on demand, but if we need to do something more routinely and share with members this is probably not the best starting point.


moseshll:
Runs fine for me.

PR #350, opened 2025-02-28
Small fixes pulled out from #342

aelkiss:
I pulled out a couple of small changes not directly related to cost report performance from https://github.com/hathitrust/holdings-backend/pull/342 so that we can get them merged independently of the churn and experiments happening on that branch.


spec/overlap/ht_item_overlap_spec.rb
```diff
@@ -46,16 +46,33 @@
 
   # we end up inserting the same kinds of things...
   # this adds the given item and holdings to cluster c
-  def cluster_item_holdings(item:, holdings:)
```
aelkiss, line 49: @mwarin we'd had a TODO here to use `load_test_data` so I went ahead and to-did it

PR #351, opened 2025-03-03
DEV-1530,DEV-1563,DEV-1566,DEV-1569: Run cost report in parallel via catalog solr dump

aelkiss:
* memoize HtItem#cluster, Cluster#ht_items, Cluster#holdings, Cluster#ocn_resolutions
* Cluster#reload - for invalidating memoized things (doesn't actually reload, just sets to nil & "lazy-reloads")
* FrequencyTable#append: vivify keys when needed
* CostReport can take a path or a directory with precomputed frequency tables
* CostReport can sum the directory of frequency tables
* add_frequency_tables -> phctl report cost-report --precomputed-frequency-table-dir
* run cost report w/ solr records -  fetch solr records, instantiate clusters & htitems using data from that, sum frequency tables at the end
* adjust connection validation -- was revalidating the sql connection multiple times for every query, leading to a lot of 'SELECT NULL's - may have been a reason for poor performance in production


aelkiss:
Supercedes #342

aelkiss:
@moseshll @mwarin This starts cleaning up some of the churn from the previous (DEV-1530-cost-report-performance) branch; I plan to address the remaining TODOs on DEV-1569 about putting the pieces together to run the whole cost report.

lib/cluster.rb
```diff
@@ -134,11 +136,11 @@ def ocns=(ocns)
   end
 
   def ocn_resolutions
-    Clusterable::OCNResolution.with_ocns(ocns)
+    @ocn_resolutions ||= Clusterable::OCNResolution.with_ocns(ocns).to_a
   end
 
   def ht_items
-    Clusterable::HtItem.with_ocns(ocns)
+    @ht_items ||= Clusterable::HtItem.with_ocns(ocns).to_a
```
aelkiss, line 143: These memoizations aren't necessarily required for the cost report as it stands now given that we are building htitems from the solr records and applying directly to the cluster, but I don't think they're a problem.

lib/data_sources/holdings_db.rb
```diff
@@ -17,9 +17,9 @@ class HoldingsDB < SimpleDelegator
 
     def initialize
       @rawdb = self.class.connection
-      # Always check that we're actually connected and reconnect if necessary
+      # Check once every few seconds that we're actually connected and reconnect if necessary
       @rawdb.extension(:connection_validator)
-      @rawdb.pool.connection_validation_timeout = -1
+      @rawdb.pool.connection_validation_timeout = 5
```
aelkiss, line 22: We can observe if we get any kind of connection drops -- I think it will make more sense to automatically retry than to constantly revalidate the connection.

aelkiss:
@mwarin @moseshll We may want to look at this together, there's a fair amount here, but most of the relevant stuff is in the new files.

mwarin (approved):
I think this is all fine and good. It adds a bit of complexity, and couples holdings to the catalog Solr. We'll want to make corresponding updates to one or more Confluence pages to reflect these changes.

Approve.

lib/cost_report_workflow.rb
```diff
@@ -0,0 +1,88 @@
...
+    system("split -l #{chunk_size} #{allrecords_ndj} #{working_directory}/records_")
+  end
+
+  def queue_frequency_table_jobs
+    batch = Sidekiq::Batch.new
+    callback_params = {"frequency_table_dir" => working_directory}
+    batch.description = "Generate frequency tables for cost report"
+    batch.on(:success, CostReportWorkflow::Callback, callback_params)
+    batch.jobs do
+      Dir.glob("#{working_directory}/records_??").each do |chunk|
```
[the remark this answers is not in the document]
aelkiss: That's a good point.

lib/cost_report_workflow.rb
```diff
@@ -0,0 +1,88 @@
...
+require "reports/cost_report"
+require "sidekiq/batch"
+require "sidekiq_jobs"
+require "solr/cursorstream"
+
+class CostReportWorkflow
+  class CostReportWorkflow::Callback
+    def on_success(_status, options)
+      Reports::CostReport.new(precomputed_frequency_table_dir: options["frequency_table_dir"]).run
+      FileUtils.remove_entry(options["frequency_table_dir"])
```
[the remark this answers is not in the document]
aelkiss: By default there will be about 1,800 files in that directory, about 18 gigabytes total. I'm open to saving it for debugging purposes for the time being, but I don't know that I'd want to keep it (or have to manually clean up) every time we run a cost report.

spec/cost_report_workflow_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+        chunk_size: 2,
+        inline_callback_test: true
+      )
+    end
+
+    it "runs cost report when all chunks complete" do
+      workflow.run
+
+      expect(true)
+      #      expect(File.exist?("#{ENV["TEST_TMP"]/costreports/:w
```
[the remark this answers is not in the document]
aelkiss: Yup, needs more work.

mwarin (changes requested):
UNAPPROVEd as per @aelkiss 's wishes, so we don't accidentally merge without doing the desired changes.

lib/cluster.rb
```diff
@@ -150,7 +152,14 @@ def commitments
   end
 
   def holdings
-    Clusterable::Holding.with_ocns(ocns)
+    @holdings ||= Clusterable::Holding.with_ocns(ocns).to_a
+  end
+
+  # invalidate the cache after adding items elsewhere
+  def reload
```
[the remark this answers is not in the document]
aelkiss: Part of the reason I left it as `reload` was that we used it as that in mongo-land. I went ahead and changed `reload` to `invalidate_cache` everywhere, even in code we'll eventually remove and tests we aren't currently running.

lib/cost_report_workflow.rb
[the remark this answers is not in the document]
aelkiss: With 10,000-line chunks, we also will end up beyond this. I need to think a bit because records_?? will also match e.g. `records_aa.freqtable.json`, which we don't want (if somehow there were extra files present from a previous run in that directory)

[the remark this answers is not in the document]
aelkiss: I think I will output to e.g. `records_00001.ndj`, and then write to `records_00001.freqtable.json`, which should be a bit easier to manage.

lib/cost_report_workflow.rb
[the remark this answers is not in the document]
aelkiss: For now I'm going to leave it. At least for the first few runs we'll probably want to inspect it.

aelkiss:
@mwarin Ready for re-review. I think I addressed all your comments. Especially take a look at the now actually-existent tests.

lib/cost_report_workflow.rb
```diff
@@ -63,7 +63,7 @@ def dump_solr_records
   end
 
   def split_solr_records
-    system("split -l #{chunk_size} #{allrecords_ndj} #{working_directory}/records_")
+    system("split -d -a 5 --additional-suffix=.ndj -l #{chunk_size} #{allrecords_ndj} #{working_directory}/records_")
```
mwarin, line 68: I like these `split` parameters. Didn't know it could do all that!
I will try to remember this for future use.

lib/cluster.rb
```diff
@@ -156,7 +156,7 @@ def holdings
   end
 
   # invalidate the cache after adding items elsewhere
-  def reload
+  def invalidate_cache
```
mwarin, line 159: This makes more sense now.

lib/reports/cost_report.rb
```diff
@@ -146,6 +146,9 @@ def total_cost_for_member(member)
 
     # Reads either a set of frequency tables from a given directory containing
     # json files, or a single one from a file.
+    #
+    # If neither are given, read nothing, and compute the frequency table on
```
mwarin, line 150: :heavy_check_mark:

lib/reports/frequency_table_from_solr.rb
```diff
@@ -10,6 +10,8 @@
 
 module Reports
   class FrequencyTableFromSolr
+    attr_reader :solr_records, :output_file, :batch_size
```
mwarin, line 13: :heavy_check_mark:

spec/cost_report_workflow_spec.rb
```diff
@@ -83,16 +88,52 @@
...
+      # {"umich"=>{"mpm"=>{"1"=>8}, "spm"=>{"1"=>3}}}
+      expect(ft.frequencies(organization: :umich, format: :mpm))
+        .to contain_exactly(Frequency.new(bucket: 1, frequency: 8))
+
+      expect(ft.frequencies(organization: :umich, format: :spm))
+        .to contain_exactly(Frequency.new(bucket: 1, frequency: 3))
     end
 
-    it "cleans up when cost report completes" do
+    xit "cleans up when cost report completes" do
```
mwarin, line 136: If this is a consequence of `CostReportWorkflow::Callback.on_success` (currently) not deleting the intermediate files, I'd like a comment in addition to the test being `xit`ted, lest we forget why.

mwarin (approved):
I see all of my feedback addressed, and other than wanting a single additional comment on one of the tests, I don't think there is anything blocking an APPROVE here.

spec/cost_report_workflow_spec.rb
aelkiss: Done.

PR #356, opened 2025-03-13
DEV-1569: invoke cost report workflow via phctl

aelkiss:
Just a quick change to add the ability to invoke the workflow via phctl.

I'm not sure `costreport-workflow` is the right name here, and there's potential confusion for this vs. `phctl report costreport`. I'm also not sure whether we want to continue to retain the ability for `CostReport` to build a frequency table itself vs. needing to take one -- I don't know that we need to fix it now, but it may be related insofar as what needs to be exposed via phctl.

### DIFF
--- a/lib/phctl.rb
+++ b/lib/phctl.rb
@@ -109,6 +109,15 @@ module PHCTL
       run_common_job(Reports::CostReport, options)
     end
 
+    desc "costreport-workflow --ht-item-count NUM --ht-item-pd-count NUM (--chunk-size SIZE)", "Dump records from solr, split into chunks of chunk-size records, generate frequency tables for each chunk, sum the resulting frequency tables, and generate a cost report based on that table."
+    option :chunk_size, type: :numeric, default: 10000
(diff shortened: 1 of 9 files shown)


lib/phctl.rb
```diff
@@ -109,6 +109,15 @@ def costreport
       run_common_job(Reports::CostReport, options)
     end
 
+    desc "costreport-workflow --ht-item-count NUM --ht-item-pd-count NUM (--chunk-size SIZE)", "Dump records from solr, split into chunks of chunk-size records, generate frequency tables for each chunk, sum the resulting frequency tables, and generate a cost report based on that table."
```
mwarin, line 112: Just noticing that the output from `bash phctl.sh report` is starting to get hard to read with such long invocations, but I reckon that's just life.

mwarin (approved):
I'm on board with this. If we end up keeping `ht_item_count`  & `ht_item_pd_count` as input parameters we should document the reasoning behind it, or it'll quickly stop making sense. Good enough for our current purposes, though.

PR #357, opened 2025-03-17
Remove cluster_ocns table, clustering, and & loading

aelkiss:
Main functional changes:
* Clusters are no longer persisted to the database
* `Cluster` (given ocns) can pull in related OCNs from catalog record & concordance
* Remove "Clustering" classes that load data & connect them to clusters.

Most of the changes here are related to:
* removing code
* changing tests to consistently use `load_test_data`
* changing references to `Cluster.for_ocns` (which now always returns a cluster rather than a potentially-empty array of clusters)
* removing a smattering of other Mongo-specific things that are no longer in use

Main things to review:
* Anything that has changed (rather than just wholesale removal) in the code
* `Cluster.for_ocns`
* Anything that has changed in the tests besides data loading/setup

PR #358, opened 2025-03-17
DEV-1571: estimates

aelkiss:
This is likely to be somewhat sluggish since we have to do several database queries for each OCN, but gets us something for now. We can take a look at performance and decide if we'd rather query via Solr.

Depends on #357 (which @mwarin looked at but I'd like to get a quick look from @moseshll as well as since it's pretty big)

Marked as draft just because I want to merge #357 first.

### DIFF
--- a/lib/reports/estimate.rb
+++ b/lib/reports/estimate.rb
@@ -31,13 +31,13 @@ module Reports
     def find_matching_ocns(ocns = @ocns)
       ocns.each do |ocn|
         marker.incr
-        cluster = Cluster.find_by(ocns: ocn.to_i,
-          "ht_items.0": {"$exists": 1})
+        cluster = Cluster.for_ocns([ocn.to_i])
         next if cluster.nil?
 
         @num_ocns_matched += 1
-
-        next if clusters_seen.include?(cluster._id)
+        # If data changes while this is running, items from some clusters could
+        # get double-counted.
+        next if clusters_seen.include?(cluster.ocns.hash)
 
         count_matching_items(cluster)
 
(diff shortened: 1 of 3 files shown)


aelkiss (commented):
Note that the estimate will be rather off if it's running against the dev (rather than production) database, and also relies on getting the item count & pd count.

We will likely want to look at a production environment fairly soon.

lib/reports/estimate.rb
```diff
@@ -31,13 +31,12 @@ def cost_report
...
         marker.incr
-        cluster = Cluster.find_by(ocns: ocn.to_i,
-          "ht_items.0": {"$exists": 1})
+        cluster = Cluster.for_ocns([ocn.to_i])
         next if cluster.nil?
 
         @num_ocns_matched += 1
-
-        next if clusters_seen.include?(cluster._id)
+        # this will be an issue if data changes while this is running
```
mwarin, line 38: > this will be an issue if data changes while this is running

What kind of issue? A cluster gets counted more/less than it should? If so, it's "just" an estimate, does not need to be 100% accurate.

mwarin (approved):
Wow, that was much fewer changes than I thought were going to be necessary.
I don't see any issues, just one Q about one comment.
Approve.

lib/reports/estimate.rb
aelkiss: Right, overlap for some clusters could get double-counted.

aelkiss:
Rebased, will merge once tests pass.

lib/reports/estimate.rb
aelkiss: I updated the comment to reflect this.

PR #359, opened 2025-03-18
Quick fixes for max ocn

aelkiss:
Quick fixes for a very minor annoyance with tests that were slower than they needed to be.

* use Faraday instead of open-uri
* use FileUtils.touch instead of sleeping in max_ocn_spec

### DIFF
--- a/lib/scrub/max_ocn.rb
+++ b/lib/scrub/max_ocn.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require "json"
-require "open-uri"
+require "faraday"
 require "services"
 
 OCLC_URL = "https://www.oclc.org/apps/oclc/wwg"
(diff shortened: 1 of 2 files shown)


lib/scrub/max_ocn.rb
```diff
@@ -66,7 +66,7 @@ def make_call
       if @mock
         IO.read("spec/fixtures/max_oclc_response.json")
       else
-        URI.parse(OCLC_URL).open.read
+        Faraday.get(OCLC_URL).body
```
aelkiss, line 69: We're including faraday in the gemfile anyway (not 100% sure why), but we might as well use it in preference to open-uri, not sure it makes much difference really though.

spec/scrub/max_ocn_spec.rb
```diff
@@ -31,8 +31,8 @@
     expect(File.exist?(loc)).to be(false)
     smo = described_class.new(mock: true, age_limit: 1)
     smo.ocn
+    FileUtils.touch(loc, mtime: Time.now - 10)
```
mwarin, line 34: Well that never occurred to me. Sure makes the test faster.

mwarin (approved):
Works and should make the tests faster. Approve.

PR #361, opened 2025-03-19
DEV-1571: improvements for estimate

aelkiss:
* fix for overcounting matched OCNs
* aligned pd/ic logic with cost report
* check OCNs we've already seen rather than hashes -- should be more resilient to data changes; also avoids collisions

### DIFF
--- a/lib/reports/estimate.rb
+++ b/lib/reports/estimate.rb
@@ -8,7 +8,7 @@ module Reports
   # Generate IC estimate from a list of OCNS
   class Estimate
     attr_accessor :ocns, :ocn_file, :h_share_total, :num_ocns_matched, :num_items_matched, :num_items_pd,
-      :num_items_ic, :clusters_seen, :marker
+      :num_items_ic, :ocns_seen, :marker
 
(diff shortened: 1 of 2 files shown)


lib/reports/estimate.rb
```diff
@@ -87,15 +89,16 @@ def total_estimated_ic_cost
     private
 
     def count_matching_items(cluster)
-      @clusters_seen << cluster.ocns.hash
+      ocns_seen.merge(cluster.ocns)
 
       @num_items_matched += cluster.ht_items.count
       cluster.ht_items.each do |ht_item|
-        if ht_item.access == "allow"
+        if !Clusterable::HtItem::IC_RIGHTS_CODES.include?(ht_item.rights)
```
mwarin, line 96: I think this would read better reversed & without the negation:

```ruby
if Clusterable::HtItem::IC_RIGHTS_CODES.include?(ht_item.rights)
  @num_items_ic += 1
else
  @num_items_pd += 1
  next
end
```

mwarin (approved):
Approval with one optional suggestion.

PR #362, opened 2025-03-19
DEV-1613 enable & fix skipped tests in spec/scrub_runner

moseshll:
- Add `Clusterable::Holding.count` method to take the place of moribund spec helper `cluster_count`
- Add comment with future work breadcrumb in spec helper `cluster_count`

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -145,6 +145,8 @@ def fixture(filename)
   File.join(__dir__, "fixtures", filename)
 end
 
+# This does not work. We now use CLusterable::Holding.count for scrub_runner_spec.rb
+# For commitments and ht_items we recommend a `count` method on the appropriate class.
 def cluster_count(field)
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -145,6 +145,8 @@ def fixture(filename)
   File.join(__dir__, "fixtures", filename)
 end
 
+# This does not work. We now use CLusterable::Holding.count for scrub_runner_spec.rb
+# For commitments and ht_items we recommend a `count` method on the appropriate class.
```
aelkiss, line 149: Can we just remove `cluster_count`?

aelkiss (approved):
👍 Looks like the specs were pretty easy to fix. Worth removing `cluster_count` entirely?

spec/spec_helper.rb
moseshll: We decided to leave it as a breadcrumb since other `xit` and/or `xdescribe` tests use it and it's out of scope to go a-hunting around those use cases.

PR #363, opened 2025-03-19
removed holdings deleter class & spec

mwarin:
This class was there as a helper for deleting specific holdings in MongoDB, where it was a pain to do so otherwise.

Since we are no longer operating under the oppressive yoke of MongoDB, this can go away.
It wasn't plugged in to anything important, so this should be a clean delete.


moseshll (approved):
Geez. There's nothing to review LOL. APPROVE

PR #364, opened 2025-03-19
DEV-1616: remove phctl "cleanup holdings"

mwarin:
Another holdings cleanup thing that we no longer need.
Not very connected, so deleting a couple of blocks of code should do it.
Fixed an unrelated brittle test in `spec/overlap/cluster_overlap_spec.rb` while I was doing it

### DIFF
--- a/spec/clusterable/ht_item_spec.rb
+++ b/spec/clusterable/ht_item_spec.rb
@@ -173,4 +173,32 @@ RSpec.describe Clusterable::HtItem do
       expect(no_ocn1.batch_with?(no_ocn3)).to be false
     end
   end
+
+  describe "Factory (spec/factories/ht_item.rb)" do
+    it "can specify billing_entity" do
+      ht_item = build(:ht_item, billing_entity: "umich")
+      expect(ht_item.billing_entity).to eq "umich"
+    end
+    it "can specify collection_code" do
+      ht_item = build(:ht_item, collection_code: "MIU")
+      expect(ht_item.collection_code).to eq "MIU"
+    end
+    it "makes sure collection_code and billing_entity belong to the same org" do
+      ht_item_with_billing_entity = build(:ht_item, billing_entity: "umich")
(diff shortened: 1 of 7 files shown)


spec/overlap/cluster_overlap_spec.rb
```diff
@@ -6,15 +6,17 @@
 RSpec.describe Overlap::ClusterOverlap do
   include_context "with tables for holdings"
 
-  let(:spm) { build(:ht_item, enum_chron: "", billing_entity: "ucr") }
+  let(:spm) { build(:ht_item, enum_chron: "", collection_code: "PU") }
```
mwarin, line 9: Not related to the task at hand, but I kept getting failed tests due to this.
The best way to build a ht_item with a predetermined organization as `billing_entity` is to set the `collection_code` that is mapped to that organization. 

Clunky? Yes. Pretty? No.

mwarin: Yeah. It's been too annoying for too long. Looking at it.

mwarin: See e1bf96de47508d2a5547c2ec0c0be9c66b0720ea for some additions to the `ht_item` factory, and some associated tests.

spec/clusterable/ht_item_spec.rb
```diff
@@ -173,4 +173,32 @@
       expect(no_ocn1.batch_with?(no_ocn3)).to be false
     end
   end
+
+  describe "Factory (spec/factories/ht_item.rb)" do
```
mwarin, line 177: These tests show that you can now, in tests using the factory for ht_items, set `billing_entity` directly and get the expected results.

aelkiss (approved):
Makes sense to me. Now we can make some of the tests make more sense where right now we are specifying the collection code, but no rush to do that.

moseshll (approved):
We worked on the factory issue on Thu and I'm satisfied with the result.

mwarin:
Squashing and merging...

PR #365, opened 2025-03-20
DEV-1571 - split out read-only vs. read-write databases

aelkiss:
Primarily this splits out the read-only & read-write databases, but also:
* try trilogy mysql driver
* provide cluster to holding & htitem on construction to avoid repetitive mysql queries if we do e.g. Cluster.for_ocns(whatever).ht_items and then reach through those ht items to get the cluster -- could be a cause of performance issues for estimates
* fix a couple issues with duplicate & misformatted queries


spec/support/holdings_tables.rb
```diff
@@ -2,18 +2,20 @@
 require "services"
 
 RSpec.shared_context "with tables for holdings" do
-  let(:db) { Services.holdings_db }
+  let(:db_rw) { Services.holdings_rw_db }
+  let(:db_ro) { Services.ht_ro_db }
```
moseshll, line 6: It should be obvious after a moment's reflection that this DB is read-only outside the test context, but viewed in isolation it's momentarily confusing that we wrap this in a transaction. Might be worth a single line comment. Or, the naming might be suspect since the point is not that the DB is `ro` but rather that it's the `ht` database.

moseshll (approved):
Couple of nitpicky comments for consideration. I had no other issues with this update. APPROVE

lib/data_sources/mariadb.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require "dotenv"
+Dotenv.load(".env")
+
+require "delegate"
```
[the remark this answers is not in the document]
aelkiss: Should we remove `dotenv` then? I guess we aren't really using it. It does appear to work without explicitly requiring "delegate" (and we aren't requiring that anywhere else) so maybe that's available by default? Likewise I think it must be auto-requiring `trilogy` when we set that as the adapter.

spec/support/holdings_tables.rb
aelkiss: Agreed, this is rather confusing, at least for the `:let` statements. We may not really need to even include the `rw` vs. `ro` in `Services` as such.

lib/data_sources/mariadb.rb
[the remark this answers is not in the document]
aelkiss: I found that we did still need `delegate` -- something in the tests must be implicitly requiring it, but `phctl` didn't work without it.

spec/support/holdings_tables.rb
aelkiss: I have changed this to omit the `rw` vs. `ro` in the names here.

PR #367, opened 2025-03-25
initial wip

mwarin:


spec/cluster_spec.rb
```diff
@@ -5,8 +5,8 @@
```
mwarin, line 5: This was done to fix a once-in-1000 brittle test, not directly related to the ticket.

spec/factories/holding.rb
```diff
@@ -15,6 +15,8 @@
     status {}
     uuid { SecureRandom.uuid }
     gov_doc_flag { [true, false].sample }
+    delete_flag { 0 }
+    enum_chron { "" }
```
mwarin, line 19: This (`enum_chron { "" }`) was done so that holdings built with this factory don't get `nil` for the normalized enumchron fields, which become `NULL` when persisted to the database. Wouldn't normally be a problem but caused the round-trip test in `pre_load_backup_spec.rb` to fail.

lib/scrub/scrub_runner.rb
```diff
@@ -29,35 +30,37 @@
 module Scrub
   # Scrubs and loads any new org-uploaded files.
   class ScrubRunner
+    attr_reader :force, :file_transfer, :organization, :force_holding_loader_cleanup_test
```
mwarin, line 33: Addressed over-use of `@` by adding `attr_reader`s

lib/scrub/pre_load_backup.rb
```diff
@@ -0,0 +1,83 @@
...
+      records.where(delete_flag: 1).count
+    end
+
+    private
+
+    def records
+      Services.holdings_db[:holdings].where(query)
+    end
+
+    def query
```
aelkiss, line 61: I'm waffling a bit on whether the query parts should move into `Clusterable::Holding`. Ultimately, I think it's OK for it to be here, since basically everything here relates to manipulating the database.

lib/scrub/pre_load_backup.rb
```diff
@@ -0,0 +1,83 @@
...
+    def initialize(organization:, mono_multi_serial:)
+      @organization = organization
+      @mono_multi_serial = mono_multi_serial
+      validate
+    end
+
+    def write_backup_file
+      # Note that this writes an empty file if there are no matching records.
+      File.open(backup_path, "w") do |backup_file|
+        records.each do |record|
```
aelkiss, line 29: We should use `paged_each` as in `Clusterable::Holding` to avoid waiting for the entire query to complete and (I think) loading everything into memory.

spec/cluster_spec.rb
```diff
@@ -34,6 +34,7 @@
       expect(described_class.for_ocns([ocn1, ocn2]).ocns).to contain_exactly(ocn1, ocn2)
     end
 
+    # brittle test?
```
aelkiss, line 37: Still true?

spec/factories/holding.rb
aelkiss: Probably fine. We should be treating nil/null/empty string as the same for enumchron-related purposes in any case.

aelkiss (commented):
All looks good to me, the only thing I'd suggest adding is the `paged_each` to avoid loading all holdings into memory.

spec/cluster_spec.rb
mwarin: Nope, removed.

lib/scrub/pre_load_backup.rb
mwarin: Ah, yes. Added `paged_each`.

moseshll (approved):
All looks familiar. I got burned again by the 1-in-1000 test collision so it will be good to see it fixed. APPROVE

PR #369, opened 2025-03-26
Debug logging / performance WIP

aelkiss:
I've deployed this to kubernetes so I can better get a handle on database performance there.

In addition to adding info on the DEBUG flags to the README, I've removed some other long-outdated information there.

### DIFF
--- a/lib/clusterable/holding.rb
+++ b/lib/clusterable/holding.rb
@@ -94,10 +94,18 @@ module Clusterable
       Services.holdings_table
     end
 
+    # apply order and used paged each -- fetches blocks of rows
+    # by doing e.g. select * from holdings where uuid > LAST_UUID order by uuid limit 1000
+    def self.paged_each(dataset, &block)
+      return to_enum(__method__, dataset) unless block_given?
+
+      dataset.order(:uuid).paged_each(strategy: :filter, &block)
+    end
+
     def self.all
       return to_enum(__method__) unless block_given?
 
-      table.each do |row|
+      paged_each(table) do |row|
         yield from_row(row)
       end
     end
@@ -117,7 +125,7 @@ module Clusterable
 
       dataset = table.where(organization: organization)
 
-      dataset.each do |row|
+      paged_each(dataset) do |row|
         yield from_row(row)
       end
     end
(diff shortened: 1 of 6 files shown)


lib/services.rb
```diff
@@ -27,12 +27,13 @@
 Services.register(:ht_organizations) { DataSources::HTOrganizations.new }
 Services.register(:ht_collections) { DataSources::HTCollections.new }
 Services.register(:logger) do
-  Logger.new($stderr, level: Logger::INFO)
```
aelkiss, line 30: Not sure why we chose to log to stderr, but best practice these days for k8s is to log to stdout; I've tripped across this a couple times in trying to redirect / review log output locally.

lib/clusterable/holding.rb
```diff
@@ -94,10 +94,18 @@ def self.table
       Services.holdings_table
     end
 
+    # apply order and used paged each -- fetches blocks of rows
+    # by doing e.g. select * from holdings where uuid > LAST_UUID order by uuid limit 1000
+    def self.paged_each(dataset,&block)
```
aelkiss, line 99: The mysql2 driver also supports a 'streaming' option; the trilogy driver doesn't. Probably worth comparing overall performance there.

mwarin (approved):
Can't say I understand the effect of the changes in 3582f29d343f685c8f6c8aeb762ba6dd0c6d0dd6, but I'm on board with the other changes and the tests are green.

aelkiss:
> Can't say I understand the effect of the changes in [3582f29](https://github.com/hathitrust/holdings-backend/commit/3582f29d343f685c8f6c8aeb762ba6dd0c6d0dd6), but I'm on board with the other changes and the tests are green.

With the regular `each` it fetches all rows into memory before starting to return them; with `paged_each` it does a query for a block of 1000 rows at a time. Since this is in holdings and we have the uuid as a primary key, we can do:

```sql
SELECT * FROM holdings WHERE organization = 'umich' ORDER BY uuid ASC limit 1000
```
and then once we know the last uuid in that set, we can do

```sql
SELECT * FROM holdings WHERE organization = 'umich' AND uuid > LAST_UUID ORDER BY uuid ASC limit 1000
```

I will go ahead and merge but hopefully that helps explain the rationale for `paged_each`.

PR #371, opened 2025-04-02
DEV-1561 investigate performance of concordance validation

moseshll:
- Use sqlite3 gem and executable to write deduplicated concordance file into a table.
- `ConcordanceValidation::Concordance` currently takes care of writing intermediate files:
  - The database itself
  - The deduplicated concordance
- A full validation takes 12-ish hours with this lash-up
- Delta computation reimplemented using `zcat/sort/comm`


lib/concordance_validation/concordance.rb
```diff
@@ -1,34 +1,134 @@
...
+
+    CREATE_CANONICAL_INDEX_SQL = <<~SQL
+      CREATE INDEX IF NOT EXISTS canonical_idx ON concordance (canonical)
+    SQL
+
+    TEST_MEMORY_DB = false
+
+    # Match lines that are exactly X<tab>X and delete them.
+    # The literal regex that get passed to sed is /^\(.*\)\t\(\1\)$/d
+    SED_DEDUPE_REGEX = "/^\\(.*\\)\\t\\(\\1\\)$/d"
```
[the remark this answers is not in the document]
mwarin: No, but in all seriousness, this is about as good as it can be done. I was wondering if there was a way to do `Regexp.new(...).to_s` for this, but I doubt that would work.

lib/concordance_validation/concordance.rb
```diff
@@ -1,34 +1,134 @@
...
+
+    def canonical_to_variant(canonical)
+      @db.execute("SELECT variant FROM concordance WHERE canonical=? AND variant!=?", [canonical, canonical])
+        .flatten
+        .uniq
     end
 
     def file_handler
-      if /\.gz$/.match?(@infile)
+      if /\.gz$/.match?(@concordance_file)
```
mwarin, line 131: `@concordance_file.end_with?(".gz")` might be half a cpu-cycle shorter and more Ruby-idiomatic.

Besides, is `file_handler` even used anywhere?

lib/concordance_validation/delta.rb
```diff
@@ -6,58 +6,40 @@
...
-
-    def write(fout, diffs)
-      diffs.keys.sort.each do |canonical_ocn|
-        diffs[canonical_ocn].each do |variant_ocn|
-          fout.puts [variant_ocn, canonical_ocn].join("\t")
-        end
-      end
-      fout.flush
+      # Lines only in old concordance == deletes
+      comm_cmd = "bash -c 'comm -23 <(#{sort_cmd @old_concordance}) <(#{sort_cmd @new_concordance}) > #{deletes_file}'"
```
mwarin, line 16: I think you can skip the `bash -c` part, and just have `comm_cmd` be:

```ruby
comm_cmd = "comm -23 <(#{sort_cmd @old_concordance}) <(#{sort_cmd @new_concordance}) > #{deletes_file}"
```

spec/concordance_validation/delta_spec.rb
```diff
@@ -9,45 +9,69 @@
...
-  end
-
-  describe "#initialize" do
-    it "opens the old and new concordances for reading" do
-      expect(delta.old_conc.to_a.size).to eq(2)
-      expect(delta.new_conc.to_a.size).to eq(2)
-    end
+    File.write(old_concordance, old_concordance_data)
+    File.write(new_concordance, new_concordance_data)
+    `gzip -c #{old_concordance} > #{old_concordance_gz}`
```
mwarin, line 48: I'd use `system` here instead of backticks, since you're not capturing the output.

lib/concordance_processing.rb
```diff
@@ -9,31 +10,40 @@ def validate(fin, fout)
...
-    c.variant_to_canonical.each_key do |variant|
-      next if c.variant_to_canonical[variant].count.zero?
-
+    milemarker = Milemarker.new(batch_size: 10_000, name: "validate concordance")
+    milemarker.logger = Services.logger
+    c.db.prepare("SELECT variant FROM concordance").execute.each do |variant|
+      variant = variant[0]
+      # Can this ever happen? This is from the in-memory hash table version.
+      # next if c.variant_to_canonical[variant].count.zero?
       begin
```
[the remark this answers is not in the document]
moseshll: And because the behavior of each error type is essentially the same, these can be combined. The double `puts` in ErrorA is redundant since we can make the ErrorA class responsible for that "Cycles" message.

lib/concordance_validation/concordance.rb
moseshll: No, it's not used. I got rid of it, yay!

lib/concordance_validation/delta.rb
moseshll: I think the command needs to specify `bash`, otherwise tests complain ```Syntax error: "(" unexpected```. I think some part of how we're doing `<(subcommand)` is a feature of recent bashes. That was the impression I got.

aelkiss (approved):
This all looks good to me. I had some thoughts about alternate approaches (in the comments) but I think what we have here makes more sense.

PR #372, opened 2025-04-03
DEV-1590: API for ETAS access queries

mwarin:
See https://hathitrust.atlassian.net/browse/DEV-1590

## Context

In a rare move away from "using the database as API", here's a Sinatra API that can be used to answer holdings questions (such as "how many holdings matching the volume `ht_id:foo` has `organization:bar` reported?")

### item_access

Querying `v1/item_access?ht_id=foo&organization=bar` should return:

```ruby
{                                      
  "copy_count" => cluster.copy_counts[organization],
  "format" => cluster.format,                       
  "n_enum" => ht_item.n_enum,                       
  "ocns" => ht_item.ocns.sort                       
}.to_json                                  
```

It should return a `#404` if any of the arguments are missing, if the route is wrong, or if the `ht_id` does not match a `ht_item`.

`copy_count` defaults to 0, even if the given organization does not exists.

### item_held_by

Querying  `v1/item_held_by?ht_id=foo` should return:

```ruby
{                                      
  "organizations" => cluster.organizations_in_cluster
}.to_json
```

`organizations` should never be empty, if the item exists, since all items have at least a `billing_entity` organization (the org that submitted the item). 

## Testing

Run tests: `bundle exec rspec spec/api/etas_api_spec.rb`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,8 +19,10 @@ gem "trilogy"
 gem "zinzout"
 gem "puma"
 gem "rack-session"
+gem "rackup"
 gem "sidekiq"
 gem "sidekiq-batch", git: "https://github.com/breamware/sidekiq-batch"
+gem "sinatra"
 gem "solr_cursorstream"
 
 group :test do
@@ -28,6 +30,7 @@ group :test do
   gem "factory_bot"
   gem "faker"
   gem "hathifiles_database", git: "https://github.com/hathitrust/hathifiles_database.git", branch: "main"
+  gem "rack-test"
   gem "rspec"
   gem "rspec-sidekiq"
   gem "simplecov"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -116,6 +116,8 @@ GEM
     method_source (1.1.0)
     milemarker (1.0.1)
     minitest (5.25.4)
+    mustermann (3.0.3)
+      ruby2_keywords (~> 0.0.1)
     mysql2 (0.5.6)
     net-http (0.6.0)
       uri
@@ -148,9 +150,17 @@ GEM
       nio4r (~> 2.0)
     racc (1.8.1)
     rack (3.1.12)
+    rack-protection (4.1.1)
+      base64 (>= 0.1.0)
+      logger (>= 1.6.0)
+      rack (>= 3.0.0, < 4)
     rack-session (2.1.0)
       base64 (>= 0.1.0)
       rack (>= 3.0.0)
+    rack-test (2.2.0)
+      rack (>= 1.3)
+    rackup (2.2.1)
+      rack (>= 3)
     rainbow (3.1.1)
     rake (13.2.1)
     rbs (3.8.1)
@@ -209,6 +219,7 @@ GEM
       ruby-lsp (~> 0.23.0)
     ruby-prof (1.7.1)
     ruby-progressbar (1.13.0)
+    ruby2_keywords (0.0.5)
     securerandom (0.4.1)
     sequel (5.90.0)
       bigdecimal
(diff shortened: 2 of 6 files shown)


lib/api/etas_api.rb
```diff
@@ -0,0 +1,59 @@
+require "cluster"
+require "clusterable/ht_item"
+require "services"
+require "sinatra"
+
+class EtasAPI < Sinatra::Base
+  set :show_exceptions, ENV["show_sinatra_exceptions"] || false
+  no_matching_data_error = {"application_error" => "no matching data"}.to_json
+
+  get "/ping" do
```
aelkiss, line 10: Is the thought this could be used as a route for a health check?

lib/api/etas_api.rb
```diff
@@ -0,0 +1,59 @@
...
+    validate_params(params, ["organization", "ht_id"])
+
+    organization = params["organization"]
+    ht_id = params["ht_id"]
+
+    ht_item = Clusterable::HtItem.find(item_id: ht_id)
+    cluster = ht_item.cluster
+
+    return_doc = {
+      "lock_id" => ht_item.ocns.sort.join(","),
```
aelkiss, line 25: This won't always correspond to the lock id as computed now, maybe we should just call it "ocns" and return it as a JSON array. I also wonder if we should return format?

lib/api/etas_api.rb
```diff
@@ -0,0 +1,59 @@
...
+
+class EtasAPI < Sinatra::Base
+  set :show_exceptions, ENV["show_sinatra_exceptions"] || false
+  no_matching_data_error = {"application_error" => "no matching data"}.to_json
+
+  get "/ping" do
+    "pong"
+  end
+
+  get "/etas/item_access" do
```
aelkiss, line 14: I wonder if we might be able to have a more generic name than `/etas/item_access` -- ideally this API wouldn't need to know anything about ETAS, and is just returning information about holdings. 

It might be worth thinking about what the other API routes we need are. We know we need similar (but not identical) kinds of information for resource sharing, section 108, umich ssd access -- but I think we could meet those needs from the same endpoint and with different/additional count info.

The other need is to get a list of all the organizations that hold an item, but without the need for copy counts, to support catalog indexing. I don't know if that's the same endpoint but without an organization parameter or a different endpoint.

Maybe worth discussing some.

moseshll, line 14: I would recommend versioning the API right from the get-go -- in other words `get "/v1/path/to/a/thing"` etc. And also echo Aaron's comment about other routes. We should try to get a spec for a basic suite of endpoints that if designed well should protect us from too much version churn.

moseshll: rights_api has an example of redirecting to the versioned URL from an unversioned one.

config.ru
```diff
@@ -0,0 +1,2 @@
+require "api/etas_api"
```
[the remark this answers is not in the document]
mwarin: Moved to `bin/config.ru`, added `api` service as suggested.

lib/api/etas_api.rb
mwarin: Renamed the API to `HoldingsAPI` and changed this route to `/item_access`.

mwarin: Yes. Implemented!

lib/api/etas_api.rb
mwarin: Changed `lock_id` (string) ==> `ocns` (array), added `format` (value of `cluster.format` for the cluster of the ht_item).

aelkiss (approved):
I think this makes sense as a first pass.

We will probably need some iteration on `item_held_by` for performance reasons (so we can batch some of the queries) but I think this is a good start.

One thing this doesn't account for is the `mapto_inst_id` stuff going on in https://github.com/hathitrust/babel/blob/main/mdp-lib/Access/Holdings.pm#L66-L73. I believe that primarily affects California because they submit their holdings separately, but got access to things for ETAS if _any_ UC institution holds them.

Regardless, I'd say we should go ahead and merge this assuming that you and @moseshll are overall happy with the _interface_ of the API, and we can make further changes to the data or operation as needed.

moseshll (approved):
I'm happy with this, APPROVE

PR #373, opened 2025-04-11
DEV-1640: check ocn counts as part of scrub and reject file if diff is too great

mwarin:
## Context

`Scrub::RecordCounter` is the class that performs the line-count check between new scrubbed files and old scrubbed files (that have the same `organization` & `mono_multi_serial`) before they are loaded, in order to alert & abort in cases where the absolute change is greater than 5% (or whatever `Settings.scrub_line_count_diff_max` is set to).

This PR adds the ability to compare number of _distinct ocns_ between old & new as well, since that is likely a better predictor of fee change than plain line counts.

For instance: Assume we have already loaded a file with 1000 records of umich spm holdings via `bash phctl.sh scrub ORG`. We should then get alerted & stopped when we next try to load a file with way too few (10) or way too many (100K) holdings. This is the existing functionality.

What is being added in this PR is a similar warning when the number of distinct ocns is way too big or small compared to the most recent set of loaded ocns.

The alert consists of a message of what is wrong, and the file being loaded should be rejected. It should be possible to override and load it anyways by adding `--force` (`bash phctl.sh scrub ORG --force`)

### DIFF
--- a/lib/scrub/autoscrub.rb
+++ b/lib/scrub/autoscrub.rb
@@ -19,8 +19,9 @@ module Scrub
     # by another ruby class.
     attr_reader :output_struct, :out_files, :logger_path, :item_type
 
-    def initialize(path)
+    def initialize(path, force = false)
       @path = path
+      @force = force
 
       # @member_id and @item_type are used in the path to scrub_logger, but we
       # also need somewhere to log to before we know @member_id and @item_type
@@ -48,7 +49,7 @@ module Scrub
       @encoding = Utils::Encoding.new(@path)
       convert_to_utf8 unless @encoding.ascii_or_utf8?
 
-      Services.scrub_logger.info("Started scrubbing #{@path}")
+      Services.scrub_logger.info("Started scrubbing #{@path}, --force=#{@force}")
       write_to do |out_file|
         @member_holding_file.parse do |holding|
           out_file.puts(holding.to_json)
--- a/lib/scrub/record_counter.rb
+++ b/lib/scrub/record_counter.rb
@@ -4,14 +4,19 @@ require "utils/line_counter"
 
 module Scrub
   class RecordCounter
-    attr_reader :organization, :item_type, :struct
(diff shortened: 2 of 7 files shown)


lib/scrub/autoscrub.rb
```diff
@@ -19,8 +19,9 @@ class AutoScrub
     # by another ruby class.
     attr_reader :output_struct, :out_files, :logger_path, :item_type
 
-    def initialize(path)
+    def initialize(path, force = false)
```
mwarin, line 22: just passing force along for logging purposes

spec/scrub/record_counter_spec.rb
```diff
@@ -15,14 +15,28 @@
...
-  def put_x_lines_in_file(x, file)
+
+  # Write a testfile with a certain number of { "ocn": i } ndjs.
+  def testfile(lines:, file:)
     File.open(file, "w") do |f|
-      1.upto(x) do |i|
-        f.puts(i)
+      1.upto(lines) do |i|
+        hash = {"ocn" => i}
+        f.puts hash.to_json
```
moseshll, line 24: I think I might compact this down to `f.puts({"ocn" => 1}.to_json)` or `f.puts({ocn: 1}.to_json)` (I was going to suggest a version without parentheses not realizing `puts` would get confused by the hash literal). Line 34 as well.

lib/scrub/record_counter.rb
```diff
@@ -1,18 +1,24 @@
+require "scrub/member_holding_file"
```
moseshll, line 1: This `require` not needed? Unless `count_distinct_ocns` is intended to move there?

moseshll (approved):
Just a couple stylistic notes. I like this and tests pass, so APPROVE

lib/scrub/record_counter.rb
mwarin: Indeed, this is a remnant from an abandoned idea. Removed.

PR #374, opened 2025-04-15
DEV-1515 load most recent concordance in development

moseshll:
- Add `ConcordanceLoader` class and tests.
  - Two subclasses handle full concordance and delta scenarios.
  - `phctl load concordance YYYYMMDD` means load deltas.
  - `phctl load concordance PATH` means truncate and load full concordance.
- Add `Loader::FileLoader#add_batch_deletes` to possibly replace the apparently unused `load_deletes` method (should be a lot more efficient?)
- Test fixtures from `spec/concordance_validation/data` moved to `spec/fixtures/concordance/raw`

Note: this includes support for loading a full concordance file (in addition to .adds/.deletes) but it's unlikely we would ever want to do it that way. So I am soliciting opinions on whether support for full concordance loading should be removed.


lib/concordance_validation/delta.rb
```diff
@@ -1,11 +1,29 @@
...
 module ConcordanceValidation
   # Performs a diff of two validated concordances
   class Delta
+    DIFF_FILE_BASE_TEMPLATE = "comm_diff_YYYYMMDD.txt"
+    def self.diffs_directory
+      File.join(Settings.concordance_path, "diffs")
+    end
+
+    def self.diff_file_base(date: Date.today)
+      DIFF_FILE_BASE_TEMPLATE.gsub("YYYYMMDD", date.strftime("%Y-%m-%d"))
```
mwarin, line 16: gsub -> sub

lib/loader/concordance_loader.rb
```diff
@@ -0,0 +1,106 @@
...
+      line.chomp!
+      return nil if line.empty?
+      line.split("\t").map(&:to_i)
+    end
+  end
+
+  # Updates existing concordance data in the oclc_concordance table using an .adds and .deletes file.
+  # The superclass provides all the default behavior. Defining an subclass to make the semantics
+  # a little bit clearer.
+  class ConcordanceLoaderDelta < ConcordanceLoader
```
mwarin, line 87: Neat. I like this subclass pattern w/ `self.for`.

lib/sidekiq_jobs.rb
```diff
@@ -52,6 +53,24 @@ def perform(filename)
       end
     end
 
+    class Concordance
+      include Sidekiq::Job
+      def perform(filename_or_date)
+        # TODO: should we wrap this whole thing in a Sequel transaction?
```
mwarin, line 59: I think the answer is yes?

spec/integration_spec.rb
```diff
@@ -39,6 +39,28 @@ def phctl(*args)
       expect { phctl("load", "holdings", fixture("umich_fake_testdata.ndj")) }
         .to change { Clusterable::Holding.table.count }.by(10)
     end
+
+    describe "concordance" do
+      around(:each) do |example|
+        saved_concordance_path = Settings.concordance_path
+        Settings.concordance_path = File.join(__dir__, "fixtures", "concordance")
```
mwarin, line 46: I think you could do:

```ruby
  Settings.concordance_path = fixture("concordance")
```

mwarin (approved):
Looks good! No red flags or big question marks.

Approve!

lib/sidekiq_jobs.rb
moseshll: Actually no, we don't want the extra headache. (The tests start failing.) If something goes wrong we fix and re-run.

PR #375, opened 2025-04-17
Map-reduce workflow for overlap report, cost report, estimate

aelkiss:
There are quite a few changes here worth reviewing, and in retrospect it would have been better to do this in pieces.

I think the most relevant changes are:

* Running estimates and overlap reports by fetching records from solr with the desired OCNs (a given list of OCNs for an estimate, or the depositor's list of OCNs from holdings for an overlap report
* Extracting the machinery we used to run a cost report in parallel as a more generic `Workflow::MapReduce` that takes a data source (i.e. the thing that dumps the solr records), a map step, and a reduce step; see comments in that class. I then split up things from e.g. `Report::Estimate` to `Workflow::Estimate`.
* Fetching holdings via batches of OCNs rather than individual OCNs. (`SolrBatch`). This is the most confusing part and needs more attention.

~There is still a considerable amount of cleanup & refactoring to do here, listed at https://hathitrust.atlassian.net/browse/DEV-1629~


lib/solr_batch.rb
```diff
@@ -0,0 +1,41 @@
...
+    fetch_holdings(organization)
+  end
+
+  private
+
+  def fetch_holdings(organization)
+    ocn_map = Hash.new { |h, k| h[k] = [] }
+    records.each do |record|
+      cluster = record.cluster
+      # FIXME: severely janky
```
aelkiss, line 20: This definitely needs some re-thought.

lib/solr_batch.rb
```diff
@@ -0,0 +1,41 @@
+require "clusterable/holding"
+require "solr_record"
+
+# A chunk of solr records for which we can fetch holdings in a batch
```
aelkiss, line 4: Needs more documentation

config/settings.yml
```diff
@@ -15,3 +15,12 @@ replace_commitment_report_path: <%= ENV["REPLACE_COMMITMENT_REPORT_PATH"] %>
 shared_print_report_path: <%= ENV["SHARED_PRINT_REPORT_PATH"] %>
 shared_print_update_report_path: <%= ENV["SHARED_PRINT_UPDATE_REPORT_PATH"] %>
 target_cost: <%= ENV["TARGET_COST"] %>
+solr_data_source:
```
aelkiss, line 18: I tried to exact all the "magic numbers" to settings, and at least git "batch size" here a prefix.

lib/solr_batch.rb
aelkiss: @moseshll I think this is a bit more comprehensible than before but let me know what you think here

lib/phctl.rb
```diff
@@ -203,6 +180,72 @@ def weeding_decision(organization)
...
+          test_mode: options[:test_mode]
+        }
+      end
+    end
+
+    desc "costreport --ht-item-count NUM --ht-item-pd-count NUM (--chunk-size SIZE)", "Dump records from solr, split into chunks of chunk-size records, generate frequency tables for each chunk, sum the resulting frequency tables, and generate a cost report based on that table."
+    option :ht_item_count, type: :numeric
+    option :ht_item_pd_count, type: :numeric
+    def costreport_workflow
+      components = {
```
aelkiss, line 213: I spent a fair amount of time trying to come up with an option for instantiating a workflow that could:
* have the same interface in phctl / via sidekiq as when directly running in testing
* put the responsibility for marshalling/unmarshalling parameters in `SidekiqJobs`
* not involve confusing stuff in `Workflows::MapReduce`

I don't know if I completely succeeded, but hopefully this is a reasonable compromise among those ends.

lib/reports/cost_report.rb
```diff
@@ -177,20 +172,6 @@ def read_freq_tables(dir, file)
       end
     end
 
-    def compile_frequency_table
```
aelkiss, line 180: We no longer expect to do this in a standalone way -- we were only really using this in test; the tests are now changed to precompute a frequency table and pass it in.

lib/reports/cost_report.rb
aelkiss, line 1: I _didn't_ move this to workflow because we might still want to run this standalone e.g. if we want to re-run a cost report from a given frequency table with a new target cost

lib/sidekiq_jobs.rb
```diff
@@ -34,10 +35,82 @@
 end
 
 module Jobs
+  # make functions available as Jobs::whatever
```
aelkiss, line 38: I tried to move the responsibility for serializing/deserializing things to JSON (or json-able types) here rather than any of the actual jobs needing to know anything about that.

lib/sidekiq_jobs.rb
```diff
@@ -34,10 +35,82 @@
...
+    case value
+    when Hash
+      prepare_params(value)
+    when Array
+      value.map { |entry| prepare_value(entry) }
+    when Symbol, Class
+      value.to_s
+    when String, Integer, true, false, nil
+      value
+    when ->(v) { v.respond_to?(:to_h) }
```
aelkiss, line 59: i.e. if `value` can respond to `to_h`, call that, then run it back through

lib/sidekiq_jobs.rb
```diff
@@ -34,10 +35,82 @@
...
+      prepare_params(value.to_h)
+    else
+      raise "Can't prepare value '#{value}' as sidekiq job param"
+    end
+  end
+
+  def symbolize_keys(params)
```
aelkiss, line 66: I know ActiveSupport does some of this, but we were really trying not to rely on it.

lib/solr_record.rb
```diff
@@ -18,8 +19,8 @@ def cluster
     end
   end
 
-  def ht_items
-    @ht_items ||= cluster.ht_items =
+  def parse_ht_items
```
aelkiss, line 22: We want to ensure this always happens rather than relying on lazy-loading. If we make a `SolrRecord` we are always going to use its ht items.

lib/workflows/map_reduce.rb
```diff
@@ -0,0 +1,115 @@
...
+  # Gets data (one record per line) from a data source, splits it into chunks,
+  # then runs a job on each chunk (the "mapper"); when the mapping steps
+  # complete, runs a callback (the "reducer") with the output
+  # from the mapper.
+  #
+  # Jobs are run on files containing the given records_per_job lines (default 10,000)
+  class MapReduce
+    class Callback
+      def on_success(_status, options)
+        reducer = Object.const_get(options["component_class"])
```
aelkiss, line 18: Because this is more tightly coupled to Sidekiq it has to deal with deserializing itself. Could maybe do something different here in the future.

lib/workflows/map_reduce.rb
```diff
@@ -0,0 +1,115 @@
...
+        # Don't delete intermediate files for now to aid in debugging.
+        # In the future: consider removing by default, but disabling that if a debug flag is present.
+        # FileUtils.remove_entry(options["working_directory"])
+      end
+    end
+
+    def initialize(
+      working_directory: default_working_directory,
+      records_per_job: 10000,
+      components: {},
```
aelkiss, line 31: I don't love that components is a hash, but we are validating it, and doing it this way (vs. dynamically defining more attributes/methods/whatever) seems better. I had tried a version where there was a `WorkflowSpec` that encapsulated the components, but IMO it ended up increasing verbosity without increasing clarity.

aelkiss:
@moseshll @mwarin This is fully ready for review. It may make sense to look at it and talk through together again because there is a lot here -- but less than it may look like, as much was moved & reorganized from reports to workflows.

See the top-level comment as well as a few things inline.

aelkiss:
I rebased against mariadb and squashed the WIP, so individual commits might be more comprehensible and there should be less churn.

spec/workflows/estimate_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+    let(:ocn_count) { File.join(ENV["TEST_TMP"], "ocn_count.estimate.json") }
+
+    let(:data_source) { described_class.new(ocn_file: ocn_file) }
+
+    describe "#dump_records" do
+      before(:each) { File.write(ocn_file, ocns.join("\n")) }
+
+      context "with four ocns and two items" do
+        let(:ocns) { [1, 2, ht_allow.ocns, ht_deny.ocns].flatten }
+        let(:ht_allow) { build(:ht_item, rights: "pd") }
```
mwarin, line 135: The `counts input & matching ocns` test below will fail if either `ht_allow` and/or `ht_deny` get assigned ocn 1/2 from the factory, or if they get the same ocn.
I would specify ocns 3 & 4 for `ht_allow` & `ht_deny` to remove this brittleness.

spec/workflows/overlap_report_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+      let(:h2) { build(:holding, organization: "ualberta") }
+      let(:ht) { build(:ht_item, ocns: [h.ocn], access: "deny") }
+      let(:ht2) { build(:ht_item, ocns: [h.ocn], access: "allow", rights: "pd") }
+
+      before(:each) do
+        load_test_data(h, h2, ht, ht2)
+        mock_solr_oclc_search(solr_response_for(ht, ht2))
+      end
+
+      it "has a line for each ht_item in the holding organization rpt" do
```
mwarin, line 107: (and one header line)

lib/clusterable/holding.rb
```diff
@@ -110,12 +110,16 @@ def self.all
       end
     end
 
-    def self.with_ocns(ocns, cluster: nil)
+    def self.with_ocns(ocns, cluster: nil, organization: nil)
       return to_enum(__method__, ocns, cluster: cluster) unless block_given?
 
-      dataset = table.where(ocn: ocns.to_a)
+      dataset = table.where(ocn: ocns.to_a.map(&:to_s))
```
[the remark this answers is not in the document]
aelkiss: MariaDB doesn't always use the index if given an integer (e.g. `WHERE ocn = 12345`, but it does when given a string (`WHERE ocn = '12345'`). I can't say that I understand why that is.

lib/workflows/map_reduce.rb
```diff
@@ -0,0 +1,115 @@
...
+    end
+
+    def split_records
+      system("split -d -a 5 --additional-suffix=.split -l #{records_per_job} #{allrecords} #{working_directory}/records_")
+    end
+
+    def inline_reduce
+      # In test, where sidekiq is not running, we do this
+      # instead of relying on the on_success-hook.
+      if ENV["DATABASE_ENV"] == "test"
```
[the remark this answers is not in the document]
aelkiss: Maybe not? I don't think it will do anything awful other than try to run the callback too early which won't produce any useful output.

spec/frequency_table_spec.rb
```diff
@@ -58,6 +60,180 @@
...
+            ocns: ht_serial.ocns,
+            billing_entity: "not_ht_serial.billing_entity",
+            access: "deny",
+            rights: "ic"
+          )
+        end
+        let(:holding_serial) do
+          build(:holding,
+            ocn: ht_serial.ocns.first,
+            enum_chron: "3",
```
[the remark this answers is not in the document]
aelkiss: These were mostly copied from cost report; I can look through & see.

spec/workflows/estimate_spec.rb
aelkiss: I will go ahead and specify 3/4.

One thing I saw recently with `factory_bot` that I think would help significantly with some of these potentially brittle tests around ocns is sequences: https://thoughtbot.github.io/factory_bot/sequences/inline-sequences.html

aelkiss: https://hathitrust.atlassian.net/browse/ETT-118, but I think the sequences would be good to try using the next time we find ourselves writing a test like this.

aelkiss:
> A couple issues I noticed in running it in dev:
> 
>     * The callback for the overlap report failed:
> 
> 
> ```
>  TypeError: no implicit conversion of nil into String
>  /usr/src/app/lib/workflows/overlap_report.rb:146:in 'File.exist?'
> ```
> 
>     * The default working directory is set to `/htprep/holdings-testing-mariadb/costreports/work` regardless of the kind of report

I changed it to just use `/htprep/holdings-testing-mariadb/work` rather than nesting it under a particular report type -- varying it would probably require moving the logic to generate the path elsewhere, which ends up complicating things, and there's no particular need for that.

The `nil` issue was because `overlap_reports_path` wasn't set (it was still `etas_overlap_reports_path`) in the settings; I'll update that in ht_tanka.

aelkiss:
@mwarin I think I addressed the issues you pointed out in the latest commit (https://github.com/hathitrust/holdings-backend/pull/375/commits/563a91a52aa63459a945f40a251730de48b8b170). Do you want me to go ahead and mark those comments resolved or do you want to look again?

mwarin:
I've resolved my comments.

lib/workflows/solr.rb
```diff
@@ -0,0 +1,73 @@
...
+      end
+    end
+
+    # Fetches holdings for batches of solr records from an input file and hands
+    # them off for further analysis.
+    #
+    # If organization is given, limits holdings loaded to that organization;
+    # otherwise, loads all holdings.
+    class Analyzer
+      # TODO address with organization in solr batch
```
moseshll, line 53: I think this TODO is TODONE?

lib/solr_batch.rb
moseshll: Yes this makes sense.

moseshll (approved):
Looks good to me. Tests pass and I sorta understand everything.

PR #376, opened 2025-04-21
 DEV-1639: abort `Scrub::ScrubRunner` if it detects a type mismatch

mwarin:
DEV-1639

This PR adds a new class, `Scrub::TypeChecker`, which checks the `mono_multi_serial` (a.k.a. `item_type`) of files-about-to-be-scrubbed against the holdings already loaded by the same member, and aborts if the types-about-to-be-scrubbed are not a subset of the types already loaded.

It's plugged into `Scrub::ScrubRunner.run`, in order to catch the type mismatch as early as possible. There are no changes needed to `phctl` or sidekiq jobs.

If the `TypeChecker` aborts the `ScrubRunner`, a human will need to intervene... we don't have a good automatic solution.

### DIFF
--- a/lib/scrub/scrub_runner.rb
+++ b/lib/scrub/scrub_runner.rb
@@ -16,8 +16,10 @@ require "loader/holding_loader"
 require "scrub/autoscrub"
 require "scrub/chunker"
 require "scrub/malformed_file_error"
+require "scrub/member_holding_file"
 require "scrub/pre_load_backup"
 require "scrub/record_counter"
+require "scrub/type_checker"
 require "sidekiq_jobs"
 require "sidekiq/batch"
 require "utils/file_transfer"
@@ -46,6 +48,8 @@ module Scrub
       Services.logger.info "Running org #{organization}."
       new_files = check_new_files
       Services.logger.info "Found #{new_files.size} new files: #{new_files.join(", ")}."
+      check_new_types(new_files)
+
       new_files.each do |new_file|
         run_file(new_file)
       end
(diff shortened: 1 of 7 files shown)


lib/scrub/type_checker.rb
```diff
@@ -0,0 +1,42 @@
...
+  class TypeChecker
+    attr_reader :organization, :new_types, :old_types, :message
+
+    def initialize(organization:, new_types: [], old_types: [])
+      @organization = organization
+      @new_types = new_types
+      @old_types = old_types.any? ? old_types : get_old_types
+    end
+
+    def validate
```
moseshll, line 13: Tests aside, if the caller is going to raise when this method returns `false` and then retrieve the exception message, I would consider just `raise @message` as the negative return value from this. Maybe do the logging here, too. That would potentially make less stuff in `ScrubRunner#check_new_types`.

lib/scrub/scrub_runner.rb
```diff
@@ -174,6 +178,25 @@ def download_to_work_dir(file)
 
     private
 
+    def check_new_types(files)
+      # Get the types of the files we are loading.
+      types = []
```
moseshll, line 183: Any reason not to use a `Set` for this, and either `to_a` it before passing it to the TypeChecker, or even have TypeChecker take sets in its initializer? You want to sort them only for display, and that's the edge case.

moseshll (approved):
Just a couple of additional ideas added inline that may be worth a think. No issues with the PR as it stands, APPROVE

lib/scrub/type_checker.rb
mwarin: I've axed the `message` part and now it's just raising. Added a new error class.

lib/scrub/scrub_runner.rb
mwarin: Actually swapping in Set works just fine, so I did.

PR #378, opened 2025-04-24
DEV-1678: start using Cluster::Overlap in HoldingsAPI

mwarin:
... to correctly handle copy counts of mpms, added test to that effect.

This PR also contains some refactoring and removal of dead code.


spec/api/holdings_api_spec.rb
```diff
@@ -52,6 +52,11 @@ def parse_body(call)
     JSON.parse(last_response.body)
   end
 
+  include_context "with tables for holdings"
```
aelkiss, line 55: extraneous/leftover -- remove

aelkiss (approved):
All looks great to me. There's just one small extraneous leftover from the first iteration of adding a test in `holdings_api_spec.rb`; otherwise I think this is good to go.

PR #379, opened 2025-04-28
ETT-194: get all holdings on a record given items in JSON form

aelkiss:
This use is quite coupled to its use from catalog indexing (given that it needs a solr record) but IMO that's OK given the desire for performance in catalog indexing, especially the full reindex. That said, given the coupling that already exists to the catalog and that we already are passing around solr records in holdings, I think it makes sense to do here and doesn't introduce any _new_ coupling.

Next step is to take a look at https://github.com/hathitrust/hathitrust_catalog_indexer/blob/main/lib/ht_traject/ht_print_holdings.rb#L13 and see whether we can in fact generate this partial record format there.

### DIFF
--- a/lib/api/holdings_api.rb
+++ b/lib/api/holdings_api.rb
@@ -2,6 +2,7 @@ require "clusterable/ht_item"
 require "overlap/cluster_overlap"
 require "services"
 require "sinatra"
+require "solr_record"
 
 class HoldingsAPI < Sinatra::Base
   set :show_exceptions, ENV["show_sinatra_exceptions"] || false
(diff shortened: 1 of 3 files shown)


aelkiss:
It looks like this should be doable from the catalog indexing side:

https://github.com/hathitrust/hathitrust_catalog_indexer/blob/main/lib/ht_traject/ht_item.rb#L91 can get the `ht_json` field at that point (via `to_json`); that gets called from https://github.com/hathitrust/hathitrust_catalog_indexer/blob/a5a0add89ec6c1f136f51811912da08e41b50056/indexers/ht.rb#L8, and the context has (in `output_hash` there) the other fields we need -- id, format, oclc, oclc_search.

One additional defense for this coupling (as opposed to just asking for a list of holdings for the items, without the context of the OCNs for the record, the enumchrons for the items, etc) is that it ensures that the results we get back depend *only* on the data in solr and the holdings info -- not on potentially out-of-sync info in the hathifiles database.

PR #380, opened 2025-04-29
ETT-195, add route to Holdings API for getting orgs with brittle/lost/missing holdings

mwarin:
* Added route `/v1/item_held_by_brlm` to satisfy https://hathitrust.atlassian.net/browse/ETT-195.
* Added tests


spec/api/holdings_api_spec.rb
```diff
@@ -232,8 +232,62 @@ def parse_body(call)
         build(:holding, organization: "umich", ocn: ocn),
         build(:holding, organization: "smu", ocn: ocn)
       )
-      response = parse_body(v1(url_template("item_held_by", item_id: item_id_1, organization: "umich")))
+      response = parse_body(v1(url_template("item_held_by", item_id: item_id_1)))
       expect(response["organizations"].sort).to eq ["smu", "umich"]
     end
   end
+
```
mwarin, line 239: For these tests at least we don't need any of the organizations to be real organizations, so I went with a..d.

lib/api/holdings_api.rb
```diff
@@ -46,6 +46,26 @@ class HoldingsAPI < Sinatra::Base
     return_doc.to_json
   end
 
+  # Which organization have holdings that:
+  # * are in the same cluster as item_id and
+  # * has condition: BRT, status: LM, or both
+  # Only report organizations with a count greater than zero.
+  get "/v1/item_held_by_brlm" do
```
aelkiss, line 53: I would like to ponder whether it makes sense for this to be a separate route or a parameter on the existing route.

aelkiss: I think that it probably does make sense to try to integrate this into `/v1/item_held_by`. Perhaps a `limit` parameter, e.g. `limit=brlm`; maybe `limit=current` (i.e. not withdrawn) for resource sharing. Ultimately, I think this makes the API more open to future changes in functionality (including for resource sharing) without needing to change as much in the code or to the interface of the API itself.

Maybe this eventually looks something like:

```ruby
  get "/v1/item_held_by" do
    # common stuff to get parameters & cluster
    validate_params(params: params, required: ["item_id"])
    # plus whatever you need to get params["limit"]   

    cluster = Clusterable::HtItem
      .find(item_id: params["item_id"])
      .cluster

    organizations = case params["limit"]
    when nil
       cluster.organizations_in_cluster
    when "brlm"
       orgs_with_brlm_holdings(cluster)
    when "current"
       orgs_with_current_holdings(cluster) 
    else
       raise "unknown limit parameter"
    end
    
   { "organizations" => organizations }.to_json
  end 
```

spec/api/holdings_api_spec.rb
```diff
@@ -213,17 +213,17 @@ def parse_body(call)
 
   describe "v1/item_held_by" do
     it "returns an application error if the item_id does not match an ht_item" do
-      response = parse_body(v1(url_template("item_held_by", item_id: item_id_1, organization: "umich")))
+      response = parse_body(v1(url_template("item_held_by", item_id: item_id_1)))
```
aelkiss, line 216: I think I made this change on another branch as well so you may end up with a merge conflict

spec/api/holdings_api_spec.rb
```diff
@@ -232,8 +232,62 @@ def parse_body(call)
...
       )
-      response = parse_body(v1(url_template("item_held_by", item_id: item_id_1, organization: "umich")))
+      response = parse_body(v1(url_template("item_held_by", item_id: item_id_1)))
       expect(response["organizations"].sort).to eq ["smu", "umich"]
     end
   end
+
+  describe "v1/item_held_by_brlm" do
+    it "returns an application error if the item_id does not match an ht_item" do
+      response = parse_body(v1(url_template("item_held_by_brlm", item_id: item_id_1)))
```
aelkiss, line 242: Would it make sense to further extract `parse_body(v1(url_template(` to de-duplicate that in the tests?

spec/api/holdings_api_spec.rb
```diff
@@ -232,8 +232,62 @@ def parse_body(call)
...
+      expect(response["application_error"]).to eq "no matching data"
+    end
+    it "reports nothing if there are no holdings" do
+      load_test_data(htitem_1)
+      response = parse_body(v1(url_template("item_held_by_brlm", item_id: item_id_1)))
+      expect(response["organizations"]).to eq []
+    end
+    it "reports nothing if all holdings are status:CH & condition:''" do
```
aelkiss, line 250: I hate to gripe about style (I'm usually fine with whatever standardrb does), but I find it makes it much easier to read the tests when the `it/end` blocks have a line of whitespace in between them...

spec/api/holdings_api_spec.rb
```diff
@@ -232,8 +232,62 @@ def parse_body(call)
...
+      load_test_data(htitem_1)
+      load_test_data(build(:holding, organization: "a", ocn: htitem_1.ocns.first, status: "CH", condition: ""))
+      response = parse_body(v1(url_template("item_held_by_brlm", item_id: item_id_1)))
+      expect(response["organizations"]).to eq []
+    end
+    it "reports nothing if all holdings are status:WD & condition:''" do
```
aelkiss, line 256: All the tests here make sense, but it may be ultimately preferable to move some of them either to tests for `Clusterable::Holding#brt_lm_access?` or to `Cluster#access_counts`. (Right now it doesn't appear we have any specific tests for those, so it makes sense to be comprehensive here.)

aelkiss (commented):
Everything makes sense here although there are a few possible cleanups/improvements I've suggested.

The one potentially blocking thing here is that I do think my preference would be to integrate this functionality into `/v1/item_held_by`, but I'm definitely open to other points of view if there are reasons to prefer separate API endpoints.

lib/api/holdings_api.rb
moseshll: For testing Holdings.pm having an additional route meanse another endpoint to mock. That's not the main consideration of course. I would lean toward lumping into one route since it's the "same kinda" information being requested.

lib/api/holdings_api.rb
```diff
@@ -46,6 +46,26 @@ class HoldingsAPI < Sinatra::Base
...
+    # ArgumentError if missing item_id.
+    validate_params(params: params, required: ["item_id"])
+    item_id = params["item_id"]
+
+    ht_item = Clusterable::HtItem.find(item_id: item_id)
+    cluster = ht_item.cluster
+    # cluster.access_counts is a hash of {organization => count}
+    nonzero_access_counts = cluster.access_counts.select { |org, count| count > 0 }
```
moseshll, line 61: Let me gripe briefly about `access_counts` as being a horrible name (I know we're at the mercy of the schema). Is there an argument to be made for doing an `alias_method` on it so we can actually mention `brlm` or `brittle_lost_missing` in the code itself?

moseshll:
I generally +1 AE's comments, with one dissent: the proposed `item_held_by` parameter `limit` which immediately suggests "limit the number of results to this page size." `category` would be an improvement and there are surely others that would be better yet.

spec/api/holdings_api_spec.rb
mwarin: added vertival space between tests

aelkiss:
> I generally +1 AE's comments, with one dissent: the proposed `item_held_by` parameter `limit` which immediately suggests "limit the number of results to this page size." `category` would be an improvement and there are surely others that would be better yet.

Agreed, I think we could find a better name for that parameter.

lib/api/holdings_api.rb
aelkiss: Now would be the time to change it if we want to, and I would agree it's worth considering.

spec/api/holdings_api_spec.rb
mwarin: simplified helper methods somewhat

spec/api/holdings_api_spec.rb
mwarin: added test for `Clusterable::Holding#brt_lm_access?`

lib/api/holdings_api.rb
```diff
@@ -36,13 +36,21 @@ class HoldingsAPI < Sinatra::Base
...
-    return_doc = {
-      "organizations" => cluster.organizations_in_cluster
-    }
+    organizations = case constraint
+    when nil
+      cluster.organizations_in_cluster
+    when "brlm"
+      cluster.access_counts.select { |org, count| count > 0 }.keys.sort
+    else
+      raise ArgumentError, "Invalid constraint #{constraint}"
```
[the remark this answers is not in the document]
mwarin: yep, removed

spec/api/holdings_api_spec.rb
mwarin: added tests for `Cluster#access_counts`

lib/api/holdings_api.rb
mwarin: I don't have any strong feelings here, and I think this can be easily refactored if/when we decide to add `alias_method :ham_sandwich, :access_counts ` or whatever. So I'm not going to do anything.

aelkiss (approved):
Looks good to me now.

moseshll (approved):
Happy with this, APPROVE

PR #381, opened 2025-05-02
attempting to remedy some scrub slowness

mwarin:
Adding `strategy: :filter` to the `paged_each` in `Scrub::PreLoadBackup`


aelkiss (approved):
Just putting this here for reference https://sequel.jeremyevans.net/rdoc/classes/Sequel/Dataset.html#method-i-paged_each

PR #382, opened 2025-05-07
ETT-256: log requests to stdout; don't dump exceptions

aelkiss:
This enables logging for general requests and suppresses backtraces (which previously were saved even in our `error` handlers)

It might still be nice to have backtraces for "unexpected" errors, but I think it isn't critical, and the error does still get logged (just not the full backtrace)


moseshll (approved):
Tested and verified to work as advertised vs the default branch. I note that the Docker healthcheck now spams the log, which is fine with me and I can suppress it if needed to cut the chatter.

PR #386, opened 2025-05-15
ETT-193: (Holdings API) determine if an item is held by an institution and is brittle/lost/missing

mwarin:
Updated `/v1/item_access` to always provide a `brlm_count` alongside `copy_count`. Does not need `constraint=brlm`. This, being one of many ways to do it, seemed like a good compromise and was deemed acceptable by @moseshll .

An unrelated change (accounting for most of the lines in the diff) was implicitly setting the API version to 1 in all the test calls.

### DIFF
--- a/lib/api/holdings_api.rb
+++ b/lib/api/holdings_api.rb
@@ -22,16 +22,17 @@ class HoldingsAPI < Sinatra::Base
 
     organization = params["organization"]
     item_id = params["item_id"]
-
     ht_item = Clusterable::HtItem.find(item_id: item_id)
     overlap_record = Overlap::ClusterOverlap.overlap_record(organization, ht_item)
 
     return_doc = {
+      "brlm_count" => overlap_record.access_count,
       "copy_count" => overlap_record.copy_count,
       "format" => ht_item.cluster.format,
       "n_enum" => ht_item.n_enum,
       "ocns" => ht_item.ocns.sort
     }
+
     return_doc.to_json
   end
 
--- a/spec/api/holdings_api_spec.rb
+++ b/spec/api/holdings_api_spec.rb
@@ -19,7 +19,9 @@ RSpec.describe "HoldingsApi" do
   # ht_item 2 is an mpm
   let(:item_id_2) { "test.123456" }
   let(:ocn2) { [123, 456] }
-  let(:htitem_2) { build(:ht_item, :mpm, item_id: item_id_2, ocns: ocn2, enum_chron: "v.1-5 (1901-1905)") }
+  let(:enum_chron) { "v.1-5 (1901-1905)" }
+  let(:n_enum) { "1-5" }
+  let(:htitem_2) { build(:ht_item, :mpm, item_id: item_id_2, ocns: ocn2, enum_chron: enum_chron) }
 
   # ht_item 3 is an spm with 3 ocns
   let(:item_id_3) { "test.123456789" }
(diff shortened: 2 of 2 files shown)


lib/api/holdings_api.rb
```diff
@@ -22,16 +22,17 @@ class HoldingsAPI < Sinatra::Base
 
     organization = params["organization"]
     item_id = params["item_id"]
-
     ht_item = Clusterable::HtItem.find(item_id: item_id)
     overlap_record = Overlap::ClusterOverlap.overlap_record(organization, ht_item)
 
     return_doc = {
+      "brlm_count" => ht_item.cluster.access_counts[organization],
```
mwarin, line 29: This is the main change, adding `brlm_count` to the return doc, regardless of any `constraint` parameter.

spec/api/holdings_api_spec.rb
```diff
@@ -223,28 +220,48 @@ def parse_response(...)
...
 
       # as set up above, upenn's holdings (empty enumchron) matches on ht1 both ht2
-      response = parse_response("item_access", v: 1, organization: holding.organization, item_id: ht1.item_id)
+      response = parse_response("item_access", organization: holding.organization, item_id: ht1.item_id)
       expect(response["copy_count"]).to eq 1
-      response = parse_response("item_access", v: 1, organization: holding.organization, item_id: ht2.item_id)
+      response = parse_response("item_access", organization: holding.organization, item_id: ht2.item_id)
       expect(response["copy_count"]).to eq 1
     end
+
```
mwarin, line 228: These are the added tests for `brlm_count`.

spec/api/holdings_api_spec.rb
```diff
@@ -223,28 +220,48 @@ def parse_response(...)
...
+    it "brlm_count indicates number of BRT/LM holdings by organization" do
+      load_test_data(
+        htitem_1,
+        build(:holding, organization: "umich", ocn: ocn)
+      )
+
+      # brlm_count if given &constraint=brlm, default 0
+      response = parse_response("item_access", item_id: item_id_1, organization: "umich", constraint: "brlm")
```
moseshll, line 236: You can get rid of the three instances of `constraint: "brlm"` from the API call.

moseshll (approved):
One recommended change (unneeded `constraint` kwarg) noted in the inline comment. Otherwise I'm quite happy with this.

spec/api/holdings_api_spec.rb
mwarin: Yep. Good catch. Removed & updated comments.

spec/api/holdings_api_spec.rb
aelkiss: We should probably have a test for an MPM...

aelkiss (changes requested):
I like the change for making the version fixed in the spec file.

We do need to make sure that the BRLM stuff works as expected for MPMs, which should be straightforward to change in holdings_api.rb but will need another test in holdings_api_spec.rb.

lib/api/holdings_api.rb
mwarin: Yep, old thinking on my behalf, I keep forgetting that the overlap classes exist. Fixed.

spec/api/holdings_api_spec.rb
mwarin: Yep. Added.

aelkiss (approved):
Looks good to me now, thanks.

I'll briefly investigate whether we can remove the count methods in Cluster.

mwarin:
> Looks good to me now, thanks.
> 
> I'll briefly investigate whether we can remove the count methods in Cluster.

Sorry, do you want me to wait with the merge until that investigation has concluded?

aelkiss:
> > Looks good to me now, thanks.
> > I'll briefly investigate whether we can remove the count methods in Cluster.
> 
> Sorry, do you want me to wait with the merge until that investigation has concluded?

Just to close the loop here: I didn't need to wait on the merge. `SerialOverlap` and `SinglePartOverlap` both use the `copy_counts` methods, etc, so we can't remove them from `Cluster`.

PR #387, opened 2025-05-20
ETT-183: mapto_instid

aelkiss:
When we query for the value in mapto_instid (here, "stanford_mapped"), we should get the holdings for stanford.

We could have some additional tests, e.g:

* we should get the copy count we expect when there are multiple institutions with the same mapto_inst_id

### DIFF
--- a/lib/api/holdings_api.rb
+++ b/lib/api/holdings_api.rb
@@ -23,11 +23,17 @@ class HoldingsAPI < Sinatra::Base
     organization = params["organization"]
     item_id = params["item_id"]
     ht_item = Clusterable::HtItem.find(item_id: item_id)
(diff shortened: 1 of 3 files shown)


moseshll (approved):
The code checks out fine for me. In tracing thru some of the `mapto` logic, I'm troubled that in the production and dev DBs we allow `NULL` for `mapto_inst_id`. If we want to keep that, then I would advocate for `DataSources::HTOrganizations` interpreting a NULL as "self", currently that is not the behavior. Better, though, to tighten up the DB schema and make sure Otis is well behaved.

aelkiss:
Agreed. I think we can update the schema, as there are no organizations that have a blank mapto_inst_id.

PR #388, opened 2025-05-21
ETT-185: expire cache of `ht_institutions`, `ht_billing_members`, `ht_collections`

mwarin:
What I've got:

* failing test for reloading cached service objects


lib/services.rb
```diff
@@ -49,3 +49,12 @@
   end
 end
 Services.register(:redis_config) { raise "Redis not configured" }
+Services.register(:cache_timestamp) { Time.now.to_i }
+
+# Certain elements that need to be reloadable go in here.
+def Services.reload_cached!
```
aelkiss, line 55: This seems like a pretty clean way of doing this

lib/api/holdings_api.rb
```diff
@@ -15,6 +15,16 @@ class HoldingsAPI < Sinatra::Base
     "pong"
   end
 
+  post "/v1/reload_cached" do
```
aelkiss, line 18: The issue here is that because there are many workers and many threads for the holdings API, calling this endpoint will reload the cache in _one_ of them, but you can't be guaranteed to reload it for _all_. I'm a bit torn on exactly how best to do it.

spec/services_spec.rb
```diff
@@ -0,0 +1,109 @@
+# frozen_string_literal: true
+
+# require "spec_helper"
+require "services"
+
+RSpec.describe Services do
+  describe "#cache_timestamp" do
+    it "is set and recent" do
```
aelkiss, line 8: This is probably fine? We could maybe use https://github.com/travisjeffery/timecop if need be but I'm not sure exactly what we'd do with it.

spec/services_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+        content_provider_cluster: "foo",
+        responsible_entity: "foo",
+        original_from_inst_id: "foo",
+        billing_entity: "foo"
+      )
+      # Make sure we actually got it into the db...
+      expect(Services.ht_db[:ht_collections].where(collection: "foo").count).to eq 1
+
+      # ... but it is not yet reflected in the cached object.
+      expect { Services.ht_collections["foo"] }.to raise_error(KeyError, expected_error)
```
aelkiss, line 38: I'm wondering if it should re-query once on `KeyError`?

lib/data_sources/ht_collections.rb
```diff
@@ -50,11 +56,21 @@ def load_from_db
...
-    # TODO - should reload periodically -- when a new job starts, at least.
-    # Re-checking the database on a miss might also be sufficient.
-    def [](collection)
+    def load_data(collections)
+      @collections = data_from_db
+      @cache_timestamp = Time.now.to_i
+    end
+
+    def collection_info(collection:, retry_it: true)
+      if Time.now.to_i - @cache_timestamp > 3600
```
moseshll, line 65: I suggest exporting `3600` as a named constant, since we're duplicating the magic number (in `3601` form) in the tests. This applies to collections and institutions.

moseshll: I found to my surprise that you _can_ say `Timecop.travel(Time.now + described_class::CACHE_TTL + 1)`

mwarin: Excellent suggestion. Implemented in [75b14a2](https://github.com/hathitrust/holdings-backend/pull/388/commits/75b14a2de07ceef4d68292b13a395f3083ee1dea)

PR #390, opened 2025-05-28
main is now mariadb

aelkiss:


aelkiss:
Most of the uncovered stuff vs. main is related to shared print.

We should probably look at the remaining skipped tests and either re-enable them or remove them. I'll make an ETT issue for that.

PR #391, opened 2025-05-29
ETT-185: check for non-existing organizations in holdings api

mwarin:
Fix for https://hathitrust.atlassian.net/browse/ETT-185?focusedCommentId=384190

### DIFF
--- a/lib/api/holdings_api.rb
+++ b/lib/api/holdings_api.rb
@@ -19,8 +19,7 @@ class HoldingsAPI < Sinatra::Base
   get "/v1/item_access" do
     # ArgumentError if missing organization / item_id.
     validate_params(params: params, required: ["organization", "item_id"])
-
-    organization = params["organization"]
+    organization = organization_exists?(params["organization"])
     item_id = params["item_id"]
     ht_item = Clusterable::HtItem.find(item_id: item_id)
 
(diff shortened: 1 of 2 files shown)


aelkiss (changes requested):
@angelinanz confirms that lost/missing items are not eligible for resource sharing, so I think we need to expose this as "current_holdings_count".

https://docs.google.com/document/d/1knKPyQ2AV173K2pxsDIZFlNa3FymB_aLoE1Aix0Jmgo/edit?tab=t.0

lib/api/holdings_api.rb
```diff
@@ -122,4 +121,15 @@ def validate_params(params:, required:)
       raise ArgumentError, argument_errors.join("; ")
     end
   end
+
+  # Should be used in any route that takes an organization as an argument,
+  # to check that it is either an existing organization or a mapto.
+  def organization_exists?(organization)
```
aelkiss, line 127: Would it make sense for `HtOrganizations` to have a `has_organization?` method? Obviously we need to raise the `ArgumentError` here.

aelkiss (commented):
I think this should solve the issue, but I wonder if it might make sense to move some of the logic to `HtOrganizations`. I don't feel strongly though.

aelkiss (approved):
I can't seem to remove my erroneous review. Hopefully the "Approve" will cancel it out.

PR #395, opened 2025-06-12
ETT-344, special budget diff

mwarin:
temporary-ish changes to add comparison of latest cost report against the report used to set the most recent budget

### DIFF
--- a/bin/cost_changes.sh
+++ b/bin/cost_changes.sh
@@ -49,7 +49,6 @@ function diff_percent_totals () {
 
 # The last column from the other 3 output files.
 function hilites () {
-    echo -e "member_id\ttotal\tdiff\tdiff_pct";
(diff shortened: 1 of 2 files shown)


bin/cost_changes.sh
```diff
@@ -69,3 +68,10 @@ echo $totals_file;
 echo $diff_file;
 echo $diffp_file;
 echo $hilites_file;
+
+# Special handling until the budget is set in 2025:
+prev_budget_report=`echo "$all_reports" | grep -Po '\S+costreport_20240801.tsv'`
```
mwarin, line 73: These lines generate the extra data and puts them in 2 separate files that can be pasted into the google sheet version of the report.

mwarin:
I've already run the cost report with this version of `cost_changes.sh` twice in production, so I'm happy with it and I'm going to merge it to main.

PR #406, opened 2025-11-25
ETT-1150: Consistently compute overlap for MPMs

aelkiss:
This fixes a bug in the holdings API (where info on `item_held_by` for multiparts was completely wrong) and makes overlap computation for multiparts consistent between the cost report, holdings API, and overlap reports. See notes in the individual commits & inline.

One potential pitfall is that this involves significantly more object creation than the existing method in `HtItemOverlap`. However, it _shouldn't_ involve additional database queries, which is typically the bulk of the time for any operation involving holdings; I think we are already doing a pretty good job of memoizing things in `Cluster` to prevent duplicate computations.

If we see that cost reports, catalog indexing, etc get significantly slower though we can try to do some additional profiling/optimization which could involve changing things that currently use `HtItemOverlap` to instead use `ClusterOverlap` directly, and/or additional caching of `Cluster`s.

### DIFF
--- a/lib/api/holdings_api.rb
+++ b/lib/api/holdings_api.rb
@@ -53,16 +53,18 @@ class HoldingsAPI < Sinatra::Base
     ht_item = Clusterable::HtItem.find(item_id: item_id)
     cluster = ht_item.cluster
 
-    organizations = case constraint
+    overlaps = Overlap::ClusterOverlap.new(cluster).for_item(ht_item)
+
+    relevant_overlaps = case constraint
     when nil
-      cluster.organizations_in_cluster
+      overlaps
     when "brlm"
-      cluster.access_counts.select { |org, count| count > 0 }.keys.sort
+      overlaps.select { |o| o.access_count > 0 }
     else
       raise ArgumentError, "Invalid constraint."
     end
 
-    return_doc = {"organizations" => organizations}
+    return_doc = {"organizations" => relevant_overlaps.map(&:org)}
     return_doc.to_json
   end
 
--- a/spec/frequency_table_spec.rb
+++ b/spec/frequency_table_spec.rb
@@ -193,15 +193,21 @@ RSpec.describe FrequencyTable do
       end
 
       context "with Serials" do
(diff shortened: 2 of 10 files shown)


lib/api/holdings_api.rb
```diff
@@ -53,16 +53,20 @@ class HoldingsAPI < Sinatra::Base
     ht_item = Clusterable::HtItem.find(item_id: item_id)
     cluster = ht_item.cluster
 
-    organizations = case constraint
+    overlaps = Overlap::ClusterOverlap.new(cluster).for_item(ht_item)
```
aelkiss: Before, this was just completely omitting the case of MPMs.

spec/overlap/multi_part_overlap_spec.rb
```diff
@@ -20,7 +27,6 @@
     build(:ht_item,
       :mpm,
       enum_chron: "",
-      n_enum: "",
```
aelkiss: n_enum is a computed field; setting it here directly caused some unexpected issues.

spec/overlap/multi_part_overlap_spec.rb
```diff
@@ -95,10 +100,33 @@
       expect(overlap.matching_holdings.count).to eq(1)
     end
 
-    it "does not find holdings with the wrong enum" do
```
aelkiss: The previous test had one htitem and a holding with a different enumchron from the holding. That now *matches* (to avoid matches, we need to have *multiple* holdings, at least one of which matches a *different* htitem in the same cluster/record)

spec/frequency_table_spec.rb
```diff
@@ -193,42 +193,36 @@
       end
 
       context "with Serials" do
+        # umich deposited one htitem,
```
aelkiss: None of the logic of this test has changed, but the use of fake organizations made it hard to read and apparently brittle.

moseshll (approved):
This all looks fine to me. APPROVE

PR #425, opened 2026-04-23
Add slack notification on successful holdings load

eumalin:
Posts a slack message via incoming webhook when a holdings load completes successfully.

Message format: `Holdings load complete for org (type) - x records loaded, y old records removed`

If the slack webhook url is not set, it will do nothing.

I added an integration test for the success case, and unit tests for the unconfigured (no webhook URL) and network failure cases, since that's cheaper than wiring them through the full scrub pipeline in the integration spec.

### DIFF
--- a/lib/loader/holding_loader.rb
+++ b/lib/loader/holding_loader.rb
@@ -3,6 +3,7 @@
 require "loader/loaded_file"
 require "scrub/pre_load_backup"
 require "utils/file_transfer"
+require "utils/slack_notifier"
 
 module Loader
   # Constructs batches of Holdings from incoming file data
@@ -91,6 +92,20 @@ module Loader
         organization: options["organization"],
         mono_multi_serial: options["mono_multi_serial"]
       )
+
+      # Count only newly loaded records (delete_flag=0); must run before delete_marked! clears the old ones.
+      new_count = Services.holdings_db[:holdings]
+        .where(organization: options["organization"],
+               mono_multi_serial: options["mono_multi_serial"],
+               delete_flag: 0)
+        .count
+
+      Utils::SlackNotifier.post(
+        "Holdings load complete for *#{options["organization"]}* " \
+        "(`#{options["mono_multi_serial"]}`) — " \
+        "#{new_count} records loaded, #{pre_load_backup.marked_count} old records removed."
+      )
+
       if pre_load_backup.marked_count > 0
         Services.logger.info "deleting old #{options["mono_multi_serial"]} records for #{options["organization"]}"
         pre_load_backup.delete_marked!
--- a/lib/loader/holding_loader.rb
+++ b/lib/loader/holding_loader.rb
@@ -94,10 +94,10 @@ module Loader
       )
 
       # Count only newly loaded records (delete_flag=0); must run before delete_marked! clears the old ones.
-      new_count = Services.holdings_db[:holdings]
+      new_count = Clusterable::Holding.table
         .where(organization: options["organization"],
-               mono_multi_serial: options["mono_multi_serial"],
-               delete_flag: 0)
+          mono_multi_serial: options["mono_multi_serial"],
+          delete_flag: 0)
         .count
 
       Utils::SlackNotifier.post(
--- a/spec/integration_spec.rb
+++ b/spec/integration_spec.rb
@@ -251,6 +251,8 @@ RSpec.describe "phctl integration" do
   end
 
   describe "Scrub" do
+    after { Settings.slack_webhook_url = nil }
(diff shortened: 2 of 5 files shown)


lib/loader/holding_loader.rb
```diff
@@ -91,6 +92,20 @@ def on_success(_status, options)
...
       )
+
+      # Count only newly loaded records (delete_flag=0); must run before delete_marked! clears the old ones.
+      new_count = Services.holdings_db[:holdings]
+        .where(organization: options["organization"],
+          mono_multi_serial: options["mono_multi_serial"],
+          delete_flag: 0)
+        .count
+
+      Utils::SlackNotifier.post(
```
aelkiss, line 103: It may be preferable to do the notification after `delete_marked!` runs - that can take a while as sometimes institutions submit millions of holdings, which can take quite a while to delete. (It would be good to figure out a better approach to this eventually, but it's not in scope now)

lib/loader/holding_loader.rb
```diff
@@ -91,6 +92,20 @@ def on_success(_status, options)
         organization: options["organization"],
         mono_multi_serial: options["mono_multi_serial"]
       )
+
+      # Count only newly loaded records (delete_flag=0); must run before delete_marked! clears the old ones.
+      new_count = Services.holdings_db[:holdings]
```
aelkiss: We should be able to say `Clusterable::Holdings.table` instead of `Services.holdings_db[:holdings]`

spec/integration_spec.rb
```diff
@@ -251,6 +251,8 @@ def cleanup(output)
   end
 
   describe "Scrub" do
+    after { Settings.slack_webhook_url = nil }
+
```
aelkiss: In general how we manage config in the app is pretty confusing with `Settings` vs. environment variables and where overrides for settings could come from.

What you have here works, but it might be preferable to use [climate_control](https://github.com/thoughtbot/climate_control) to provide a bit stronger guarantee that ensures the env var is modified only within the scope of the test

```ruby
let(:webhook_url) { "https://hooks.slack.com/services/TEST/WEBHOOK" }
around(:each) do |example|
  ClimateControl.modify(SLACK_WEBHOOK_URL: webhook_url)
    example.run
  end
end
```

@moseshll any thoughts? I remember this kind of thing (temporarily overriding something in `Settings`) has come up before, but I don't recall whether we landed on anything completely satisfactory

aelkiss (commented):
Overall tests make sense and I think it should work. I have a couple things to consider, and I do think we should wait to post the notification until after the cleanup is done.

spec/integration_spec.rb
eumalin: As I understand it, this pattern may also require that `Settings.reload!` is added before `example.run`. An example of this is here https://github.com/hathitrust/holdings-backend/blob/28f551fa2b7683d5f44a142136e614f583d0b3c6/spec/spec_helper.rb#L117-L120 `Settings.slack_webhook_url` holds a frozen value. `ClimateControl.modify` would have no effect on it, unless the reload is called. Is this correct?

Would it be cleaner to just have the slack notifier use `ENV["SLACK_WEBHOOK_URL"]` directly, so no reload is needed?

moseshll: No, we did not. We have a reasonable use case at https://github.com/hathitrust/hathitrust_catalog_indexer/blob/b860e21a555499df97e84d0ac6ae9ef36ad60884/spec/spec_helper.rb#L70 but it's not profoundly satisfying. Canister/Services becomes brittle in its own way when testing is involved. So wherever possible I try to use ClimateControl.

moseshll: One possibility is to use dependency injection in the notifier with something like
```
  def self.post(message, url: Settings.slack_webhook_url)
```
As a way to sidestep mucking around with ENV and Settings. I just gave this a try and I can get tests to pass, webmock's happy.

As you say, @eumalin values get frozen in the settings. I was going to suggest something like ```ENV value || Settings value``` but then just passing a optional URL with a default seems cleaner.

moseshll (approved):
I'm happy with the result here. APPROVE

spec/integration_spec.rb
eumalin: I tried ClimateControl but it didn't work cleanly because Settings caches config at load time, and reloading doesn't reliably reset values to nil. Dependency injection was simpler, unit tests just pass the URL directly, and no env or settings wrangling was needed.

kron-spar (approved):

aelkiss (approved):
Changes look good.

PR #427, opened 2026-04-28
Use correct logic for getting new record count

eumalin:
In #425, the logic to get the count for new records was not correct. It assumed that new records have the `delete_flag` set to 0, but the field is not included when writing holdings, so in the DB they are NULL.

The PR fixes the new count by subtracting the number of old records from the total.

Updated a test to verify that it will return 6 records (instead of 0 records in #425).

### DIFF
--- a/lib/loader/holding_loader.rb
+++ b/lib/loader/holding_loader.rb
@@ -93,12 +93,13 @@ module Loader
         mono_multi_serial: options["mono_multi_serial"]
       )
 
-      # Count only newly loaded records (delete_flag=0); must run before delete_marked! clears the old ones.
-      new_count = Clusterable::Holding.table
+      # Count newly loaded records by subtracting old (delete_flag=1) from total;
+      # must run before delete_marked! clears the old ones.
+      total_count = Clusterable::Holding.table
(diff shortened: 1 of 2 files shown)


lib/loader/holding_loader.rb
```diff
@@ -93,12 +93,13 @@ def on_success(_status, options)
...
       )
 
-      # Count only newly loaded records (delete_flag=0); must run before delete_marked! clears the old ones.
-      new_count = Clusterable::Holding.table
+      # Count newly loaded records by subtracting old (delete_flag=1) from total;
+      # must run before delete_marked! clears the old ones.
+      total_count = Clusterable::Holding.table
         .where(organization: options["organization"],
-          mono_multi_serial: options["mono_multi_serial"],
-          delete_flag: 0)
```
aelkiss, line 100: Looking at the actual database it looks like this could be `delete_flag: 0` or `delete_flag: nil` (i.e. NULL in the database). I think the approach you have here now is probably the safest; in the future, we could look at cleaning up the data, or considering if there's some better way to approach this problem in general.

aelkiss (approved):

PR #428, opened 2026-04-30
Add Slack notifications for scrub rejection

eumalin:
Follow-up for #425.

Adds failure/rejection notifications:
* `run_file` branches on error type. `MalformedFileError` gets rejected, everything else gets failed.
* `run` wraps `check_new_types` and notifies on `TypeCheckError` before re-raising.

For diff rejections, `err.message` already has filename, percentages, and line counts, so I didn't add a filename prefix.

### DIFF
--- a/lib/scrub/scrub_runner.rb
+++ b/lib/scrub/scrub_runner.rb
@@ -95,6 +95,7 @@ module Scrub
       Services.scrub_logger.error err.message
 
       # MalformedFileError message already contains the filename and diff details; other errors need err.class for triage.
+      # TODO: Consider adding err.class to MalformedFileError's Slack message for consistency with other error types.
       slack_msg = if err.is_a?(MalformedFileError)
         "Holdings scrub rejected for *#{organization}* — #{err.message}"
       else
--- a/lib/scrub/scrub_runner.rb
+++ b/lib/scrub/scrub_runner.rb
@@ -24,6 +24,7 @@ require "sidekiq_jobs"
 require "sidekiq/batch"
 require "utils/file_transfer"
 require "utils/line_counter"
+require "utils/slack_notifier"
 
 # Example:
 # runner = Scrub::ScrubRunner.new(ORG)
@@ -49,7 +50,16 @@ module Scrub
       Services.logger.info "Running org #{organization}."
       new_files = check_new_files
       Services.logger.info "Found #{new_files.size} new files: #{new_files.join(", ")}."
-      check_new_types(new_files) if type_check
+      if type_check
+        begin
+          check_new_types(new_files)
+        rescue Scrub::TypeCheckError => err
+          Utils::SlackNotifier.post(
+            "Holdings scrub rejected for *#{organization}* — #{err.message}"
+          )
+          raise
+        end
+      end
 
       new_files.each do |new_file|
         run_file(new_file)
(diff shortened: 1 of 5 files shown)


spec/scrub/scrub_runner_spec.rb
```diff
@@ -183,5 +202,49 @@ def count_loaded_files
...
+      expect(stub).to have_been_requested.once
+    end
+
+    it "posts a 'failed' Slack notification with error class on unexpected error" do
+      remote_d = DataSources::DirectoryLocator.new(Settings.remote_member_data, org1)
+      remote_d.ensure!
+      FileUtils.cp(fixture_file, remote_d.holdings_current)
+      remote_file = sr.check_new_files.first
+
+      allow_any_instance_of(Scrub::RecordCounter).to receive(:acceptable_diff?).and_raise(RuntimeError, "disk full")
```
aelkiss, line 228: This makes sense in terms of changing the behavior to raise a normally-unexpected error, since we don't really have any other way to inject that behavior.

aelkiss (commented):
The changes in `scrub_runner.rb` look good.

I think the stubs for `a_string_including("...").and(a_string_including("...")` are OK -- I think it's more readable than a regex, and I think captures the things we expect to see there without being overspecified.

In the specs I'd like to extract something for setting up the mocked Slack endpoint similar to `spec/support/mock_solr_response.rb`. I think this would look something like:

```ruby
RSpec.shared_context "with mocked slack API endpoint" do
  WEBHOOK_URL = "https://hooks.slack.invalid/services/TEST/WEBHOOK"

  def stub_slack_webhook(body)
    stub_request(:post, WEBHOOK_URL)
     .with(body: body)
    .to_return(status: 200)
  end

  around(:each) do |example|
    old_webhook_url = Settings.slack_webhook_url
    Settings.slack_webhook_url = WEBHOOK_URL
    example.run
    Settings.slack_webhook_url = old_webhook_url
  end
end
```

then any example groups where we need this would have

```ruby
include_context "with mocked slack API endpoint"
```

and can avoid the `allow(Settings).to receive(...)`, for example

```ruby
    it "posts a Slack notification on type check rejection" do
      remote_d = DataSources::DirectoryLocator.new(Settings.remote_member_data, org1)
      remote_d.ensure!
      FileUtils.cp(fixture_file, remote_d.holdings_current)
      load_test_data(build(:holding, organization: org1, mono_multi_serial: "mix"))

      stub = stub_slack_webhook(a_string_including("umich")
          .and(a_string_including("rejected"))
          .and(a_string_including("mismatch")))

      expect { sr.run }.to raise_error(Scrub::TypeCheckError)
      expect(stub).to have_been_requested.once
    end
```

Let me know if you want to give that a try; if you run into problems with it I can take a look.

lib/scrub/scrub_runner.rb
```diff
@@ -84,6 +94,15 @@ def run_file(member_submitted_file)
       Services.scrub_logger.error "Unexpected error. Please contact HathiTrust."
       Services.scrub_logger.error err.message
 
+      # MalformedFileError message already contains the filename and diff details; other errors need err.class for triage.
+      slack_msg = if err.is_a?(MalformedFileError)
```
aelkiss, line 99: Might point to a need to make `MalformedFileError` more consistent with the others; not sure it's worth fixing now; could consider it if we're working on error formatting in the future anyway. Maybe worth a note for a future TODO here or in `MalformedFileError`?

eumalin: Done, added a TODO comment.

eumalin:
@aelkiss Done, extracted to `mock_slack_webhook.rb` to match the `mock_solr_response.rb` pattern. That does make the tests cleaner.

The `around(:each)` sets the webhook URL for every test, including non-Slack tests that also trigger SlackNotifier.post. Those started failing with `NetConnectNotAllowedError` because WebMock had no stub for them, so I added a catch all stub in around `stub_request(:post, webhook_url).to_return(status: 200)`. The Slack-specific tests still check the message body as before.

spec/integration_spec.rb
```diff
@@ -290,6 +290,34 @@ def cleanup(output)
...
+        stub_request(:post, webhook_url)
+          .with(body: a_string_including("umich")
+            .and(a_string_including("rejected"))
+            .and(a_string_including("Diff too big"))
+            .and(a_string_including("umich_mon_full_20220101.tsv"))
+            .and(a_string_including("Line diff too great")))
+          .to_return(status: 200)
+        phctl(*%w[scrub umich])
+
+        expect(
```
aelkiss: @eumalin One other small cleanup thing -- above you have:

```ruby
expect(stub).to have_been_requested.once
```

whereas here we duplicate the expected body. Is it possible to use that here as well?

eumalin: Updated

aelkiss (approved):
Looks great now.

PR #429, opened 2026-04-30
set up vscode dev container with ruby-lsp, debugger, port forwarding

eumalin:
Sets up vscode devcontainer integration. That way we get syntax support, go to definition, and a working debugger by reopening in the container.

The main issue was that `workspaceFolder` pointed to `/workspaces/holdings-backend` while the container's `WORKDIR` is `/usr/src/app`.

I also added `postCreateCommand: bundle install` and a new `settings.json` telling ruby-lsp to use system Ruby and `standard` as the formatter.

The debug config needs `DATABASE_ENV=test` and the test DB vars set, otherwise the debugger exits immediately with a `DATABASE_ENV must be 'test'` error. I also removed the old "Debug test" config which used minitest syntax and never worked here.

I confirmed that port forwarding for the API server and Sidekiq web UI works in vscode when using dev containers.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,49 +2,30 @@
 // README at: https://github.com/devcontainers/templates/tree/main/src/docker-existing-docker-compose
 {
 	"name": "holdings-backend - dev",
-
-	// Update the 'dockerComposeFile' list if you have more compose files or use different names.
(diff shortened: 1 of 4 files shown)


aelkiss (approved):
I gave it a try in vscode on Linux and it worked for me with the given instructions!

PR #431, opened 2026-05-05
ETT-1485: Add estimate notification

eumalin:
Adds a Slack notification at the end of the estimate workflow so we know when a cost estimate finishes.

Example notification: `Estimate complete for *ocn_list.txt* — 1/6 OCNs matched, 0 IC items, estimated cost *$0.00*.`

### DIFF
--- a/lib/workflows/estimate.rb
+++ b/lib/workflows/estimate.rb
@@ -151,6 +151,13 @@ module Workflows
             "#{num_items_ic} (#{pct_items_ic.round(1)}%) are in copyright."
           ].join("\n")
         end
+
+        Utils::SlackNotifier.post(
+          "Estimate complete for *#{File.basename(ocn_file)}* — " \
+          "#{num_ocns_matched}/#{total_ocns} OCNs matched, " \
+          "#{num_items_ic} IC items, " \
+          "estimated cost *$#{sprintf("%0.2f", total_estimated_ic_cost)}*."
+        )
       end
 
       def pct_ocns_matched
--- a/spec/workflows/estimate_spec.rb
+++ b/spec/workflows/estimate_spec.rb
@@ -71,6 +71,20 @@ RSpec.describe Workflows::Estimate do
           1 (50.0%) are in copyright.
         EOT
       end
+
+      context "Slack notification" do
+        include_context "with mocked slack API endpoint"
+
+        it "posts a notification with the estimate results" do
+          stub = stub_slack_webhook(a_string_including("Estimate complete")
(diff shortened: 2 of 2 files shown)


spec/integration_spec.rb
```diff
@@ -130,6 +130,38 @@ def cleanup(output)
...
+        before do
+          mock_solr_oclc_search(solr_response_for(ic_item))
+        end
+
+        it "posts a Slack notification with IC item count and estimated cost" do
+          stub = stub_slack_webhook(a_string_including("Estimate complete")
+            .and(a_string_including("ocn_list.txt"))
+            .and(a_string_including("1/6 OCNs matched"))
+            .and(a_string_including("1 IC items"))
+            .and(a_string_including("$4999.50"))) # $9999 target_cost / 1 DB item * 0.5 h_share (depositing institution counts as one holder alongside prospective_member)
```
eumalin: @moseshll or @aelkiss does this math for $4999.50 estimate make sense in the test? I'm new to the code and would like to confirm that it looks correct.

moseshll: Gosh, I think so. It's hard to mock real-world situations here because of the sheer mass of data, so it passes the sniff test for me.

aelkiss: I'd compare to https://github.com/hathitrust/holdings-backend/blob/main/spec/workflows/estimate_spec.rb#L65-L72

In that case we have:
* two items loaded in the database, one of which is pd and one is ic
* target cost of $20
* the depositor of the one ic item is considered to hold it as well, so the cost breakdown is:

cost per volume: $10
cost for "0.5" h-share of the one in-copyright volume: $5

aelkiss: In the integration spec we have:

* effectively one item that exists (via the mocked solr response for the ic item)
* target cost of $9999 (which I think is a default in the test environment)
* depositing library pays for half the item ($4999.50)
* contributor pays $4999.50

So I think this is correct.

That said, it might make sense to move it to https://github.com/hathitrust/holdings-backend/blob/main/spec/workflows/estimate_spec.rb where we have a bit more control over the data that's loaded and an expectation already of what the output should look like. I think the reason we did it in the integration spec for loading holdings is that we didn't have a great place to put it otherwise there.

eumalin: Makes sense, I updated the test by moving it to estimate_spec.rb. That does seem to be a better place for it and is cleaner.

moseshll (approved):
I like it! APPROVE

aelkiss (approved):

PR #432, opened 2026-05-06
Set format explicitly for ht_items in tests when enumchron is specified

aelkiss:
This caused brittle tests, i.e. tests could fail if the format was randomly selected as 'SE' rather than 'BK', enumchron was provided, and we were relying on mpm matching behavior.

This attempts to set format explicitly for all ht_items and holdings when enumchron is specified.

Although it may not be specifically relevant or necessary for all tests, it should make it less confusing to debug issues if the data we load is more reflective of the way it would be in the actual system - i.e. when a holding has enumchron it has format mpm, and when an ht_item has enumchron it's reflected as mpm or ser.

I'm especially not sure about setting for holdings as well as ht_items because there isn't any behavior in the system for matching (so far as I'm aware) that relies on the format given for the holding. However, it is the way the data should be coming in, and should make it less confusing to debug if there are issues in the tests.

### DIFF
--- a/spec/api/holdings_api_spec.rb
+++ b/spec/api/holdings_api_spec.rb
@@ -209,10 +209,9 @@ RSpec.describe HoldingsAPI do
       expect(response["format"]).to eq "mpm"
     end
 
-    # Brittle test?
     it "correctly identifies mpms as not held when not held" do
-      ht1 = build(:ht_item, enum_chron: "v.1", billing_entity: "umich")
-      ht2 = build(:ht_item, enum_chron: "v.2", ocns: ht1.ocns, billing_entity: "umich")
+      ht1 = build(:ht_item, :mpm, enum_chron: "v.1", billing_entity: "umich")
+      ht2 = build(:ht_item, :mpm, enum_chron: "v.2", ocns: ht1.ocns, billing_entity: "umich")
       holding = build(:holding, enum_chron: "v.1", ocn: ht1.ocns.first, organization: "upenn")
       load_test_data(ht1, ht2, holding)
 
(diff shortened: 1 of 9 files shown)


aelkiss:
https://github.com/hathitrust/holdings-backend/pull/432/changes#diff-adcc7e76b9e0b535feef2192ca6b7b2e44a964a8dd3bf8e5d94645af33fa67ccR213-R215 was the test we had originally identified as brittle; an AI tool identified the case when ht1 and ht2 both get `bib_fmt: "SE"` as a cause of brittleness -- in that case the serials matching would be used, which does not take enumchron into account. There could potentially be other brittleness in the test as well (anecdotally it seems to have failed more often than we would expect if this was the only issue), but I concur that it would cause problems in this test.

eumalin:
These changes make sense to me.

spec/reports/cost_report_spec.rb has three holdings paired with mpm ht_items. They all start with `build(:holding`, and they are missing `mono_multi_serial`.

https://github.com/hathitrust/holdings-backend/blob/ht-item-factory-explicit-format/spec/reports/cost_report_spec.rb#L279-L287

What do you think about pinning those as well?

aelkiss:
> These changes make sense to me.
> 
> spec/reports/cost_report_spec.rb has three holdings paired with mpm ht_items. They all start with `build(:holding`, and they are missing `mono_multi_serial`.
> 
> https://github.com/hathitrust/holdings-backend/blob/ht-item-factory-explicit-format/spec/reports/cost_report_spec.rb#L279-L287
> 
> What do you think about pinning those as well?

Yes if we're going to set `mono_multi_ser` for holdings in tests, we should do that consistently wherever it's implied by the tests. I found a number of additional places as well.

eumalin (approved):
LGTM

aelkiss:
@moseshll Curious if you have an opinion on setting `mono_multi_serial` for holdings in the tests. Otherwise I'll go ahead and squash my commits & merge.

moseshll:
@aelkiss yeah I was looking at the places where it's unset (added an `after(:build)` in the factory) and I think if we really don't care about `mono_multi_serial` in the test we needn't make an explicit choice.

moseshll (approved):
I'd say we're good to go with what we have.

PR #433, opened 2026-05-06
ETT-1336 Holdings: scrub should ignore Dropbox subdirectories when lo…

moseshll:
…oking for TSV files

As mentioned in the originating ticket, we would like to stash originals (like XML that has been converted to TSV) in the yearly (e.g., `/2026`) directory that `ScrubRunner` searches for the latest batch of holdings files. Alas, `ScrubRunner` knows to skip `.log` files but not directories. Hence this additional check.

The result, log files and directories are invisible to scrub runner when it's scanning for current holdings files.

Commenting out the added scrub_runner.rb line 177 will cause the test to fail.

### DIFF
--- a/lib/scrub/scrub_runner.rb
+++ b/lib/scrub/scrub_runner.rb
@@ -173,6 +173,8 @@ module Scrub
       remote_files.each do |f|
         # Possibly filtering other files here.
         next if f["Name"].end_with?(".log")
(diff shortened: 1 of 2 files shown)


aelkiss (approved):
Looks like this was a pretty quick fix. 👍

eumalin (approved):

PR #435, opened 2026-05-07
ETT-1487 Add slack notification for cost reports

eumalin:
Posts a Slack notification when the cost report finishes.

Example notification: `Cost report complete — *1* volumes, target cost *$10.00*, cost per volume *$10.0000*, written to *costreport_20250115.tsv*.`


aelkiss (approved):
Looks good! The frequency table will be empty so the cost report will be empty as well, but that doesn't matter for the purpose of posting a message that it completed.

PR #437, opened 2026-05-08
ETT-315: Remove unused shared print code

aelkiss:
Remove shared print code from the holdings system.

This functionality was used in the MongoDB version of the system, but was not ported to the current MariaDB-based system. California Digital Library will be developing new functionality to handle most of the shared print workflows.

We may find some of this info useful for historical purposes especially if we eventually need to keep a copy of the shared print data and do reporting/overlap calculations against the HathiTrust corpus, but it's unlikely we would use the functionality exactly as it was.


eumalin (approved):
Some settings keys may still be present that are unused, which could be cleaned up later. Looks good to me.

moseshll (approved):
Pleasantly surprised at the amount of cruft being jettisoned outside `spec/` -- this all looks good.

aelkiss:
> Some settings keys may still be present that are unused, which could be cleaned up later. Looks good to me.

This is a good point. I removed a variety of other things related to shared print that are no longer used or relevant including the settings.

lib/phctl.rb
```diff
@@ -44,11 +39,6 @@ def concordance(filename_or_date)
       run_job(Jobs::Load::Concordance, filename_or_date)
     end
 
-    desc "cluster-file FILENAME", "Add a whole file of clusters in JSON format."
-    def cluster_file(filename)
```
aelkiss, line 48: This was a holdover from the Mongo implementation; not used except in one skipped test in phctl (which we should get working, but that should be part of ETT-314)

eumalin (approved):

PR #439, opened 2026-05-13
ETT-1488 Add slack notification for sidekiq failures

eumalin:
Adds Slack alerts to a separate channel when Sidekiq jobs fail, posting on the 3rd failure and every subsequent retry and again when the job exhausts all retries, with job args included so there's enough context to investigate. The threshold is `NOTIFY_AT_RETRY_COUNT` in `SidekiqFailureNotifier` and `SLACK_ALERTS_WEBHOOK_URL` needs to be set.

Example alert:
```
Sidekiq job failed (will retry): Jobs::Load::Holdings
Args: ["/umich_2024-01-15-120000.ndj"]
Error: Mysql2::Error: Lost connection to MySQL server during query
```

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -22,9 +22,7 @@ end
 Sidekiq.configure_server do |config|
   config.redis = Services.redis_config
 
-  config.server_middleware do |chain|
-    chain.add Utils::SidekiqFailureNotifier::Middleware
-  end
+  config.error_handlers << Utils::SidekiqFailureNotifier.method(:on_error)
 
   config.death_handlers << lambda { |job, exception|
     Utils::SlackNotifier.post(
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -1,5 +1,6 @@
 require "sidekiq"
 require "services"
+require "utils/sidekiq_failure_notifier"
 
 # Add anything here you need for sidekiq initialization
 
@@ -20,6 +21,17 @@ end
 
 Sidekiq.configure_server do |config|
   config.redis = Services.redis_config
+
+  config.server_middleware do |chain|
+    chain.add Utils::SidekiqFailureNotifier::Middleware
+  end
+
+  config.death_handlers << lambda { |job, exception|
+    Utils::SlackNotifier.post(
+      Utils::SidekiqFailureNotifier.death_message(job, exception),
+      url: Utils::SidekiqFailureNotifier.alerts_url
+    )
+  }
 end
 
 Sidekiq.configure_client do |config|
--- a/lib/utils/sidekiq_failure_notifier.rb
+++ b/lib/utils/sidekiq_failure_notifier.rb
@@ -4,7 +4,7 @@ require "utils/slack_notifier"
 
 module Utils
   module SidekiqFailureNotifier
-    # retry_count is nil on 1st attempt, 0 on 2nd, 1 on 3rd — notify on 3rd failure
+    # retry_count is nil on 1st attempt, 0 on 2nd, 1 on 3rd — notify on 3rd+ failure
     NOTIFY_AT_RETRY_COUNT = 1
 
     def self.alerts_url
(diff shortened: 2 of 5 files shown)


eumalin:
@moseshll @aelkiss thoughts on which retries we should notify about? This PR skips 1st and 2nd retries, notifies on 3rd retry, and also when the job is exhausted. How do you feel about including more retry notifications, or are we good with what we have in the PR?

lib/utils/sidekiq_failure_notifier.rb
```diff
@@ -0,0 +1,40 @@
...
+      "Sidekiq job failed (no more retries): #{job["class"]}\n" \
+        "Args: #{job["args"].inspect}\n" \
+        "Error: #{exception.class}: #{exception.message}"
+    end
+
+    class Middleware
+      def call(_worker, job, _queue)
+        yield
+      rescue => e
+        if job["retry_count"] == NOTIFY_AT_RETRY_COUNT
```
aelkiss: I'm inclined to use `>=` here given that some noisiness in the alerts channel is OK. I still think waiting until the 3rd failure to notify makes sense though as if it's really a transient issue there's no need to notify us, but I could be convinced that for now we should notify for any failure. @moseshll what do you think?

config/initializers/sidekiq.rb
```diff
@@ -20,6 +21,17 @@
 
 Sidekiq.configure_server do |config|
   config.redis = Services.redis_config
+
+  config.server_middleware do |chain|
```
aelkiss: Looking at https://github.com/sidekiq/sidekiq/wiki/Error-Handling I wonder if we could use `config.error_handlers` in a parallel way to `config.death_handlers` rather than needing to add middleware that always gets called?

spec/utils/sidekiq_failure_notifier_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      msg = described_class.death_message(job, error)
+      expect(msg).to include("Sidekiq job failed (no more retries)")
+      expect(msg).to include("Jobs::Load::Holdings")
+      expect(msg).to include("/data/umich.ndj")
+      expect(msg).to include("RuntimeError")
+      expect(msg).to include("disk full")
+    end
+  end
+
+  describe "Middleware" do
```
aelkiss: I wondered if it would be feasible to test a job directly with this, but it seems like it would be hard without running using actual redis, threading, etc, and IMO that's too much complexity for tests. While there is some testing support for testing jobs with middleware (https://github.com/sidekiq/sidekiq/wiki/Testing#testing-middleware), the problem is that we really want to test the retry mechanism.

The tests are good, but to test it end to end, I tried a manual test in development. I set the alert URL in `config/environments/development.local.yml` (this overrides `config/environments/development.yml`), then did

```bash
docker compose up -d processor
docker compose run dev ./phctl.sh pry
```

```ruby
Jobs::Load::Holdings.perform_async("test-ignore-nonexistent.tsv")
```

I gave this a try, and it all worked fine. In order to test the 'no more retries' logic I added `sidekiq_options retry: 3` in `Jobs::Load::Holdings` (in `sidekiq_jobs.rb`).

aelkiss (commented):
I tested end-to-end in development and it does work; questions are:
* when we want to alert
* if it's preferable to use `error_handlers` rather than the middleware

config/initializers/sidekiq.rb
eumalin: Yeah, agreed. I updated the code to use error_handlers.

lib/utils/sidekiq_failure_notifier.rb
eumalin: I updated code to use `>=`. Now it alerts on the 3rd failure and every subsequent retry.

moseshll (approved):
Nothing to add. APPROVE

aelkiss (approved):
Looks good with `error_handlers`. I tried again locally in dev with the webhook url set and saw the expected behavior with posts to slack.

PR #443, opened 2026-05-21
ETT-1447 phctl report costreport should accept ht item count & ht ite…

moseshll:
…m pd count

- Add `--ht_item_count` and `--ht_pd_item_count` parameters to `phctl report costreport`
- Add test for proper conversion to Integer and inclusion in params
- The ticket notes that writing this information to the frequency table file would be preferable; however, I am a bit leery of introducing a new version of the freqtable file format if not needed. Hence this simplistic solution.

### DIFF
--- a/lib/phctl.rb
+++ b/lib/phctl.rb
@@ -82,11 +82,13 @@ module PHCTL
   end
 
   class Report < JobCommand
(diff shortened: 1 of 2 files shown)


aelkiss (approved):
This seems fine. We can consider other ways to retain this data in the future if need be (if not in the frequency table, some other file we output with run info, or something), but even if we decide to do that I think this is a good addition.

PR #444, opened 2026-05-27
ETT-314: Enable or remove remaining skipped tests

aelkiss:
This removes the remaining (disabled) mongodb-specific code and gets the remaining skipped tests working. See detail in the commits. It's probably worth looking at each commit individually.

I think the code removal and the phctl inline tests are straightforward enough, but the others may be worth some thought -- specifically whether we need a test for `Overlap::Overlap` at all given that it's an abstract class, and whether we ought to implement a `Cluster.reload` method for use in the tests.

It's probably also worth reviewing the Jira ticket to make sure I didn't remove something that I should have updated/reimplemented instead.


spec/overlap/overlap_spec.rb
```diff
@@ -3,77 +3,28 @@
...
+  let(:holding2) do
+    build(:holding,
+      ocn: ht_item.ocns.first,
+      organization: "umich",
+      condition: "brt",
+      enum_chron: "V.1")
   end
+  let(:holding3) { build(:holding, ocn: ht_item.ocns.first, organization: "smu") }
 
-  describe "#lock_id" do
```
eumalin, line 54: Wondering if lock_id is tested anywhere else and if it makes sense to add some coverage for it

spec/cluster_spec.rb
```diff
@@ -218,14 +220,6 @@
         expect(c.copy_counts["umich"]).to eq(2)
         expect(c.copy_counts["smu"]).to eq(1)
       end
-
-      xit "cached counts should be invalidated when holdings/ht_items are changed" do
```
eumalin, line 222: Is the behavior still tested somewhere, or are we dropping it?

eumalin:
Really nice to see all this dead code gone.

spec/overlap/overlap_spec.rb
aelkiss: It is no longer needed, and the implementation had already been removed.

spec/cluster_spec.rb
aelkiss: The behavior was never fully implemented, even before the move from Mongo to MariaDB. See the notes from the relevant commit (fbc41e906c921cc49224e6d162184202479c9d8e):

> This test was skipped even pre-mariadb. In general we shouldn’t need
this – loading data in the mariadb world no longer relies on cluster
directly (since it can insert into underlying database tables);
reporting can happen on whatever snapshot of data we get (if it’s
changing underneath us we’ll likely get some kind of inconsistency in
the report output anyway).
>
> invalidate_cache was only used in the tests; we could restore it and
rename it reload which would be more similar to the ActiveRecord-style
interface and reflect how we were using it; still, I hesitate to add
functionality on Cluster that's only used in the tests.

Let me know what you think.

eumalin: Sounds good to me!

eumalin (approved):
LGTM

moseshll (approved):
It looks like we are covered.

I note that the `FileMutex` class, `Services.loading_flag`, and some `loading_flag_path` settings in `config/` appear to have been used only by the defunct hathifiles loader. Those can go, but maybe in a followup ticket, your call.

aelkiss:
> I note that the `FileMutex` class, `Services.loading_flag`, and some `loading_flag_path` settings in `config/` appear to have been used only by the defunct hathifiles loader. Those can go, but maybe in a followup ticket, your call.

I think they can go now.

PR #445, opened 2026-05-28
ETT-1480 Add scrub without load

eumalin:
Adds `phctl scrub_file ORG FILENAME`, a command to download and scrub a specific holdings file from Dropbox without loading it into the database. Useful for inspecting a file before committing to a full `phctl scrub` run.

The implementation adds `ScrubRunner#scrub_file`, which reuses the existing `download_to_work_dir` and `AutoScrub` machinery but stops after scrubbing, returning the output files.

`--force` was intentionally left off this command. `AutoScrub` accepts the parameter but only logs it, it doesn't change behavior, so exposing it would be misleading.

### DIFF
--- a/lib/phctl.rb
+++ b/lib/phctl.rb
@@ -209,6 +209,14 @@ module PHCTL
       Scrub::ScrubRunner.new(org, options).run
     end
 
+    desc "scrub_file ORG FILENAME", "Download and scrub a specific file for ORG from Dropbox without loading"
+    def scrub_file(org, filename)
+      Scrub::ScrubRunner.new(org, options).scrub_file(filename)
+    rescue => err
+      warn err.message
+      exit 1
+    end
+
     # report
     desc "report", "Generate a report"
     subcommand "report", Report
--- a/lib/scrub/scrub_runner.rb
+++ b/lib/scrub/scrub_runner.rb
@@ -66,6 +66,17 @@ module Scrub
       end
     end
 
+    def scrub_file(filename)
+      Dir.mktmpdir do |tmp_dir|
+        remote_file = File.join(remote_dir, filename)
+        file_transfer.download(remote_file, tmp_dir)
+        downloaded_file = File.join(tmp_dir, File.basename(filename))
+        scrubber = Scrub::AutoScrub.new(downloaded_file, force)
+        scrubber.run
+        scrubber.out_files
+      end
+    end
+
     def run_file(member_submitted_file)
       Services.logger.info "Running member_submitted_file #{member_submitted_file}"
       downloaded_file = download_to_work_dir(member_submitted_file)
--- a/spec/scrub/scrub_runner_spec.rb
+++ b/spec/scrub/scrub_runner_spec.rb
@@ -10,6 +10,7 @@ require "scrub/autoscrub"
 require "scrub/pre_load_backup"
 require "scrub/record_counter"
 require "scrub/scrub_runner"
+require "scrub/malformed_header_error"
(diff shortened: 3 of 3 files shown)


lib/scrub/scrub_runner.rb
```diff
@@ -66,6 +66,14 @@ def run
       end
     end
 
+    def scrub_file(filename)
+      member_submitted_file = {"Path" => filename, "Name" => filename}
+      downloaded_file = download_to_work_dir(member_submitted_file)
```
aelkiss: The only issue I have is whether by doing `download_to_work_dir` that would then (later) get listed by `check_new_files` or whether it would be ignored. 

I think we could test this by:
* doing `scrub_file`
* ensuring that the file we just checked is still included in `check_new_files`.

I suspect it would not be, because `download_to_work_dir` puts the file in `DataSources::DirectoryLocator.for(:local, organization)` and `check_old_files` checks that same location. That behavior is kind of an issue in general -- if there's an error scrubbing in `run_file`, there is an exception handler that removes the downloaded file from the local cache, thus getting it to show up as new again. But that behavior can be kind of brittle because if autoscrub gets interrupted at the wrong time or in a way that the exception handler doesn't run, the downloaded file can be in the local cache but it also wouldn't have finished loading. I don't know that we need to fix that in general now, but here in `scrub_file` we should probably just download to a temporary directory.

aelkiss (commented):
See comment about the potential issue with `download_to_work_dir`

spec/scrub/scrub_runner_spec.rb
```diff
@@ -153,6 +153,20 @@ def count_loaded_files
...
+      remote_d = DataSources::DirectoryLocator.new(Settings.remote_member_data, org1)
+      local_d.ensure!
+      remote_d.ensure!
+      FileUtils.cp(fixture_file, remote_d.holdings_current)
+      out_files = nil
+      expect { out_files = sr.scrub_file(fixture_file_name) }.not_to change { Clusterable::Holding.count }
+      expect(out_files).not_to be_empty
+      expect(out_files.all? { |f| File.exist?(f) }).to be true
+    end
+  end
```
moseshll, line 176: I'd like to see a malformed file context, using a fixture like `umich_mon_full_20220101_headerfail.tsv`. Running the command would raise `Scrub::MalFormedHeaderError` in this case.

lib/phctl.rb
```diff
@@ -219,6 +219,11 @@ def scrub(org)
       Scrub::ScrubRunner.new(org, options).run
     end
 
+    desc "scrub_file ORG FILENAME", "Download and scrub a specific file for ORG from Dropbox without loading"
+    def scrub_file(org, filename)
+      Scrub::ScrubRunner.new(org, options).scrub_file(filename)
```
moseshll, line 214: It might be worth `rescue`ing `Scrub::XXXException` and just log an error with the exception message, and suppress the stack trace.

moseshll (approved):
Aaron's comment is the one that should be addressed. Feel free to ignore the two suggestions I added, they're not that interesting.

lib/scrub/scrub_runner.rb
eumalin: I updated it so that it downloads to a temp directory

spec/scrub/scrub_runner_spec.rb
eumalin: Added a test.

aelkiss (approved):
Changes look good to me!

PR #447, opened 2026-06-03
ETT-1497 automate holdings XML load

moseshll:
- Add a `phctl convert-xml ORG` command which is intended to run without Sidekiq.
- Add `ExLibrisHoldings` class (subsequently renamed to `AlmaHoldings`) which takes care of download, extraction, conversion, and upload.
- Modify `ScrubRunner` to ignore `.xml` and `.tar.gz` files uploaded by member organization.
- On ETT-1497 I commented that `zinzout` and `Gem::Package::TarReader` might suffice for extraction of compressed uploads; however, extracting directly to a file did not seem straightforward and using the `tar` executable directly seems to give (barely) enough control to accomplish the goal with less code.
- I am undecided over whether to remove the `phctl parse parse-holdings-xml` command. That may be suitable for a follow-up.


lib/ex_libris_holdings.rb
```diff
@@ -0,0 +1,140 @@
...
+  # Remove temporary directory after upload
+  def cleanup
+    # force = true to ignore exceptions if tempdir hasn't been created yet
+    FileUtils.remove_entry(local_directory, true)
+    @local_directory = nil
+  end
+
+  # Upload the converted tsv files to remote directory
+  def upload(file:)
+    Utils::FileTransfer.new.upload(file, remote_directory)
```
eumalin: This makes a new `FileTransfer` but we already have `@file_transfer` from the constructor. Can just be `file_transfer.upload(...)`.

lib/alma_holdings.rb
```diff
@@ -0,0 +1,140 @@
...
+  # remote_directory is for testing/debugging
+  def initialize(organization:, remote_directory: nil)
+    @organization = organization
+    @remote_directory = remote_directory || DataSources::DirectoryLocator.for(:remote, organization).holdings_current
+    @file_transfer = Utils::FileTransfer.new
+  end
+
+  def run
+    Services.logger.info "Running org #{organization}."
+    if candidates.size.zero?
```
eumalin, line 35: `candidates` is called twice here (lines 35 and 41), so we hit the remote listing twice per run. Could we assign it to a local variable at the top of run and reuse it?

lib/ex_libris_holdings.rb
```diff
@@ -0,0 +1,140 @@
...
+    File.join(local_directory, File.split(remote_file).last)
+  end
+
+  # Extracts a single XML file from a .tar.gz submission and returns its path.
+  # If the input is an XML file, just return its path.
+  def extract(file:)
+    return file if file.end_with?(".xml")
+
+    # List the files in the archive. There better be only one.
+    file_list = `tar -tzPf #{file}`.split("\n")
```
eumalin: `file_list.first` comes from inside the archive, could be a potential security issue. If a filename with a single quote in it breaks the quoting on line 131, we get arbitrary shell execution.

What do you think about escaping on both values, or switching to `Open3.capture3` with an array so there's no shell involved at all?

aelkiss: We have an example of using `Open3` in `lib/utils/encoding.rb`, which I think has worked well enough. Alternatively (and maybe we want to do that anyway) we could validate that `file` matches an expected regular expression -- although maybe we don't have a spec for that?

lib/scrub/scrub_runner.rb
```diff
@@ -182,8 +182,11 @@ def check_new_files
       # Include in new_files only those remote_files whose name is not in old_files.
       new_files = []
       remote_files.each do |f|
-        # Possibly filtering other files here.
+        # Ignore logs we upload
         next if f["Name"].end_with?(".log")
+        # Ignore Alma XML, see `ExLibridHoldings` class
```
aelkiss: typo: `ExLibrisHoldings`

lib/alma_holdings.rb
```diff
@@ -0,0 +1,140 @@
...
+      Services.logger.warn "No XML files for #{organization} at #{remote_directory}, bailing out"
+      return
+    end
+
+    xml_files = []
```
aelkiss, line 40: Potentially confusing to call this `xml_files` since usually this will be a `.tar.gz`?

lib/ex_libris_holdings.rb
```diff
@@ -0,0 +1,140 @@
...
+# - Passes the extracted files to `ExLibrisHoldingsXmlParser`
+# - Uploads the resulting TSV files back up to the member's Dropbox
+#
+# See the `phctl convert-xml` command.
+# It is intended to be run in the foreground (no Sidekiq)
+#
+# Example:
+#   `ExLibrisHoldings.new(organization: "umich").run`
+
+class ExLibrisHoldings
```
aelkiss: Would it makes sense to call it `AlmaHoldings` (and rename ExLibrisHoldingsXmlParser) since this process is specific to files exported from Alma rather than other ExLibris ILS systems?

lib/alma_holdings.rb
```diff
@@ -0,0 +1,140 @@
...
+  def extract(file:)
+    return file if file.end_with?(".xml")
+
+    # List the files in the archive. There better be only one.
+    file_list = `tar -tzPf #{file}`.split("\n")
+    if file_list.count != 1
+      raise "unexpected contents for #{file}: #{file_list}"
+    end
+
+    # Avoid silly edge cases trying to find the correct name for the extracted XML files.
```
aelkiss, line 122: Is there a reason to have `mon_ser` in the output filename at all, if we can't guarantee/don't enforce what's in there anyway? We could also potentially use a sequential suffix rather than random.

lib/alma_holdings.rb
```diff
@@ -0,0 +1,140 @@
...
+  end
+
+  # Extracts a single XML file from a .tar.gz submission and returns its path.
+  # If the input is an XML file, just return its path.
+  def extract(file:)
+    return file if file.end_with?(".xml")
+
+    # List the files in the archive. There better be only one.
+    file_list = `tar -tzPf #{file}`.split("\n")
+    if file_list.count != 1
```
aelkiss, line 118: Is there a spec that requires only one compressed xml file per .tar.gz, or is that just observed in practice?

aelkiss (commented):
Overall flow of control & functionality look fine; tests look good. See comments inline about class & variable naming as well as some questions regarding handling the submitted files.

I pondered whether we would want to kick off autoscrub if everything works as expected, but IMO fine to leave that as a separate step for now. This definitely automates a somewhat laborious part of the process, and it should generally run quickly enough. There may be other things we need to do in between parsing these and loading anyway, like getting old/new counts, potentially dealing with change of formats, etc. Some of those we will likely automate later, but we aren't ready to do that now.

lib/alma_holdings.rb
moseshll: Just observed behavior plus hints from the Alma docs. I guess it doesn't matter if we iterate over multiple files in an archive. Since the parser does the mon/ser partitioning it's agnostic to number of input files. At this stage I'd rather be pedantic until we know we need to handle multiple files per tarball. Fortunately accommodating such a thing would be only a small code change (but would clutter up the tests/fixtures if not really needed.)

lib/alma_holdings.rb
moseshll: Purely for debugging. A hint as to which derivative belongs to which member-uploaded file. I like the idea of using a sequence.

lib/alma_holdings.rb
moseshll: Nope, these are actually the XML files. `extract` returns the full path to the extracted (or provided to Dropbox without compression) XML in the temp directory.

lib/ex_libris_holdings.rb
moseshll: Well that was a bit of a slog. Lots of AI hallucinations in the wild about what you can do with redirection to file in `Open3`. The latest commit has the tar manipulation isolated to private methods, but I think this deserves a new class.

moseshll: New class added, with a demonstration of hardened vs naive implementation.

lib/ex_libris_holdings.rb
moseshll: Yes, that is the final (I hope) commit. Except I didn't rename the test fixtures starting with `exlibris...` since that seems like a bit of a slog.

moseshll:
I believe I have responded to all of the feedback -- if you would be so kind as to check out the latest iteration things should be in fairly good shape.

lib/utils/tar.rb
```diff
@@ -0,0 +1,63 @@
...
+    TAR_EXE_PATH = "/usr/bin/tar"
+
+    def initialize(path:)
+      raise "not found: #{path}" unless File.exist?(path)
+
+      @path = path
+    end
+
+    # List the files in a .tar.gz file
+    def list
```
moseshll, line 21: Could memoize this, but as the class is currently used (outside of tests) it would gain us nothing.

lib/utils/tar.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    # Extract named file `file_name` to a new file at `destination_path`.
+    # Uses the path to `tar` explicitly to bypass shell expansion
+    # since `file_name` is tainted.
+    def extract(file_name:, destination_path:)
+      status = nil
+      stderr_s = ""
+
+      File.open(destination_path, "w") do |destination_file|
+        Open3.popen3(TAR_EXE_PATH, "-xzf", path, file_name, "--to-stdout") do |stdin, stdout, stderr, wait_thr|
```
aelkiss, line 38: This seems like a lot of complexity given the need, but it looks correct to me, and I don't know that with `Open3` there's anything simpler we could do. It's also something we can potentially extract and re-use in the future if we have similar needs (i.e. file redirection without needing to use the shell).

Another option could be doing e.g.

```ruby
system('tar','-xzf','--one-top-level',File.dirname(destination_path),path,file_name)
FileUtils.mv(File.dirname(destination_path) + '/' + File.basename(file_name), destination_path)
```

However that probably still gives the opportunity for a variety of other things to go wrong especially if `file_name` has questionable things in it, so I think the `Open3` version is preferable if it works reliably.

aelkiss (approved):
This all looks good to me now.

spec/alma_holdings_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+      local_dir = holdings.local_directory
+      holdings.cleanup
+      expect(File.directory?(local_dir)).to eq(false)
+    end
+  end
+
+  describe "#upload" do
+    it "copies a file from the local directory to the remote" do
+      local_file = File.join(holdings.local_directory, "umich_mon.tsv")
+      remote_file = File.join(holdings.local_directory, "umich_mon.tsv")
```
eumalin: looks like `remote_file` here points to `local_directory` instead of `remote_directory`, so the assertion just checks that the file we created still exists. `maybe File.join(remote_directory, "umich_mon.tsv")`?

eumalin:
The changes look good to me, thank you for addressing the comments!

spec/alma_holdings_spec.rb
moseshll: Yes, a couple patches to make sure this test is really doing what it should.

eumalin (approved):

lib/utils/tar.rb
moseshll: I was able to get this to work on the command line but could not get `system` to execute it without errors. Going to "can of worms" it in the interest of momentum. If we run into maintenance issues then I would consider gunzipping the file and using bundler's built in tar support. Now that we have the tests in place that would probably not be a big issue.

aelkiss: Yeah, wasn't suggesting that we change course now.

PR #448, opened 2026-06-08
Fix duplicate Slack notifications from overlap report workflow

eumalin:
We got two duplicate "Overlap report complete" Slack messages last week. Traced it to the `on_success` batch callback: `Writer#run` posts to Slack, then cleanup tries `FileUtils.remove_entry` on the working directory, which raised `ENOTEMPTY` from a filesystem race. Sidekiq retried the whole callback, `run` fired again, resulting in the second Slack message.

There are two fixes here with some help from AI.
* First, rescue `SystemCallError` around the cleanup so a race can't crash the callback and trigger a retry.
* Second, moved the Slack post out of `Writer#run` into a `notify` method that the callback calls after cleanup, so the ordering is correct going forward even if cleanup has issues.

Also applied the same split to `Estimate::Writer` since it had the identical pattern and the same risk.

Added tests for the rescue path, that `notify` is called from the callback, that `run` alone doesn't post to Slack, and also added a test that runs the full pipeline in test mode and asserts that exactly one Slack message is sent at the end.

### DIFF
--- a/lib/workflows/estimate.rb
+++ b/lib/workflows/estimate.rb
@@ -136,6 +136,9 @@ module Workflows
       end
 
       def run(output_filename = report_file(ocn_file))
+        if Dir.glob(File.join(working_directory, "*.estimate.json")).empty?
+          raise "No estimate files found in #{working_directory}"
+        end
         Services.logger.info "Target Cost: #{cost_report.target_cost}"
         Services.logger.info "Cost per volume: #{cost_report.cost_per_volume}"
 
@@ -151,7 +154,6 @@ module Workflows
             "#{num_items_ic} (#{pct_items_ic.round(1)}%) are in copyright."
           ].join("\n")
         end
-
       end
 
       def notify
--- a/lib/workflows/estimate.rb
+++ b/lib/workflows/estimate.rb
@@ -152,6 +152,9 @@ module Workflows
           ].join("\n")
         end
 
+      end
+
+      def notify
         Utils::SlackNotifier.post(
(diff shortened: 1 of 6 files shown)


aelkiss:
See my comment on https://hathitrust.atlassian.net/browse/ETT-1566 about the issue with the retry uploading an empty overlap report, because the source data had already been removed. I think the comment about

> rescue SystemCallError around the cleanup so a race can't crash the callback and trigger a retry.

should address this in part, but we may want to have the callbacks verify that there is *some* data there rather than accepting *no* data as OK.

Does `Reports::CostReport` also have a similar issue as the callback for `Workflow::CostReport`?

eumalin:
> See my comment on https://hathitrust.atlassian.net/browse/ETT-1566 about the issue with the retry uploading an empty overlap report, because the source data had already been removed. I think the comment about
> 
> > rescue SystemCallError around the cleanup so a race can't crash the callback and trigger a retry.
> 
> should address this in part, but we may want to have the callbacks verify that there is _some_ data there rather than accepting _no_ data as OK.
> 
> Does `Reports::CostReport` also have a similar issue as the callback for `Workflow::CostReport`?

Yeah, good call. I added a guard at the top of run in both `OverlapReport::Writer` and `Estimate::Writer` that raises if the working directory has no files to reduce. so if the callback gets retried after cleanup, it'll fail loudly instead of uploading an empty report.

`Reports::CostReport` happens to already raise in that scenario ("no frequency table provided"), so no change needed there.

aelkiss (approved):
This all looks good to me now!

PR #449, opened 2026-06-11
ETT-1476 - get item count when starting costreport workflow

aelkiss:
We added this functionality to specify it on DEV-1569 / commit 472dbdc because at the time holdings was only looking at the development hathifiles database. It is now using the production database, so that need goes away.

However, it still seems useful to have it query for item counts up front (rather than when starting the cost report part) because it reduces the window in which rights updates during the process could cause inconsistencies.

### DIFF
--- a/lib/phctl.rb
+++ b/lib/phctl.rb
@@ -124,17 +124,29 @@ module PHCTL
       end
     end
 
-    desc "costreport --ht-item-count NUM --ht-item-pd-count NUM (--chunk-size SIZE)", "Dump records from solr, split into chunks of chunk-size records, generate frequency tables for each chunk, sum the resulting frequency tables, and generate a cost report based on that table."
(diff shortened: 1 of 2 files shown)


lib/phctl.rb
```diff
@@ -124,17 +124,29 @@ def base_params
...
 
-    desc "costreport --ht-item-count NUM --ht-item-pd-count NUM (--chunk-size SIZE)", "Dump records from solr, split into chunks of chunk-size records, generate frequency tables for each chunk, sum the resulting frequency tables, and generate a cost report based on that table."
+    desc "costreport [--ht-item-count NUM --ht-item-pd-count NUM] (--chunk-size SIZE)", "Dump records from solr, split into chunks of chunk-size records, generate frequency tables for each chunk, sum the resulting frequency tables, and generate a cost report based on that table."
     option :ht_item_count, type: :numeric
     option :ht_item_pd_count, type: :numeric
     def costreport_workflow
+      ht_item_count = options[:ht_item_count]
+      ht_item_pd_count = options[:ht_item_pd_count]
+
+      if !ht_item_count || !ht_item_pd_count
```
aelkiss, line 134: There are certainly more concise ways to get this, but I think (especially because this operation can be a little on the slow side) it's worth doing the logging.

moseshll (approved):
As you say, very straightforward patch. APPROVE

PR #451, opened 2026-06-19
ETT-1479 Add preflight tooling

eumalin:
Adds `phctl holdings` subcommands for pre-flight checks before loading holdings.

New commands:

- `holdings count ORG` - breakdown by format (spm/ser/etc) with a total, from the DB
```
$ phctl holdings count umich
umich holdings by format:
  mpm:  12345
  ser:  67890
  spm:  1482903
Total: 1563138
```
- `holdings dir-counts PATH` - line counts for each .tsv in a directory, sorted alphabetically, skipping .log/.xml/.tar.gz and subdirectories, with a total
```
$ phctl holdings dir-counts umich/2025
umich_mon_full_20250902.tsv: 1482903
umich_ser_full_20250902.tsv: 67890
Total: 1550793
```
- `holdings file-count PATH` - line count for a single file
```
$ phctl holdings file-count umich/2025/umich_mon_full_20250902.tsv
1482903
```
- `holdings file-sample PATH` - first 50 lines of a file, or pass `--lines N` for a different count
```
$ phctl holdings file-sample umich/2025/umich_mon_full_20250902.tsv --lines 3
1234567	bib001	umich	CH		2025-01-01		spm			0	uuid-1
1234568	bib002	umich	CH		2025-01-01		spm			0	uuid-2
1234569	bib003	umich	CH		2025-01-01		spm			0	uuid-3
```

PATH can be a local path or a remote rclone path.

### DIFF
--- a/lib/phctl.rb
+++ b/lib/phctl.rb
@@ -1,7 +1,9 @@
 require "thor"
 
 require "alma_holdings"
+require "clusterable/holding"
 require "sidekiq_jobs"
+require "utils/file_transfer"
 require "workflow_component"
 
 $LOAD_PATH.unshift(File.dirname(__FILE__))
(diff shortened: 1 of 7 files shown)


lib/phctl.rb
```diff
@@ -193,6 +195,54 @@ def deposit_holdings_analysis
...
+    desc "file-sample REMOTE_PATH", "Print first N lines of a remote holdings file"
+    option :lines, type: :numeric, default: 50
+    def file_sample(remote_path)
+      Utils::FileTransfer.new.cat(remote_path) do |io|
+        io.each_line.first(options[:lines]).each { |line| print line }
+      end
+    end
+
+    desc "dir-counts REMOTE_DIR", "Count lines in each .tsv file in a remote directory"
+    def dir_counts(remote_dir)
```
moseshll, line 222: For this subcommand, and to a much lesser extent some of the others, I would consider extracting the "intelligence gathering" part of the operation from the "reporting" bit. Maybe into a `Utils::Holdings` or `Utils::HoldingsPreflight` class with a method that returns a Hash with the counts, and the total. Then stringify that in this file.

My rationale: to try to keep `phctl.rb` slim -- it only seems to grow -- and to gain more direct access to the guts for testing, without `PHCTL` being the gatekeeper of sorts.

moseshll (approved):
I like everything here. Dunno why coverage declined a bit, it all looks fine. One note on a potential improvement. APPROVE

moseshll:
Note mostly to self: this should get a minor version bump when we build/tag/release. I always default to incrementing the patch number, and that's not appropriate here.

lib/phctl.rb
aelkiss: Agreed, this should go elsewhere; PHCTL shoud handle the arguments and build the object that does the actual work.

lib/phctl.rb
```diff
@@ -193,6 +195,54 @@ def deposit_holdings_analysis
     end
   end
 
+  class Holdings < JobCommand
+    desc "count ORGANIZATION", "Count holdings currently loaded in the database for ORGANIZATION"
+    def count(org)
```
aelkiss, line 201: I think we only need count broken down by format, but I might have missed something saying that count overall for the organization would be useful?

aelkiss (commented):
Logic and tests all look good to me; agreed with @moseshll we should extract some of the logic from phctl to another class; also not sure we need the overall `count` by organization.

lib/phctl.rb
moseshll: What I'd like to see is the command `phctl holdings count ORG` with the same output as the proposed `phctl holdings format-counts ORG`. Having the total is useful, the format counts are most useful, and I'd be happy with a one-stop shop.

eumalin: I extracted the logic into `Utils::HoldingsPreflight` and collapsed count + format-counts into one count command that shows the breakdown.

aelkiss (approved):
This looks pretty good to me now. I'm not sure if we still need all the tests in `phctl_spec` esp. if they duplicate tests that are now in `file_transfer_spec` or are testing logic we could test more easily in `file_transfer_spec`, but I don't think they hurt to have for now. If we need to change the functionaltiy for any of these and find ourselves having to change tests in multiple places, that might be a signal that some of those tests could go.

PR #452, opened 2026-06-19
Reduce sidekiq retry notifications

eumalin:
Reduce Sidekiq retry noise by notifying only at the 3rd and 7th failures instead of every retry.

### DIFF
--- a/spec/utils/sidekiq_failure_notifier_spec.rb
+++ b/spec/utils/sidekiq_failure_notifier_spec.rb
@@ -8,12 +8,6 @@ RSpec.describe Utils::SidekiqFailureNotifier do
 
   let(:error) { RuntimeError.new("disk full") }
 
-  describe "NOTIFY_AT_RETRY_COUNTS" do
-    it "triggers on the 3rd and 7th failures" do
-      expect(described_class::NOTIFY_AT_RETRY_COUNTS).to eq [1, 5]
-    end
-  end
-
   describe ".failure_message" do
     it "includes will retry notice, job class, args, error class, and message" do
       job = {"class" => "Jobs::Load::Holdings", "args" => ["/data/umich.ndj"], "retry_count" => 1}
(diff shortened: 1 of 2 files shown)


spec/utils/sidekiq_failure_notifier_spec.rb
```diff
@@ -8,9 +8,9 @@
 
   let(:error) { RuntimeError.new("disk full") }
 
-  describe "NOTIFY_AT_RETRY_COUNT" do
-    it "is 1 (triggers on the 3rd failure and beyond)" do
-      expect(described_class::NOTIFY_AT_RETRY_COUNT).to eq 1
+  describe "NOTIFY_AT_RETRY_COUNTS" do
```
moseshll: There's something a little `expect([]).to eq([])` about this. Is there a way to hide `NOTIFY_AT_RETRY_COUNTS` behind a class method like `self.alert_at_retry?(r)` and, to test it, maybe throw a series of numbers (and maybe some fuzz?) at it and make sure it returns `true` for at least one (of the numbers, not the fuzz LOL)?

Maybe there's a way to repurpose part of the first test so you don't have to commit to specific `retry_count` values in the subsequent tests. I dunno. I may be veering off into the weeds.

moseshll (approved):
This does what it needs to. Again, see comments. APPROVE

spec/utils/sidekiq_failure_notifier_spec.rb
moseshll: And maybe `NOTIFY_AT_RETRY_COUNT` is just config, and doesn't belong to the class at all?

aelkiss: IMO this is maybe not even neccessary to test as such -- the test below for posting to slack seem sufficient.

aelkiss (commented):
I think we can remove the tests for `NOTIFY_AT_RETRY_COUNTS` and leave just the slack tests. Otherwise LGTM.

spec/utils/sidekiq_failure_notifier_spec.rb
eumalin: That makes sense, I removed the test.

aelkiss (approved):
Looks good to me now!

PR #454, opened 2026-06-23
ETT-253 Holdings: handle change in format; automate/improve old holdi…

moseshll:
…ngs deletion

- Add `--allow-delete` flag to `phctl scrub` command (default `false`) as a subflag to `--type-check=false`.
  - When set, this causes `ScrubRunner` to identify defunct types and backup/delete them.
  - `Jobs::Load::HoldingsDeletion` (a new parallel to `Jobs::Load::Holdings`) calls the cleanup code directly....
  - The real work happens in `Loader::HoldingLoader::DeletionCleanup` which shares code with the existing `Loader::HoldingLoader::Cleanup`.
- There's also a first whack at an optimized `PreLoadBackup::delete_marked!`
  - The wisdom seems to be that `WHERE` and `ORDER` should be over indexed columns, I've verified that the schema conforms, and updated the repo's schema to reflect what's in production.
  - `.limit.delete` is apparently a no-no, as noted in the code. Superficially it appears to work, but it makes me nervous.
  - I have tested large deletes with a 10M row subset of the production holdings table. The 10k-at-a-time approach in this PR appears very fast there (1.5s on the first slice and getting faster with each subsequent one), but of course take that with a grain of salt.
- Tidy Last:
  - Suppress Sidekiq and rclone noise in the rspec output. Prefer "shameless green" to "shameless noise."
  - I have left the _[rspec-sidekiq] WARNING! Sidekiq will *NOT* process jobs in this environment..._ line alone, however, as it seems relevant.

### Changes Made in Response to EM & AE Feedback
Make sure all member-submitted files pass autoscrub preflighting before kicking off backup/load/delete operations.
 - Add `Scrub::Preflight` class to wrap download and autoscrub.
 - Only if no errors are detected does the operation proceed.
 - Remove `ScrubRunner#run_file` method as it tries to do way too much.
 - Extract out `handle_error` private method.

Additional tweaks and fixes:
 - `phctl.rb`
   - `phctl scrub` and `phctl scrub_file` use the default logger to handle exceptions. `exit 1` call removed.
 - `scrub/autoscrub.rb`
   - Unused `force` flag removed.
   - Replace `@out_files` with `@scrubbed_file` since class does not appear capable of generating multiple .ndjs.
     - Simpler for caller (`Scrub::Preflight` typically) not to iterate one-element array.
 - `scrub/malformed_header_error.rb`
   - The old class name `MalFormedHeaderError` is just wrong.
   - References updated in `scrub/member_holding_header.rb` and elsewhere.
 - `scrub/type_checker.rb`
   - Modify `Scrub::TypeCheckError` message to be less hand-wavey.


lib/loader/holding_loader.rb
```diff
@@ -102,16 +108,37 @@ def on_success(_status, options)
...
         "(`#{options["mono_multi_serial"]}`) — " \
         "#{new_count} records loaded, #{pre_load_backup.marked_count} old records removed."
       )
+    end
 
+    def delete_old_holdings!(options:, pre_load_backup:)
       if pre_load_backup.marked_count > 0
         Services.logger.info "deleting old #{options["mono_multi_serial"]} records for #{options["organization"]}"
+        # Get elapsed time to see how we're doing with large deletes
+        start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
```
moseshll: I'm tempted to get rid of this. The logger.info lines should tell us what we need to know about the timing, if we can even get at the logs from the success callback from sidekiq.

sql/000_schema.sql
```diff
@@ -81,24 +81,31 @@ CREATE TABLE `holdings_loaded_files` (
 
 DROP TABLE IF EXISTS `holdings`;
 CREATE TABLE `holdings` (
-  `ocn`          int(11) NOT NULL,
```
moseshll, line 84: I just plopped a `SHOW CREATE TABLE...` in here because the production schema had drifted quite a bit ahead of this.

lib/loader/holding_loader.rb
aelkiss: We should be able to query those from Loki, but worth confirming.

lib/scrub/pre_load_backup.rb
```diff
@@ -36,8 +38,20 @@ def mark_for_deletion
       records.update(delete_flag: 1)
     end
 
+    # Delete in 10k chunks ordered by primary key.
+    # See `sql/000_schema.sql` -- all WHERE and ORDER are primary or indexed.
+    # Does not do limit(N).delete in light of a comment from the author of Sequel:
+    # https://stackoverflow.com/questions/38360484/sequel-ignoring-limit#comment64179700_38377263
     def delete_marked!
-      records.where(delete_flag: 1).delete
+      loop do
```
aelkiss, line 46: We may have had something similar to this at one point in time; it may be somewhat slower than the bare `delete`, but it should avoid long-running queries & hopefully better avoid deadlocks, etc. Overall this approach makes sense to me.

lib/scrub/scrub_runner.rb
```diff
@@ -77,6 +94,37 @@ def scrub_file(filename)
...
+        mono_multi_serial: type
+      )
+      pre_load_backup.write_backup_file
+      pre_load_backup.mark_for_deletion
+
+      cleanup_data = {
+        "organization" => organization,
+        "mono_multi_serial" => type
+      }
+      batch = Sidekiq::Batch.new
```
aelkiss: I understand doing it this way insofar as we want `DeletionCleanup` to inherit from `Cleanup` and use the same interface, (i.e. call `on_success`).

This is the only place we're using `HoldingsDeletion` though, so maybe an alternative that avoids needing to set up the batch would be keeping `DeletionCleanup` as it is, then calling it directly in `HoldingsDeletion`:

```ruby
    class HoldingsDeletion
      include Sidekiq::Job

      def perform(cleanup_data)
        Loader::HoldingLoader::DeletionCleanup.new.on_success(:success, cleanup_data)
      end
    end 
```

and then in `scrub_runner`

```ruby
    cleanup_data = {
        "organization" => organization,
        "mono_multi_serial" => type
      }
      
      Jobs::Load::HoldingsDeletion.perform_async(cleanup_data)
```

lib/scrub/scrub_runner.rb
```diff
@@ -215,16 +263,13 @@ def download_to_work_dir(file)
...
     private
 
-    def check_new_types(files)
-      # Get the types of the files we are loading.
-      types = Set.new
-      files.each do |f|
-        member_holding_file = Scrub::MemberHoldingFile.new(f["Path"])
-        types << member_holding_file.item_type
+    # Get the types of the files we are loading.
+    def new_types(files)
```
aelkiss, line 237: Why not just 

```ruby
files.map { |f| Scrub::MemberHoldingFile.new(f["Path"]).item_type }.to_set
```

spec/scrub/scrub_runner_spec.rb
```diff
@@ -154,11 +154,39 @@ def count_loaded_files
...
+      let(:sr) { described_class.new(org1, {"force_holding_loader_cleanup_test" => true, "type_check" => false, "allow_delete" => true}) }
+
+      it "loads additional data and deletes the old" do
+        remote_d = DataSources::DirectoryLocator.new(Settings.remote_member_data, org1)
+        remote_d.ensure!
+        # Copy fixture to "dropbox" so there is a "new file" to "download",
+        FileUtils.cp(fixture_file, remote_d.holdings_current)
+        # In this scenario we have loaded (only) a ser file and now need to
+        # load a mon, so we want to load it anyway despite the type mismatch
+        load_test_data(build(:holding, organization: org1, mono_multi_serial: "ser"))
```
aelkiss: I *think* we have the ability to force loading e.g. a `mon` file if there is only a `ser`, *without* marking anything for deletion, which is probably normally what we'd want (i.e. we had loaded the ser file for a new member, there was something wrong with the mon file and it didn't get loaded, and later they resubmit that and want to load it.

I think the scenarios we'd want to cover are a change from one of these to another:
* mpm, spm, ser
* mon, ser
* mix

We certainly don't need to cover all the cases; the most likely might be the mpm, spm, ser to mon, ser case. That requires significantly more setup here, so maybe we do mon & ser to mix?

sql/000_schema.sql
aelkiss: 😬 Good call.

aelkiss (commented):
Overall this looks pretty good. I have a couple suggestions for code cleanup, and I'd like to see the test be more reflective of how we'd actually use it.

lib/scrub/pre_load_backup.rb
moseshll: My experience with the mariadb client on manual format change operations is, to expect round about 2 seconds per 1k records deleted via `DELETE FROM holdings WHERE ... LIMIT ...` without any ORDER BY or attempted optimizations.

lib/loader/holding_loader.rb
moseshll: I updated the Grafana link in Confluence, MW had noted it was broken. It's a big URL and may have extra junk. I also am updating the "cleanup done" log line to say _who_ (org) and _what_ (mono/multi/ser) so we can make sense of the logs. Accordingly, the timing code I added will go away, I don't think we need it.

spec/scrub/scrub_runner_spec.rb
moseshll: @aelkiss  I've pushed a commit with more realistic tests of the type you mention above.

I had intended to remove the `phctl --allow-delete` flag but I wanted to test the assertion about forcing a `mon` when there is only `ser` data. If you see the scrub_runner_spec.rb test "adding mon to existing ser by not passing --allow-delete", passing `"type_check" => true` will run into problems with the `TypeChecker`. Just allowing deletion will get rid of the `ser` data we want to keep. I can't really think of a way to do that "force keep" except with `--allow-delete=false`. Not that it's a normal situation at all, but Murphy's Law and all...

I'm thinking instead of two boolean flags `--allow-delete` and `--type-check`, this really is ternary, a "retention policy" if you will. Perhaps keywords "check" (the default, use the `TypeChecker` and bail out if there's funny business), "delete" (backup and delete everything), and "append" (no deletion except for old data in preexisting types).

spec/spec_helper.rb
```diff
@@ -39,6 +39,8 @@
 ])
 Sidekiq.strict_args!
 Sidekiq::Testing.inline!
+# Suppress "Sidekiq 8.1.6 connecting to Redis with options..." INFO line
```
moseshll, line 42: The changes in this file are TIDY only.

spec/sidekiq_jobs_spec.rb
```diff
@@ -13,7 +13,6 @@ def self.set_run(did_run)
   end
```
moseshll, line 14: TIDY-only changes here. I got sick of the noise in the rspec output.

lib/scrub/scrub_runner.rb
moseshll: Much cleaner and seems to work the same.

spec/scrub/scrub_runner_spec.rb
aelkiss: It took me a while to think through this but I agree with your conclusion of the 3 modes here instead of two boolean flags.

spec/scrub/scrub_runner_spec.rb
```diff
@@ -143,20 +148,97 @@ def count_loaded_files
...
+        # All data was backed up
+        ["spm", "mpm", "ser"].each do |type|
+          expect(
+            File.exist?(
+              Scrub::PreLoadBackup.new(organization: org1, mono_multi_serial: type).backup_path
+            )
+          ).to eq(true)
+        end
+        # One Slack notification for each of spm and mpm
+        expect(stub).to have_been_requested.twice
```
aelkiss: Do we want a stronger assertion that mon/ser data was loaded?

spec/scrub/scrub_runner_spec.rb
```diff
@@ -143,20 +148,97 @@ def count_loaded_files
...
+        # All data was backed up
+        ["mon", "ser"].each do |type|
+          expect(
+            File.exist?(
+              Scrub::PreLoadBackup.new(organization: org1, mono_multi_serial: type).backup_path
+            )
+          ).to eq(true)
+        end
+        # One Slack notification for each of mon and ser
+        expect(stub).to have_been_requested.twice
```
aelkiss: Likewise do we want a stronger assertion that mix was loaded?

aelkiss (commented):
The change to the ternary makes sense to me, I think. Just to verify my understanding below:

type_check: true, allow_delete: false - default behavior; loads new holdings & deletes old holdings if they are the same format

type_check: false, allow_delete: false - suggested "append" mode; loads a new file without deleting existing data

type_check: false, allow_delete: true - suggested "back up and delete" mode - backs up existing holdings & deletes, then reloads new holdings, without checking the formats of the new holdings.

type_check: true, allow_delete: true - not expected; not sure what this would do -- back up and delete existing holdings and then load new ones fresh, but only if they were the same type as the old ones? Not sure when the type check happens, but I can't see this being of any utility as if for some reason we needed to do it we still could without the type check.

I also still had one small comment in the test we might as well address if we're going to add the new option.

spec/scrub/scrub_runner_spec.rb
moseshll: Fixed up.

spec/scrub/scrub_runner_spec.rb
moseshll: Ditto.

spec/scrub/scrub_runner_spec.rb
moseshll: Latest update moves to a ternary system. I kept the same parameter name for better or worse, but added validation so we insist on reasonable values for `@type_check`

aelkiss (approved):
This looks good to me now!

lib/scrub/scrub_runner.rb
```diff
@@ -41,31 +41,44 @@ def initialize(organization, options = {})
...
           raise
         end
+      elsif type_check == "delete"
+        deleted_types = type_checker.deleted_types
       end
 
       new_files.each do |new_file|
         run_file(new_file)
       end
+      deleted_types.each do |deleted_type|
```
eumalin, line 91: Looks solid overall. One thing though: `deleted_types` gets computed from the new files' names before we actually load them, and `run_file` swallows errors instead of raising. So if the replacement file fails to load, we still delete the old data anyway, worst case an empty file list wipes everything for that org/type. Can we gate the deletion on the load actually succeeding? Could `run_file` return the `item_type` on failure (nil on success), and `run` skip `delete_type` for any type that shows up in that failed set?

moseshll: @eumalin actually I _think_ this is okay. For each file in `new_files` there is a call to `write_backup_file` and `mark_for_deletion` which will trigger deletion of the replaced data as part of the success hook. This is independent of the new code path where each of `deleted_types` has its own backup/delete operation; those types will not have replacement data, so they just need to get the heave-ho.

(I may be misunderstanding parts of your comment, bear with me. If there is a nuance I'm not seeing let me know.)

Granted there are many ways the DB can end up with inconsistent data, each `mono_multi_serial` loader is blissfully unaware of errors in other Sidekiq jobs. I'd love to see a modified architecture that left the committing of changes to a central coordinator rather than the Sidekiq bots.

eumalin: Agreed on the same type reload case, that's gated fine. The one I'm worried about is different though: cross-type retirement, like `mon`/`ser` being replaced by `mix`.

`deleted_types` is `old_types - new_types`, and `new_types` just comes from parsing filenames, not from a successful load. If mix is malformed and fails to parse or load in run_file, mon/ser still look absent and still get sent to delete_type, even though mix never loaded. `deleted_types.each` runs unconditionally regardless of whether the loads raised, so there's no gating there.

Let me know if I'm missing something, maybe that is okay for our case and is a low priority (not in scope for this PR)

aelkiss: I think what we want to do here to avoid cases of partially-loaded data or deleted data without old data, and regardless of branch:
* scrub _all_ the new types
* if _all_ succeed, then start the chunking/loading process for all (if something fails here, it should be fixable on our end without needing new data)
* and if we need to delete any old data, do that only after scrubbing _all_ the new files succeeds
* if _any_ fail, then bail out

This avoids a couple possibilities:
* Suppose someone submits two new files, mon & ser, and the old type is mix. We could end up with a case where we load *just* mon, and the mix is deleted, and then we have to "append" the ser. This can also happen when loading holdings for a new member without any old holdings loaded. "append" lets us fix this, but it would be better not to have holdings half-loaded for an organization
* Suppose someone submits new holdings types, but they're *all* bad - then we would delete all their old holdings and they'd have nothing

Regardless, I think verifying all the new files are valid before starting the loading or deletion process for any of them should help us avoid a variety of situations that previously would have needed manual intervention in the database to recover from.

aelkiss: Also note that the scrub process doesn't queue further jobs with sidekiq - it just runs inline prior to the actual "load" part (I'm pretty sure), so we don't need to worry about batch callbacks / success/failure there. I think.

moseshll: I believe the latest commit should catch most predictable errors and prevent anything from being loaded or deleted.

lib/scrub/scrub_runner.rb
```diff
@@ -203,28 +197,36 @@ def check_old_files
...
-      # Get the types of the files we are loading.
-      types = Set.new
-      files.each do |f|
-        member_holding_file = Scrub::MemberHoldingFile.new(f["Path"])
-        types << member_holding_file.item_type
+      # Do things Loader::HoldingLoader::Cleanup normally does:
+      # Remove any scrubbed files we have sitting around and upload the logs.
+      @preflights.each do |preflight|
+        preflight.clean_up!
+        Utils::FileTransfer.new.upload(preflight.scrubber.logger_path, remote_dir)
```
eumalin: Think there's a nil issue here. If download fails or AutoScrub.new errors out on a bad filename, the preflight never gets a `@scrubber` assigned, but it's already sitting in `@preflights`. Then handle_error hits preflight.scrubber.logger_path on that one and NoMethodErrors before we ever post to Slack or log the real error. Might be worth a `next unless preflight.scrubber` guard here?

lib/phctl.rb
```diff
@@ -251,17 +251,30 @@ def pry
...
+                 Does not generate an error if ORG has no previous holdings data.
+        delete - Back up and delete types that are not represented in the new files.
+        append - Add new holdings types without deleting any other types.
+                 Typically this would be used to correct a failed load for one file,
+                 in order to leave the other types intact.
+    DESC
     def scrub(org)
       Scrub::ScrubRunner.new(org, options).run
+    rescue => err
+      Services.logger.error err.message
```
eumalin, line 270: Do we want to move away from crashing/exit 1 to just logging and returning normally? Wonder if we want at least a nonzero exit here, even with the logging.

lib/scrub/scrub_runner.rb
```diff
@@ -58,35 +58,41 @@ def run
...
 
+      # Preflight all files and make sure autoscrub succeeds
       new_files.each do |new_file|
-        run_file(new_file)
+        @preflights << Scrub::Preflight.new(
+          force: force,
+          organization: organization,
+          remote_file: new_file["Path"]
+        )
+        @preflights.last.run
```
aelkiss: This took me a minute to understand what was happening (running the preflight we just added); might be clearer was

```ruby
preflight = `Scrub::Prelight.new()` ...
@preflights << preflight
preflight.run
```

or if you want to keep it without the temporary, and perhaps more idiomatically:

```ruby
@preflights << Scrub::Preflight.new().tap(&:run)
```

or maybe even more idiomatically

```ruby
@preflights = new_files.map do |new_file|
   Scrub::Preflight.new(...).tap(&:run)
end
```

lib/phctl.rb
aelkiss: I could see the utility of a nonzero exit esp. if at some point we might move more of this to running under Argo Workflows

lib/scrub/autoscrub.rb
```diff
@@ -151,7 +149,7 @@ def write_to
         "Output file moved to #{@output_struct.member_ready_to_load.to_path}"
       )
       # Move file and store new location in array.
```
aelkiss: Comment now doesn't make sense (there's no array)

spec/scrub/record_counter_spec.rb
```diff
@@ -198,7 +198,10 @@ def get_loaded
       # Load a large file on top of that, expect db count to remain small.
       FileUtils.copy(loadfile_large, remote.holdings_current)
       scrub_runner = Scrub::ScrubRunner.new(org, scrub_options)
-      scrub_runner.run
+      begin
+        scrub_runner.run
+      rescue
```
aelkiss: Do we want to be swallowing all exceptions in this test? If so, may be worth a comment as to why

aelkiss (commented):
I'm happy to see this get extracted from `ScrubRunner`; overall definitely an improvement. There are a few small comments to consider addressing.

lib/phctl.rb
moseshll: `exit 1` reinstated. I wanted to keep it too, but I hadn't thought of catching `SystemExit` in the integration. Easy fix.

lib/scrub/scrub_runner.rb
moseshll: Add a slightly different guard and a failing -> passing test. Also added a note that errors can cause not all files to be preflighted -- that phase ends after the first error. Which I think is okay. Maybe if we have a case of multiple failing files we will want to change that.

eumalin (approved):
Looks good!

lib/scrub/scrub_runner.rb
moseshll: The `map` approach doesn't work, because we want `@preflights` to contain all the `Preflight` objects created, even the one that raises when `run` is called. This is so cleanup can happen on the failed preflight.

That was the reason I was using `@preflights.last`. But even using a local `preflight` var is easier to understand so I changed it.

lib/scrub/autoscrub.rb
moseshll: Fixed.

spec/scrub/record_counter_spec.rb
moseshll: Modified to a `raise_error` expectation

moseshll:
@aelkiss requesting a quick check on the last two commits to address the issues we saw in testing.

aelkiss (approved):
All looks good to me.

PR #457, opened 2026-06-26
Add automated release pipeline

eumalin:
Right now releasing holdings-backend means manually running build, then creating a release in GitHub, then triggering deploy.

This PR adds a new single "Release" workflow that people can trigger.

It allows for picking a version bump, and it checks that tests already passed on main, creates the git tag, builds the Docker image, scans it for vulnerabilities, creates the GitHub release with auto-generated notes, and deploys to production.

Notes:
* `tag-release.yml` skips when the release was created by automation, since we build images tagged with the version directly rather than by SHA. It still runs normally for any manually created releases.
* If tests haven't passed for HEAD, or the tag already exists, the pipeline fails early with a clear message rather than getting halfway through.
* For testing, I added two options to skip either the release + deploy steps or the deploy step only. That will allow us to check if the new release workflow is doing tagging, building, and other steps correctly, before moving on to the rest. Also added a way to skip checking tests from main, for testing from a branch.

AI assisted

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,19 +47,33 @@ jobs:
           REPO: ${{ github.repository }}
           HEAD_SHA: ${{ github.sha }}
         run: |
-          TEST_RUN=$(gh run list \
-            --repo "$REPO" \
-            --workflow=tests.yml \
-            --branch=main \
-            --limit=5 \
-            --json headSha,status,conclusion \
-            | jq -r --arg sha "$HEAD_SHA" \
-              '[.[] | select(.headSha == $sha)] | .[0] | "\(.status)/\(.conclusion // "pending")"')
-          case "$TEST_RUN" in
-            "completed/success") echo "Tests passed for $HEAD_SHA ✓" ;;
-            "completed/"*) echo "Tests did not pass for $HEAD_SHA ($TEST_RUN)"; exit 1 ;;
-            *) echo "No completed test run found for $HEAD_SHA — wait for tests to pass first"; exit 1 ;;
-          esac
+          MAX_TRIES=10
+          WAIT_TIME=60
+          i=1
+          while true; do
+            TEST_RUN=$(gh run list \
+              --repo "$REPO" \
+              --workflow=tests.yml \
+              --branch=main \
+              --limit=10 \
+              --json headSha,status,conclusion \
+              | jq -r --arg sha "$HEAD_SHA" \
+                '[.[] | select(.headSha == $sha)] | .[0] | "\(.status)/\(.conclusion // "pending")"')
+            case "$TEST_RUN" in
+              "completed/success") echo "Tests passed for $HEAD_SHA ✓"; break ;;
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,157 @@
...
+          TAG: ${{ needs.compute-tag.outputs.new_tag }}
+        run: |
+          EXISTING=$(git ls-remote --tags origin "refs/tags/${TAG}")
+          [[ -n "$EXISTING" ]] && { echo "Tag $TAG already exists on remote"; exit 1; } || true
+          git tag "$TAG"
+          git push origin "$TAG"
+
+  build:
+    needs: [compute-tag, tag]
+    uses: ./.github/workflows/build.yml
```
aelkiss: Rather than using the build action here we should probably directly use hathitrust/github_actions/build -- adding the workflow dispatch in the build action here is just passing the arguments along to hathitrust/github_actions/build; I think the same for deploy.

.github/workflows/release.yml
```diff
@@ -0,0 +1,157 @@
...
+    with:
+      img_tag: ${{ needs.compute-tag.outputs.new_tag }}
+      ref: main
+      push_latest: ${{ inputs.push_latest }}
+      platforms: linux/amd64
+      rebuild: false
+
+  scan-image:
```
aelkiss, line 139: scan-image probably needs more of a review before we start using it routinely, but this could be the start of doing so.

.github/workflows/release.yml
```diff
@@ -0,0 +1,157 @@
...
+            --workflow=tests.yml \
+            --branch=main \
+            --limit=5 \
+            --json headSha,status,conclusion \
+            | jq -r --arg sha "$HEAD_SHA" \
+              '[.[] | select(.headSha == $sha)] | .[0] | "\(.status)/\(.conclusion // "pending")"')
+          case "$TEST_RUN" in
+            "completed/success") echo "Tests passed for $HEAD_SHA ✓" ;;
+            "completed/"*) echo "Tests did not pass for $HEAD_SHA ($TEST_RUN)"; exit 1 ;;
+            *) echo "No completed test run found for $HEAD_SHA — wait for tests to pass first"; exit 1 ;;
```
aelkiss: Do we want to add a sleep/retry here similar to what https://github.com/hathitrust/github_actions/blob/main/tag-release/action.yml#L54-L68 does?

.github/workflows/release.yml
eumalin: Updated to use hathitrust/github_actions/build.

.github/workflows/release.yml
eumalin: Also updated here to use sleep/retry.

kron-spar (approved):

aelkiss (approved):
Let's give this a try. I figure we can try it for a bit in holdings first, and if it seems like it's working well we could see about making it more generic for other repositories and moving it to https://github.com/hathitrust/github_actions.

PR #463, opened 2026-07-14
ETT-1536: Workflow for counting "non-current holdings" per member

aelkiss:
Non-current holdings include lost/missing, brittle, and withdrawn items. The desire is for a report that lists cases when organizations *only* report these non-current holdings.

Overview of the implementation:
* Get all items from solr (`Workflows::NonCurrentHoldingsAnalysis::DataSource`)
* `Workflows::NonCurrentHoldingsAnalysis::Analyzer` (the "map" phase) creates a `Overlap::ItemNonCurrentHoldings` for each item
  * This gathers all matching holdings (`build_non_current_holdings`)
  * For each organization, determine whether there are any current, non-brittle holdings -- if not, check if there are brittle, withdrawn, or lost/missing holdings. If only one of those, report that; if multiple, report multiple (`analyze_overlap_records`, `single_condition`)
  * Saves this data as one JSON line per item in HathiTrust
* `Workflows::NonCurrentHoldingsAnalysis::Writer` (the "reduce" phase) creates a summary of counts by organization and ic/pd (determined as the cost report does, per ETT-1536 -- icus and pdus are both treated as pd)

Tests mainly focus on the analysis part, with an integration test that incorporates the `Writer` phase as well.


lib/overlap/item_non_current_holdings.rb
```diff
@@ -0,0 +1,78 @@
...
+    def analyze_overlap_records(overlap_records)
+      conditions = Set.new
+
+      # i.e. make sure there are no current, non-brittle holdings
+      return unless overlap_records.all?(&:all_current_holdings_brittle?)
+
+      # there is at least one current holding that is brittle
+      conditions.add(:brittle) if overlap_records.any?(&:any_current_holdings_brittle?)
+
+      conditions.add(:lost_missing) if overlap_records.sum(&:lm_count).positive?
```
eumalin, line 48: `SerialOverlap` (lib/overlap/serial_overlap.rb) doesn't define wd_count/lm_count, they inherit the base Overlap stub that always returns 0. So for a serial, this sum is always 0 no matter what. An org whose only holding on a serial is WD or LM just disappears from the report instead of showing up. Does SerialOverlap need its own counts like MultiPartOverlap has?

lib/overlap/non_current_holdings_analysis.rb
```diff
@@ -0,0 +1,75 @@
+module Overlap
```
eumalin: Is this file a leftover? It defines the exact same `Overlap::ItemNonCurrentHoldings` class as `item_non_current_holdings.rb`. Nothing seems to require this path anywhere.

lib/overlap/item_non_current_holdings.rb
aelkiss: Good question; the print holdings spec doesn't allow reporting serials as brittle, withdrawn, or lost/missing, because they can only be reported at the title level, not at the level of an individual item (i.e. with enum/chron info indicating what issue of the serial it is) -- see "item type" in the spec - https://www.hathitrust.org/wp-content/uploads/2024/04/Apr-24-Specifications-for-submitting-print-holdings-updates-to-HathiTrust-2-3-1.pdf. So I think this is correct as-is.

lib/overlap/non_current_holdings_analysis.rb
aelkiss: Yes, I think this is an accidental commit.

aelkiss: Removed.

aelkiss:
@eumalin I removed the extra duplicate file and addressed the comment about behavior for serials. I'll try this in the staging environment I set up before I merge (mainly I want to get a handle on performance and make sure that's OK); do you have any other concerns blocking merge?

eumalin (approved):
Looks good to me!

aelkiss:
@eumalin Natalie confirms this is working as-is; there are a few small additional changes if you want to take a look
* Adding staging as a deployment target
* Adding a non-current holdings analysis command to phctl
* Bugfix for ensuring we only get non-deleted records from Solr

lib/workflows/non_current_holdings_analysis.rb
```diff
@@ -0,0 +1,142 @@
...
+      def header
+        [
+          "organization",
+          "pd: withdrawn",
+          "pd: lost/missing",
+          "pd: brittle",
+          "pd: multiple",
+          "ic: withdrawn",
+          "ic: lost/missing",
+          "ic: brittle:",
```
eumalin: Typo? There is a stray trailing colon.

spec/workflows/non_current_holdings_analysis_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+
+      workflow.run
+
+      output = File.open(Dir.glob("#{ENV["TEST_TMP"]}/overlap_reports/non_current_holdings_analysis_*.tsv").first)
+
+      lines = output.to_a
+      # one record for each organization with non-current holdings
+      expect(lines.count).to eq(4)
+
+      expect(lines[0]).to eq("organization\tpd: withdrawn\tpd: lost/missing\tpd: brittle\tpd: multiple\tic: withdrawn\tic: lost/missing\tic: brittle:\tic: multiple\n")
```
eumalin: Should there be a trailing colon in `brittle:`?

lib/workflows/non_current_holdings_analysis.rb
```diff
@@ -0,0 +1,142 @@
...
+            ].join("\t"))
+          end
+        end
+      end
+
+      def report_full_path
+        File.join(report_path, report_filename)
+      end
+
+      def gzip_report
```
eumalin: Not sure if this method is going to be used.

eumalin:
Other than the small things, everything else looks good to me.

lib/workflows/non_current_holdings_analysis.rb
aelkiss: Fixed

spec/workflows/non_current_holdings_analysis_spec.rb
aelkiss: Fixed

lib/workflows/non_current_holdings_analysis.rb
aelkiss: Nope, it was leftover from copying the outline from another file.

aelkiss:
Thanks, I'm going to go ahead and rebase, squash, and merge.

PR #465, opened 2026-07-16
Cleanup: remove unused configuration keys

aelkiss:
None of these configuration keys appear to be used anywhere in the code. I will remove them from the production config as well.


moseshll (approved):
Agreed, no evidence of use found.

eumalin (approved):

